PR #3, opened 2023-05-03
Add pylint action

BECATRUE:
As it is the initial action, it doesn't support any external libraries, such as `pyqt`.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,23 @@
+name: Pylint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


BECATRUE:
It causes some errors, but I think it is because there is no python files..

kangz12345:
In that case, how about adding this action after we decide the project file structure?

BECATRUE:
> In that case, how about adding this action after we decide the project file structure?

Oh, it looks right. Then how about closing this PR and #4, temporarily?

kangz12345:
> Oh, it looks right. Then how about closing this PR and #4, temporarily?

Okay, let's come back to these PRs after deciding the file structure.
Please close them for now.

BECATRUE:
As discuessed above, we will add GitHub actions after we decide the overall structure of this project.

PR #9, opened 2023-05-15
Becatrue/6/initial explorer

BECATRUE:
I created the explorer app and tested if the GUI runs without an error.

You can test in your local environment as follows:
1. Install `qiwis` referred to [how to download and install](https://github.com/snu-quiqcl/qiwis#how-to-download-and-install).
2. Run `qiwis` in the root directory of this repository.
```shell
>> qiwis
```

The GUI looks like the below.

image

This PR will close #6.

### DIFF
--- a/iquip/explorer.py
+++ b/iquip/explorer.py
@@ -1,6 +1,4 @@
-"""
-App module for showing the experiment list and opening an experiment.
-"""
+"""App module for showing the experiment list and opening an experiment."""
 
 from typing import Optional, Tuple
 
--- a/iquip/explorer.py
+++ b/iquip/explorer.py
@@ -10,7 +10,7 @@ import qiwis
 class ExplorerFrame(QWidget):
     """Frame for showing the experiment list and opening an experiment."""
 
-    def __init__(self, parent: Optional[QObject] = None):
+    def __init__(self, parent: Optional[QWidget] = None):
         """Extended."""
         super().__init__(parent=parent)
         # widgets
@@ -22,7 +22,7 @@ class ExplorerFrame(QWidget):
 
 class ExplorerApp(qiwis.BaseApp):
     """App for showing the experiment list and opening an experiment."""
-    
+
     def __init__(self, name: str, parent: Optional[QObject] = None):
         """Extended."""
         super().__init__(name, parent=parent)
--- a/iquip/explorer.py
+++ b/iquip/explorer.py
@@ -7,9 +7,19 @@ from typing import Optional
 from PyQt5.QtCore import QObject
 from PyQt5.QtWidgets import QLabel, QWidget
 
+from qiwis import BaseApp
+
 class ExplorerFrame(QWidget):
     def __init__(self, parent: Optional[QObject] = None):
         """Extended."""
         super().__init__(parent=parent)
         # widgets
         self.testLabel = QLabel("explorer", self)
+
+
+class ExplorerApp(BaseApp):
+    """App for showing the experiment list and opening an experiment."""
+    def __init__(self, name: str, parent: Optional[QObject] = None):
+        """Extended."""
+        super().__init__(name, parent=parent)
+        self.explorerFrame = ExplorerFrame()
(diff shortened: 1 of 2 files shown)


iquip/explorer.py
```diff
@@ -0,0 +1,33 @@
+"""
+App module for showing the experiment list and opening an experiment.
```
kangz12345: Although we have used this form in `qiwis`, I think we should write it just like the function or class docstrings, for example:
```python
"""App module for ... an experiment."""
```
or
```python
"""App module for ... blah blah.

More description blah blah.
"""
```

iquip/explorer.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+from typing import Optional, Tuple
+
+from PyQt5.QtCore import QObject
+from PyQt5.QtWidgets import QLabel, QVBoxLayout, QWidget
+
+from qiwis import BaseApp
```
kangz12345: Importing `qiwis` flatly would be better, because we might want to use other functions or classes in it, e.g., `loads()`, `Serializable`, etc.

iquip/explorer.py
```diff
@@ -0,0 +1,33 @@
...
+
+class ExplorerFrame(QWidget):
+    """Frame for showing the experiment list and opening an experiment."""
+    def __init__(self, parent: Optional[QObject] = None):
```
kangz12345: Let's put a blank line before defining the first method.

kangz12345: `QWidget`'s [constructor](https://doc.qt.io/qt-5/qwidget.html#QWidget) takes `QWidget parent`, not `QObject`!

iquip/explorer.py
```diff
@@ -0,0 +1,33 @@
...
+        """Extended."""
+        super().__init__(parent=parent)
+        # widgets
+        self.testLabel = QLabel("explorer", self)
+        # layout
+        layout = QVBoxLayout(self)
+        layout.addWidget(self.testLabel)
```
kangz12345, line 20: Perfectly satisfying our conventions!

iquip/explorer.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class ExplorerApp(BaseApp):
+    """App for showing the experiment list and opening an experiment."""
+    def __init__(self, name: str, parent: Optional[QObject] = None):
```
kangz12345, line 26: The same applies here as well as in `ExplorerFrame`.

kangz12345 (changes requested):
Nice start! 😄

iquip/explorer.py
BECATRUE: Your opinion is more appropriate! I will adopt the former.

iquip/explorer.py
BECATRUE: Wow, you're right! Then, why did we set the parents of the example apps as `QObject` in `qiwis`..?
I think we were missing it.

iquip/explorer.py
BECATRUE: Umm.. I think the remaing is more appropriate in `ExplorerApp`.

The reasons are
1. `BaseApp` is a subclass of `QObject` and `QObject` has a `QObject` parent, not `QWidget`.
Please refer to [QObject constructor](https://doc.qt.io/qt-6/qobject.html#QObject).
2. In `createApp()` of `Qiwis`, we set the parants of the apps as the `Qiwis` instance, and it is also `QObject`.

What do you think of it?

kangz12345: You are right, of course the apps and `Qiwis` are not `QWidget`s. They are `QObject`s which can use `PyQt5`'s signal and slot system.
We should keep using `QObject` for the `parent` of apps.

kangz12345: Please just make a blank line instead!

BECATRUE:
I applied your reviews! Please check it. @kangz12345

kangz12345 (approved):
LGTM

Issue #11, opened 2023-05-16
Simplify the subprocess and rearrange the file structure


BECATRUE:
### Feature you want to implement
I want to implement two things.
1. Simplify the subprocess
In this project, we will often use `subprocess` package to run `artiq_*` commands.
It's easy to use enough, but there are many overlapping codes.
So I suggest to create `utils.py` and implement it in this file.
The simple code looks as below:
   ```python
   import subprocess

   def run_command(cmd):
     return subprocess.run(cmd.split(" "), capture_output=True, text=True)
   ```
   Then, we can obtain the result by `run_command("blah blah").stdout`.
   See also: [How to use subprocess](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module)

2. Rearrange the file structure
In many project, there is a directory with the same name as the project and in it, there are code files.
On the outside of the directory, there are several setting and system files, such as `.gitignore`, `setup.py`, etc.

So I suggest as follow structure.
```
- iquip (root)
  |- iquip
     |- utils.py
     |- apps
        |- explorer.py
  |- .gitignore
  |- README.md
  |- ...
```

BECATRUE:
I think this should be a priority, so please leave your opinion! @kangz12345

kangz12345:
I agree with the file structure, thank you for suggesting!

However, I suggest using a different name for the command utility module.
The name `utils` seems too general, so it could be confusing later: _What kind of utility functions are here?_

For example, when we need more _utility functions_, we have to put them in the same module even if they are totally not related to each other.
Moreover, we are going to make `protocols` module, so the name `utils` would become more ambiguous.

In my opinion, more specific names like `command`, `cmdtools`, etc. would be better, depending on the actual contents of the module!

BECATRUE:
I agree that the name `utils` may cause a little confusion.
Then, I will choose the name `cmdtools`!

BECATRUE:
This issue is completed in #13, so I closed it.

PR #13, opened 2023-05-17
Simplify the subprocess and rearrange the file structure

BECATRUE:
As I wrote in #11, I implemented two gaols; simplifying the subprocess and rearranging the file structure.

In `run_command()` of `cmdtools.py`, the meaning of each argument is
- `capture_output`: If true, `stdout` and `stderr` will be captured.
- `shell`: If true, commands will be executed through the shell.
- `check`: If true, it checks a non-zero exit code. Also, if you don't set it, it causes a Pylint warning.
- `text`: If true, the result will be stored as a string. Otherwise, it will be stored as a bianry.

An example is as follows.


iquip/cmdtools.py
```diff
@@ -0,0 +1,15 @@
+"""Utility module for handling the command line."""
+
+import subprocess
+
+def run_command(cmd):
```
kangz12345: Please annotate the parameter and return types.

kangz12345 (changes requested):

iquip/cmdtools.py
BECATRUE: Oh, I forgot using type hinting! I will update it.

BECATRUE:
I applied it. @kangz12345

kangz12345 (approved):
LGTM

PR #19, opened 2023-05-23
Implement functions of the explorer

### LINKED ISSUE #10: Implement functions of the explorer
I want to add all functions of the explorer app.

As I wrote in #6, there are 3 goals.
- Show the directory structure where `artiq_master` is running.
- Able to reload the directory structure. (It may be updated during operation)
- ~~Open the experiment builder once an experiment clicked.~~ (will be implemented later)

We can only implement the last one perfectly after the experiment builder is completed, so I just add an empty function to open it and the other parts will be implemented in the later issue.

#### Show the directory structure
- You may be not able to access the root of `artiq_master` because you connect in remote. So we cannot use [QFileDialog](https://doc.qt.io/qt-5/qfiledialog.html).
- Instead, we can use [QTreeView](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QTreeView.html#qtreeview).
- The example GUI is as follows.

#### Reload the directory structure
- Create a button to request for reloading.

#### ~~Open the experiment builder (will be implemented later)~~
- Once you clicked an experiment in the above `QTreeView`, reqeust to open the experiment builder through `qiwiscall`.

### Additional context
As I wrote in the obsidian note, in this issue, we can use only the default `artiq_master` options.
See the note for details.

BECATRUE:
I implemented three main functions:
1. Show the file structure
2. When the button clicked, reload the file sturcture
3. Open the selected experiment (Only the skeleton code. It will be continued in the later issue.)

### Show the file structure
By the DFS, I looked for all experiment files in the path `${master_path}/repository`.
The postfix path 'repository' is the default option of `artiq_master`. See also [artiq_master](https://m-labs.hk/artiq/manual/management_system.html#artiq-master).

### When the button clicked, reload the file sturcture
Once the reload button clicked, it clears the file structure viewer and reload it.

### Open the selected experiment
When the open button clicked, the full path of the selected experiment is stored in `expPath` of `openExp()`.
After I implemeted the experiment builder in the next issue, this method will be implemented fully.

### Example
The file structure is as follows:

The GUI is as follows:

image

### How to run
To run this in your local, you must prepare the following step:
- In `setup.json`, set the `masterPath` argument as your experiment path (The path where `artiq_master` is running`).

This PR will close #10.


.github/workflows/pylint.yml
```diff
@@ -23,4 +23,4 @@ jobs:
         pip install pyqt5
     - name: Analyze the code with pylint
       run: |
-        pylint $(git ls-files 'iquip/*.py')
\ No newline at end of file
+        pylint $(git ls-files 'iquip/*.py')
```
kangz12345, line 27: What is changed here?

.pylintrc
```diff
@@ -8,4 +8,4 @@ method-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9]*
 variable-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9]*)*))$
 
 [TYPECHECK]
-ignored-modules=qiwis
\ No newline at end of file
+ignored-modules=qiwis, iquip
```
kangz12345: Why should `iquip` be ignored? We have direct access to this package, don't we?

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+
 class ExplorerFrame(QWidget):  # pylint: disable=too-few-public-methods
     """Frame for showing the experiment list and opening an experiment."""
 
     def __init__(self, parent: Optional[QWidget] = None):
         """Extended."""
         super().__init__(parent=parent)
         # widgets
-        self.testLabel = QLabel("explorer", self)
+        self.expStructure = QTreeWidget(self)
```
kangz12345: I think the name should have `Tree` since it is a `QTreeWidget`.
In addition, "exp" is not quite a good abbreviation (there are many matches: experiment, experience, explore, exponential, etc.), so please use it as a full word if you want to use it.
I would suggest `experimentTree` or `fileTree`.
Of course you may choose any other names you like, but I think `expStructure` is not so clear.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+from PyQt5.QtCore import QObject, pyqtSlot
+from PyQt5.QtWidgets import (
+    QPushButton, QTreeWidget, QTreeWidgetItem, QVBoxLayout, QWidget
+)
 
 import qiwis
 
+from iquip import cmdtools
+
 class ExplorerFrame(QWidget):  # pylint: disable=too-few-public-methods
```
kangz12345: By the way, why does pylint complain about this?
Because of its inheritance, it should not be a problem.
Is it because `PyQt5` library does not provide the source codes?

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+        self.openButton = QPushButton("Open", self)
         # layout
         layout = QVBoxLayout(self)
-        layout.addWidget(self.testLabel)
+        layout.addWidget(self.reloadButton)
+        layout.addWidget(self.expStructure)
+        layout.addWidget(self.openButton)
 
 
 class ExplorerApp(qiwis.BaseApp):  # pylint: disable=too-few-public-methods
```
kangz12345, line 33: This one as well, why does it requires the disable comment?
However, now it is clear that the comment should be removed!

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
     """App for showing the experiment list and opening an experiment."""
 
-    def __init__(self, name: str, parent: Optional[QObject] = None):
+    def __init__(self, name: str, masterPath: str = ".", parent: Optional[QObject] = None):
         """Extended."""
```
kangz12345: Please add the docstring for `masterPath`.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
         super().__init__(name, parent=parent)
+        self.repositoryPath = os.path.join(masterPath, "repository")
```
kangz12345: Hmm, the `os.path.join()` will join the path based on the current client's OS.
Will it be okay even if the OS of the master server differs from the client's?

iquip/apps/explorer.py
kangz12345: Please replace all the exp abbreviations with the full words.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+    @pyqtSlot()
+    def loadExpStructure(self):
+        """Loads the experiment file structure in self.explorerFrame.expStructure.
+
+        It assumes that all experiment files are in self.repositoryPath.
+        """
+        self.explorerFrame.expStructure.clear()
+        self._addExpFile(self.repositoryPath, self.explorerFrame.expStructure)
+
+    def _addExpFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
```
kangz12345: Please write the docstring for the arguments.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+
+        This uses the DFS algorithm.
+        1. Fetch the command result, which is a list of sub elements.
+        2. If a directory, call _addExpFile() recursively.
+        3. Otherwise, add it to self.explorerFrame.expStructure.
+        """
+        expList = cmdtools.run_command(f"artiq_client ls {path}").stdout
+        expList = expList.split("\n")[:-1]  # The last one is always an empty string.
+        for expFile in expList:
+            if expFile.endswith("/") and not expFile.startswith("_"):
```
kangz12345: Please write description about ignoring the files whose name starts with "_", in the docstring.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+        self.explorerFrame.expStructure.clear()
+        self._addExpFile(self.repositoryPath, self.explorerFrame.expStructure)
+
+    def _addExpFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
+        """Searches the sub elements and add them into self.explorerFrame.expStructure.
+
+        This uses the DFS algorithm.
+        1. Fetch the command result, which is a list of sub elements.
+        2. If a directory, call _addExpFile() recursively.
+        3. Otherwise, add it to self.explorerFrame.expStructure.
```
kangz12345: I think your code is clear enough, so this comment is not necessary!

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+    def loadExpStructure(self):
+        """Loads the experiment file structure in self.explorerFrame.expStructure.
+
+        It assumes that all experiment files are in self.repositoryPath.
+        """
+        self.explorerFrame.expStructure.clear()
+        self._addExpFile(self.repositoryPath, self.explorerFrame.expStructure)
+
+    def _addExpFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
+        """Searches the sub elements and add them into self.explorerFrame.expStructure.
```
kangz12345: Since this method is called recursively, this docstring is wrong.
In fact, it `Searches all files in path and adds them into parent.`

kangz12345: Moreover, you can say `Recursively searches ...` to emphasize that this is DFS.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+
+        This uses the DFS algorithm.
+        1. Fetch the command result, which is a list of sub elements.
+        2. If a directory, call _addExpFile() recursively.
+        3. Otherwise, add it to self.explorerFrame.expStructure.
+        """
+        expList = cmdtools.run_command(f"artiq_client ls {path}").stdout
```
kangz12345: Shouldn't we catch any exceptions here?
Since [`subprocess.run()`](https://docs.python.org/3.7/library/subprocess.html#subprocess.run) raises an exception when the call returns with non-zero value when `check=True`, I think we have to catch it.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+        expList = expList.split("\n")[:-1]  # The last one is always an empty string.
```
kangz12345: Good comment 👍

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+        for expFile in expList:
+            if expFile.endswith("/") and not expFile.startswith("_"):
+                expFileItem = QTreeWidgetItem(parent)
+                expFileItem.setText(0, expFile[:-1])
```
kangz12345: How about just leaving the trailing "/"?
It would be helpful to distinguish directories from files for the GUI users as well.

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+                self._addExpFile(os.path.join(path, expFile), expFileItem)
```
kangz12345: Here as well, will it be okay to use the client-OS-based path joining?

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+            elif expFile.endswith(".py"):
```
kangz12345: How about excluding the files whose name starts with "_"?

setup.json
```diff
@@ -4,7 +4,10 @@
             "module": "iquip.apps.explorer",
             "cls": "ExplorerApp",
             "show": true,
-            "pos": "left"
+            "pos": "left",
+            "args": {
+                "masterPath": "/home/jaehun/quiqcl/artiq/examples/"
```
kangz12345: How about setting this as the default ARTIQ master path?

iquip/apps/explorer.py
```diff
@@ -1,32 +1,94 @@
...
+                expFileItem.setText(0, expFile)
+
+    @pyqtSlot()
+    def openExp(self):
+        """Opens the experiment builder of the selected experiment.
+
+        Once the openButton is clicked, this is called.
+        If the selected element is a directory, it will be ignored.
+
+        TODO(BECATRUE): Open the experiment builder.
```
kangz12345: Let's write _when_ it should be done, e.g., when the experiment builder is implemented.

kangz12345 (changes requested):
I come up with two possible issues:
1. The running time of loading the file structure could be long since it is socket communication and the file structure might grow huge someday. Then it will block the main thread, so threading might be necessary. How do you think of this?
2. When the file structure becomes huge, the user might want to load only a part of the structure, e.g., only a specific directory. So I want to suggest _lazy loading_, i.e., loading the directory only when requested, and no recursive loading. However, if it might take some effort to implement this, then please create a new feature issue about this and keep the current implementation, since it is good enough for now.

BECATRUE:
> The running time of loading the file structure could be long since it is socket communication and the file structure might grow huge someday. Then it will block the main thread, so threading might be necessary. How do you think of this?

Yes, you're right!
Indeed, in ARTIQ, it is implemented by using `await` so it doesn't guarantee the stability of GUI.
But I think it isn't a good implementation and as you suggested, we need to implement to keep the GUI from breaking.

Then, how about extracting it as a new issue and not implementing it in this issue?
It looks not a necessary function, but an advanced function!

BECATRUE:
> When the file structure becomes huge, the user might want to load only a part of the structure, e.g., only a specific directory. So I want to suggest lazy loading, i.e., loading the directory only when requested, and no recursive loading. However, if it might take some effort to implement this, then please create a new feature issue about this and keep the current implementation, since it is good enough for now.

It looks a great idea!
As you said, it seems to need some effort, so let's make a new issue.

.github/workflows/pylint.yml
BECATRUE: Umm.. I think there is no change..

Maybe, when I resolved the rebase confliction, a commit was created about this file.

.pylintrc
BECATRUE: Umm.. you're right! I will look for a solution in terms of the import error.

iquip/apps/explorer.py
BECATRUE: Oh, I made a mistake.
As you said, the comment is not necessary.
I will remove it.

iquip/apps/explorer.py
BECATRUE: I agree with you.
I will replace `exp` into `experiment` and `expStructure` into `fileTree`.

iquip/apps/explorer.py
BECATRUE: Probably, the Pylint couldn't import `qiwis` so it causes the warning.
I think I should find out the solution to import `qiwis` in the Pylint.

.github/workflows/pylint.yml
kangz12345: Oh I see, then never mind!

iquip/apps/explorer.py
BECATRUE: Wow, you are so keen.

In according to [os.path](https://docs.python.org/3/library/os.path.html), I will use `posixpath` instead of `os.path`.

iquip/apps/explorer.py
BECATRUE: Okay! I will remove it.

iquip/apps/explorer.py
kangz12345: Oh I see. However, now it has several public methods so it does not require the comment anyway.
Of course importing `qiwis` would be necessary.

iquip/apps/explorer.py
kangz12345: Oh great, can you just import the `posixpath`?
[`pathlib`](https://docs.python.org/3.7/library/pathlib.html#pathlib.PurePosixPath) looks also helpful.

iquip/apps/explorer.py
BECATRUE: Oh, I missed that part!
How about catching an exception in `cmdtools.run_command()` and in such case, returning `None`.
It looks more convenient for using this function.

iquip/apps/explorer.py
BECATRUE: Originally, I appended the "/" to directories, but it looks a little weird for me.
I thinks it is easy enough to distinguish because directories display toggles.
Nevertheless, do you think it can be confusing?

iquip/apps/explorer.py
BECATRUE: As written above, I will replace it to `posixpath.join`.

iquip/apps/explorer.py
BECATRUE: Okay! I thought it is not necessary because it is not common to start with "_" for files.
But you think so, I will implement it.

setup.json
BECATRUE: Umm.. Unfortunately there is no default path because `artiq_dashboard` can obtain the master path using the internal socket communication.
Then, how about setting it as the path in mini PC?

iquip/apps/explorer.py
BECATRUE: Okay! I will update it.

iquip/apps/explorer.py
kangz12345: It will be convenient, but it loses the error information.
Maybe we can add a parameter to `run_command()` to return or raise the exception or just ignore it.
However, in that case, it seems using `subprocess` itself is better..
If you think using `cmdtools` is still better, you can just ignore the exception for now. Later we have to update the API when we need exception information.

iquip/apps/explorer.py
kangz12345: Oh 😄 okay then. Actually it is not so confusing.

setup.json
kangz12345: Oh, I was referring to that! Let's set it to the mini PC path.

iquip/apps/explorer.py
BECATRUE: Yes! I can import it just writing `import posixpath`.

iquip/apps/explorer.py
BECATRUE: Umm.. Let's ignore it for now and improve it later!
I don't think we need to consider it much yet.

iquip/apps/explorer.py
```diff
@@ -12,82 +12,88 @@
...
-        self._addExpFile(self.repositoryPath, self.explorerFrame.expStructure)
+        self.explorerFrame.fileTree.clear()
+        self._addFile(self.repositoryPath, self.explorerFrame.fileTree)
+
+    def _addFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
+        """Searches all files in path and adds them into parent.
 
-    def _addExpFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
-        """Searches the sub elements and add them into self.explorerFrame.expStructure.
+        A directory which starts with "_" will be ignored, e.g. __pycache__/.
```
kangz12345: Now it includes files that are not directories!

BECATRUE: Oh, I will modify it!

BECATRUE:
I appied the reviews and finally, `qiwis` is installed well! @kangz12345

kangz12345 (approved):
LGTM!

PR #23, opened 2023-05-25
Single TTL monitor widget

### LINKED ISSUE #15: Single TTL monitor widget
A widget and object which monitors a~~n ARTIQ~~ TTL input or output.
~~When the input or output changes, the monitor should be able to notice it. (See #17)~~
Note that this does not include any `qiwi` app.
It will be implemented in the next issue.

Create a module `monitor.py` and implement classes ~~`TTLMonitor`~~ `Monitor` and `TTLMonitorWidget`.

#### ~~`TTLMonitor`~~
- It is a non-GUI object which monitors a single TTL channel.
- Specific implementation will be decided later.

#### `Monitor`
- It defines the monitor interface.

#### `TTLMonitorWidget`
- It is a PyQt widget for displaying a single TTL channel status.
- It does not include the proper slot connections to actually update the status. It only includes the GUI part, without the core logic.

### Additional context
See also: #12

kangz12345:
This closes #15.

In this PR, only the `Monitor` and `TTLMonitorWidget` are implemented.
Since they are too _abstract_, I will make unit tests for those after this PR.

### DIFF
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -1,6 +1,5 @@
 """Module for monitoring devices."""
 
-import enum
 from typing import Any, TypeVar, Generic, Optional, Callable
 
 from PyQt5.QtWidgets import QWidget, QLabel, QHBoxLayout
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -10,7 +10,13 @@ class Monitor(Generic[T]):
     """An interface for a monitor.
     
     Monitors watch a single value, and notify any change of the value by calling
-      a callback function.
+    a callback function.
+    
+    In a Monitor object, there is two kinds of values: the public self.value()
+    and the internal self._value.
+    In most cases they will be the same, as self.value() returns self._value by default.
+    However, for some reason, one might want to keep it different.
+    In such cases, be aware of which kind of value you are using.
     """
 
     def __init__(
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -17,12 +17,6 @@ class Monitor(Generic[T]):
 
     This class can be used as a concrete Monitor if the usage is simple enough
     and there is no need to implement self._read().
-    
-    In a Monitor object, there is two kinds of values: the public self.value()
-    and the internal self._value.
-    In most cases they will be the same, as self.value() returns self._value by default.
-    However, for some reason, one might want to keep it different.
-    In such cases, be aware of which kind of value you are using.
     """
 
     def __init__(
@@ -42,18 +36,18 @@ class Monitor(Generic[T]):
         """Returns the latest value."""
         return self._value
 
-    def set_value(self, _value: T):
+    def set_value(self, value: T):
         """Sets the current monitored value.
         
         This method will be called by the value source, to notify the value change.
         This will call the callback function.
 
         Args:
-            _value: The new value, which should be the same kind of the internal self._value.
+            value: The new value.
         """
-        self._value = _value
+        self._value = value
         if self.updated_callback is not None:
-            self.updated_callback(self.value())
+            self.updated_callback(value)
 
     def update(self):
         """Updates the current value by actively reading the value.
@@ -67,9 +61,6 @@ class Monitor(Generic[T]):
         """Actively reads the current value and returns it.
         
         Unless this is overriden, it just returns the latest value.
-        
-        The return value must be the same kind of the internal self._value,
-        in case where self.value() differs from self._value.
         """
         return self._value
 
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -53,21 +53,13 @@ class Monitor(Generic[T]):
         if self.updated_callback is not None:
             self.updated_callback(self.value())
 
-    def update(self, callback: bool = True) -> Optional[T]:
-        """Returns the current value or None if it is unknown or invalid.
+    def update(self):
+        """Updates the current value by actively reading the value.
         
-        This method will actively read the value and update the current value if it is changed.
-        It could be helpful when the monitor is newly created and has no value yet,
+        This could be helpful when the monitor is newly created and has no value yet,
         or the value source does not support reporting all the changes.
-
-        Args:
-            callback: Whether to call updated_callback, with the updated value.
-              If updated_callback is None, it is ignored.
         """
-        self._value = self._read()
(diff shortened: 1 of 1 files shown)


BECATRUE:
First of all, how about moving this file into `apps`?
It is related to GUI and `pyqt5`, so it looks better.

iquip/monitor.py
```diff
@@ -0,0 +1,112 @@
+"""Module for monitoring devices."""
+
+from typing import Any, TypeVar, Generic, Optional, Callable
+
+from PyQt5.QtWidgets import QWidget, QLabel, QHBoxLayout
+from PyQt5.QtCore import pyqtSignal
+
+
```
BECATRUE, line 8: How about removing a blank line?
Of course, we put two lines after import in `qiwis.py`, but only one line in most other codes such as examples of `qiwis` and `explorer.py`.

BECATRUE: How about removing a blank line?
Of course, we put two lines after import in `qiwis.py`, but only one line in most other codes such as examples of `qiwis` and `explorer.py`.

iquip/monitor.py
```diff
@@ -0,0 +1,112 @@
...
+T = TypeVar("T")
+
+
+class Monitor(Generic[T]):
```
BECATRUE, line 11: I didn't expect to see the generic in python😮

iquip/monitor.py
```diff
@@ -0,0 +1,112 @@
...
+    ):
+        """
+        Args:
+            initial_value: The initial value of self._value.
+            updated_callback: A function which will be called when the value is updated.
+        """
+        self.updated_callback = updated_callback
+        self._value = initial_value
+
+    def value(self) -> T:
```
BECATRUE: How about naming it to `get_value()`?
It is the opposite role of `set_value()`, so it would be better if the name is the same format.

kangz12345:
> First of all, how about moving this file into `apps`?
> It is related to GUI and `pyqt5`, so it looks better.

Oh, I thought the `apps` was for the `qiwis`-apps.
I think these are not dependent to `qiwis`, and I will make another file in `apps` for the `qiwis` app which uses the current `monitor` module.

iquip/monitor.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+
+    valueUpdated = pyqtSignal(object)
+
+    def __init__(self, monitor: Monitor[Optional[bool]], parent: Optional[QWidget] = None):
+        super().__init__(parent=parent)
+        self.monitor = monitor
+        self.monitor.updated_callback = self._setValue
+        # widgets
+        self.stateLabel = QLabel("--")
```
BECATRUE: How about setting the parent to `self` explicitly?

iquip/monitor.py
```diff
@@ -0,0 +1,112 @@
...
+    Attributes:
+        monitor: A TTL monitor object, whose values are True, False or None, which
+          represent HIGH, LOW or UNDEFINED, respectively.
+          Its updated_callback function is _setValue method of this TTLMonitorWidget instance,
+          which implies that onlt the monitor can change the widget value state.
+        stateLabel: A QLabel object which represents the current monitor value.
+          See _setValue() for the exact text for each state.
+    """
+
+    valueUpdated = pyqtSignal(object)
```
BECATRUE, line 83: Where will this signal be used?

iquip/monitor.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __init__(self, monitor: Monitor[Optional[bool]], parent: Optional[QWidget] = None):
```
BECATRUE, line 85: Please write the docstring about `monitor`.

iquip/monitor.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Signals:
+        valueUpdated: Emitted when the monitor value is updated. The argument is
+          True, False or None.
+    
+    Attributes:
+        monitor: A TTL monitor object, whose values are True, False or None, which
+          represent HIGH, LOW or UNDEFINED, respectively.
+          Its updated_callback function is _setValue method of this TTLMonitorWidget instance,
+          which implies that onlt the monitor can change the widget value state.
```
BECATRUE: Isn't `onlt` a typo?

BECATRUE (commented):
That's a good start!

BECATRUE (changes requested):
That's a good start!

BECATRUE:
> Oh, I thought the `apps` was for the `qiwis`-apps. I think these are not dependent to `qiwis`, and I will make another file in `apps` for the `qiwis` app which uses the current `monitor` module.

Okay! Let's leave it the way it is.

iquip/monitor.py
kangz12345: Okay, I will put only one there.

iquip/monitor.py
kangz12345: That's right, I will change it!

iquip/monitor.py
kangz12345: I'm not sure for now, but I think there must be a way to know when the value is updated.

iquip/monitor.py
kangz12345: Definitely it is :)

iquip/monitor.py
BECATRUE: Oh, I agree with you. I just wondered what you were planning.

iquip/monitor.py
```diff
@@ -0,0 +1,116 @@
...
+    and there is no need to implement self._read().
+    """
+
+    def __init__(
+        self,
+        initial_value: T,
+        updated_callback: Optional[Callable[[T], Any]] = None,
+    ):
+        """
+        Args:
```
BECATRUE, line 27: Please add `Extended` into the docstring!

BECATRUE (changes requested):
This may be a final review!

iquip/monitor.py
kangz12345: Hmm, actually this is not _extended_ because:
1. I did not use `super().__init__()`, so it would be _overridden_ instead.
2. However, it is not overridden, since `Generic` does not define `__init__()`.

BECATRUE: Oh, I see! Then, there is no problem.

BECATRUE (approved):
LGTM!

PR #25, opened 2023-05-29
Unit tests for monitor

### LINKED ISSUE #24: Unit test for `Monitor` and `TTLMonitorWidget`
This will be implemented after #23

kangz12345:
This closes #24.

I apologize for the large diffs..

This PR includes:
- Unit tests for the current `iquip.monitor` module (implemented in #23)
- Tiny improvement in `iquip.monitor`: initialization of state label text
- `.pylintrc` file for pylinting test codes

You can run the unit test by:
```console
python -m unittest tests.test_monitor
```

You can run the pylint checker by:
```console
pylint tests.test_monitor --rcfile tests/.pylintrc
```

### DIFF
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -90,14 +90,14 @@ class TTLMonitorWidget(QWidget):
         """
         super().__init__(parent=parent)
         self.monitor = monitor
-        self.monitor.updated_callback = self._setValue
+        self.monitor.updated_callback = self._updateValue
         # widgets
         self.stateLabel = QLabel("--", self)
         # layout
         layout = QHBoxLayout(self)
         layout.addWidget(self.stateLabel)
 
-    def _setValue(self, value: Optional[bool]):
+    def _updateValue(self, value: Optional[bool]):
         """Sets the current value on the label.
 
         This method is internal since it is intended to be called only by the monitor callback.
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -93,13 +93,15 @@ class TTLMonitorWidget(QWidget):
         self.monitor.updated_callback = self._updateValue
         # widgets
         self.stateLabel = QLabel("--", self)
-        self._setValue(monitor.get_value())
+        self._setTextWith(monitor.get_value())
         # layout
         layout = QHBoxLayout(self)
         layout.addWidget(self.stateLabel)
 
-    def _setValue(self, value: Optional[bool]):
-        """Sets the current value on the label.
+    def _setTextWith(self, value: Optional[bool]):
+        """Sets the current text on the label with value.
+
+        This does not emit any signal.
 
         Args:
             value: TTL state value.
@@ -121,5 +123,5 @@ class TTLMonitorWidget(QWidget):
         Args:
             value: TTL state value which is passed by the monitor.
         """
-        self._setValue(value)
+        self._setTextWith(value)
         self.valueUpdated.emit(value)
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -85,6 +85,8 @@ class TTLMonitorWidget(QWidget):
     def __init__(self, monitor: Monitor[Optional[bool]], parent: Optional[QWidget] = None):
         """Extended.
 
+        Based on the current value of monitor, stateLabel text will be initialized.
+
         Args:
             monitor: A Monitor object whose value will be displayed on the widget.
         """
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -93,6 +93,7 @@ class TTLMonitorWidget(QWidget):
         self.monitor.updated_callback = self._updateValue
         # widgets
         self.stateLabel = QLabel("--", self)
+        self._setValue(monitor.get_value())
         # layout
         layout = QHBoxLayout(self)
         layout.addWidget(self.stateLabel)
--- a/iquip/monitor.py
+++ b/iquip/monitor.py
@@ -97,14 +97,11 @@ class TTLMonitorWidget(QWidget):
         layout = QHBoxLayout(self)
         layout.addWidget(self.stateLabel)
 
-    def _updateValue(self, value: Optional[bool]):
+    def _setValue(self, value: Optional[bool]):
         """Sets the current value on the label.
 
-        This method is internal since it is intended to be called only by the monitor callback.
-        It changes the label text based on the value.
-        
         Args:
-            value: TTL state value which is passed by the monitor.
+            value: TTL state value.
         """
         if value is None:
             text = "--"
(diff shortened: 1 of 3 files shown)


BECATRUE:
> I apologize for the large diffs..

Ah, that's no problem! I will leave reviews.

iquip/monitor.py
```diff
@@ -85,32 +85,45 @@ class TTLMonitorWidget(QWidget):
...
         if value is None:
             text = "--"
         elif value:
             text = "HIGH"
         else:
             text = "LOW"
-        self.valueUpdated.emit(value)
         self.stateLabel.setText(text)
+
+    def _updateValue(self, value: Optional[bool]):
```
BECATRUE, line 119: Separating into two parts looks better😃

tests/.pylintrc
```diff
@@ -0,0 +1,14 @@
+[MASTER]
+extension-pkg-whitelist=PyQt5
+init-hook="from pylint.config import find_pylintrc; from pathlib import Path; import sys; sys.path.append(str(Path(find_pylintrc()).parent))"
```
BECATRUE: What is it for?

tests/test_monitor.py
```diff
@@ -0,0 +1,116 @@
...
+
+class TestMonitor(unittest.TestCase):
+    """Unit tests for Monitor class."""
+
+    def test_init_initial_value(self):
+        for value in (True, None, 1.0, "value", object()):
+            mon = monitor.Monitor(initial_value=value)
+            self.assertEqual(mon._value, value)
+
+    def test_init_updated_callback(self):
```
BECATRUE, line 19: How about checking the default value of `update_callback`?
If you think it isn't necessary, never mind!

tests/test_monitor.py
```diff
@@ -0,0 +1,116 @@
...
+        mon = monitor.Monitor(initial_value="old", updated_callback=callback)
+        mon.set_value("new")
+        callback.assert_called_once_with("new")
+
+    def test_update(self):
+        mon = monitor.Monitor(initial_value="old")
+        with mock.patch.multiple(mon, set_value=mock.DEFAULT, _read=mock.DEFAULT) as mocked:
+            mocked["_read"].return_value = "new"
+            mon.update()
+        mocked["set_value"].assert_called_once_with("new")
```
BECATRUE, line 52: How about moving this line into `with` statement?
I think it is better because `mocked` is defined in `with` block.

tests/test_monitor.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test_read(self):
+        """Tests if _read() returns self._value.
+
+        Although _read() will be overridden in most cases,
+        this tests the default behavior of this method.
+        """
```
BECATRUE, line 59: Good explanation!

tests/test_monitor.py
```diff
@@ -0,0 +1,116 @@
...
+        _, widget = get_ttl_monitor_widget_with(None)
+        with mock.patch.multiple(
+            widget, _setTextWith=mock.DEFAULT, valueUpdated=mock.DEFAULT
+        ) as mocked:
+            widget._updateValue(True)
+        mocked["_setTextWith"].assert_called_once_with(True)
+        mocked["valueUpdated"].emit.assert_called_once_with(True)
+
+
+def get_ttl_monitor_widget_with(
```
BECATRUE, line 106: Why did you defining it outside `TestTTLMonitorWidget`?

BECATRUE: Referred to [unittest](https://docs.python.org/3/library/unittest.html#basic-example), only if the name starts with `test`, it will be tested.

tests/test_monitor.py
```diff
@@ -0,0 +1,116 @@
...
+        mocked_set_value.assert_called_once_with(True)
+        callback.assert_not_called()
+
+    def test_set_text_with(self):
+        for value, text in ((True, "HIGH"), (False, "LOW"), (None, "--")):
+            _, widget = get_ttl_monitor_widget_with(None)
+            with mock.patch.object(widget, "valueUpdated") as mocked_signal:
+                widget._setTextWith(value)
+            self.assertEqual(widget.stateLabel.text(), text)
+            mocked_signal.assert_not_called()
```
BECATRUE, line 94: Is it necessary? I think testing only `_setTextWith()` is enough..

BECATRUE (changes requested):
The code is long, but there's little to fix!

tests/.pylintrc
kangz12345: Oh, without this, pylint could not find `iquip`.
If you are doubtful if this is really necessary, we can take a closer look.
I did not fully understand why this is required, to be honest.

tests/test_monitor.py
kangz12345: Okay, I will add that test!

tests/test_monitor.py
kangz12345: It is not for test, but a helper function!
Since it does not have any dependency of the `TestCase` instance (`self` in the methods), I put it outside.

tests/test_monitor.py
kangz12345: The docstring of `_setTextWith()`, however, states that it does not emit any signal, and I thought it should be verified in the test.

tests/test_monitor.py
kangz12345: Actually I am still not sure about this.
As you said, it is defined in the `with` block so it seems intuitive to use it in the block.
However, it is in fact not a problem in Python's scope system.
The reason why I put it outside the `with` block is that the _patching_ is not required any more at that point.
I kept the same policy for other patching codes.

How do you think of this?

kangz12345: More of my opinion:

According to [PEP343](https://peps.python.org/pep-0343/#specification-the-with-statement), a `with` statement is actually a special form of `try-except-finally` statements.
In Python, variables defined in `try` block can be used outside and it is quite common as far as I know.
So I think it could be accepted to use `mocked` outside the `with` block.
Since a `with` block is a _context-managing_ area, the more important point is "until when the context should be maintained".

tests/.pylintrc
BECATRUE: Umm.. It can be resolved to relocate the pylintrc file into the root directory not `tests/`, but if you prefer the current one, that's ok!

tests/test_monitor.py
BECATRUE: Wow, thanks for the reference!
As you said, there is no problem for Python's scope system and I think it's better to leave it as it is.

tests/test_monitor.py
BECATRUE: Okay! if you intend that is doesn't emit any signal, it is okay.

tests/test_monitor.py
BECATRUE: Okay! Then, never mind.

BECATRUE (dismissed):
Please check the `tests/.pylintrc` file!
If you prefer the current one, it's okay to approve.

tests/.pylintrc
kangz12345: I tried your solution but it does not work. Did you test it?

kangz12345: Oh.. As you said, the `init-hook` was not necessary.
The import error does not occur when I run it for `tests.test_monitor`, not `tests/test_monitor.py`.

BECATRUE: It's a very difficult error to find..😂

BECATRUE (approved):
LGTM!

PR #33, opened 2023-06-02
Implement additional functions of explorer

### LINKED ISSUE #20: Non-blocking loading of file structure in Explorer
### What feature do you suggest?
See the additional context.

### How will the feature be implemented?
Using threads (or `asyncio` if it is supported in `qiwis`).

### Additional context

> The running time of loading the file structure could be long since it is socket communication and the file structure might grow huge someday. Then it will block the main thread, so threading might be necessary. How do you think of this?

Yes, you're right!
Indeed, in ARTIQ, it is implemented by using `await` so it doesn't guarantee the stability of GUI.
But I think it isn't a good implementation and as you suggested, we need to implement to keep the GUI from breaking.

Then, how about extracting it as a new issue and not implementing it in this issue?
It looks not a necessary function, but an advanced function!

_Originally posted by @BECATRUE in https://github.com/snu-quiqcl/iquip/issues/19#issuecomment-1561374272_

### LINKED ISSUE #21: Lazy loading of file structure in Explorer
### What feature do you suggest?
See the additional context.

### How will the feature be implemented?
TBD.

It looks a great idea!
As you said, it seems to need some effort, so let's make a new issue.

_Originally posted by @BECATRUE in https://github.com/snu-quiqcl/iquip/issues/19#issuecomment-1561375914_

### LINKED ISSUE #28: Unnecessary pylint disable comment in `apps.explorer`
https://github.com/snu-quiqcl/iquip/blob/ba81d66d991da9a8f9d296a60df91a0d4967ecac/iquip/apps/explorer.py#L33

Since it has more than 2 public methods, this comment is not necessary.

BECATRUE:
I implemented the additional function of #20, #21, and resolved the Pylint issue of #28.

# Threading for calling the command line
- Just used `theading.Thread`.
- In according to an [example](https://wiki.qt.io/Updating-QML-content-from-Python-threads), it may be no problem.

# Lazy loading of file structure
- When initially loaded, directories have an empty item underneath to indicate that it is a directory.
- When the directory is expanded for the first time, `_addFile()` is called and the lower files are loaded.

Close #20, Close #21, and Close #28.

### DIFF
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -1,5 +1,6 @@
 """App module for showing the experiment list and opening an experiment."""
 
+import threading
 import posixpath
 from typing import Optional, Tuple, Union
 
@@ -54,8 +55,8 @@ class ExplorerApp(qiwis.BaseApp):  # pylint: disable=too-few-public-methods
         It assumes that all experiment files are in self.repositoryPath.
         """
         self.explorerFrame.fileTree.clear()
-        self._addFile(self.repositoryPath, self.explorerFrame.fileTree)
-
+        threading.Thread(target=lambda: self._addFile(self.repositoryPath, self.explorerFrame.fileTree)).start()
+        
     def _addFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
         """Searches all files in path and adds them into parent.
 
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -33,10 +33,10 @@ class ExplorerFrame(QWidget):
 
 class FileFinderThread(QThread):
     """QThread for finding the file list using a command line.
-    
+
     Signals:
         finished(experimentList): Fetching the file list is finished.
-    
+
     Attributes:
         path: The path of the directory to search experiment files.
         parent: The widget corresponding to the path.
@@ -45,8 +45,18 @@ class FileFinderThread(QThread):
     finished = pyqtSignal(list)
 
     def __init__(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
+        super().__init__()
         self.path = path
         self.parent = parent
+
+    def run(self):
+        """Fetches the file list using a command line.
+
+        After finished, the finished signal is emitted.
+        """
+        experimentList = cmdtools.run_command(f"artiq_client ls {self.path}").stdout
+        experimentList = experimentList.split("\n")[:-1]  # The last one is always an empty string.
+        self.finished.emit(experimentList)
 
 
 class ExplorerApp(qiwis.BaseApp):
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -124,11 +124,12 @@ class ExplorerApp(qiwis.BaseApp):
         Args:
             experimentFileItem: The file item to get its full path.
         """
-        path = experimentFileItem.text(0)
+        path = [experimentFileItem.text(0)]
         while experimentFileItem.parent():
             experimentFileItem = experimentFileItem.parent()
-            path = posixpath.join(experimentFileItem.text(0), path)
-        path = posixpath.join(self.repositoryPath, path)
+            path.append(experimentFileItem.text(0))
+        path.append(self.repositoryPath)
+        path = posixpath.join(*reversed(path))
         return path
 
     def frames(self) -> Tuple[ExplorerFrame]:
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -34,14 +34,14 @@ class _FileFinderThread(QThread):
     """QThread for finding the file list using a command line.
 
     Signals:
-        finished(experimentList, widget): Fetching the file list is finished.
+        fetched(experimentList, widget): The file list is fetched.
 
     Attributes:
         path: The path of the directory to search for experiment files.
         widget: The widget corresponding to the path.
     """
 
-    finished = pyqtSignal(list, object)
+    fetched = pyqtSignal(list, object)
 
     def __init__(
         self,
@@ -59,7 +59,7 @@ class _FileFinderThread(QThread):
         super().__init__(parent=parent)
         self.path = path
         self.widget = widget
-        self.finished.connect(callback)
+        self.fetched.connect(callback)
 
     def run(self):
         """Overridden.
@@ -67,11 +67,11 @@ class _FileFinderThread(QThread):
         Fetches the file list using a command line.
 
         Searches for only files in path, not in deeper path and adds them into the widget.
-        After finished, the finished signal is emitted.
+        After finished, the fetched signal is emitted.
         """
         experimentList = cmdtools.run_command(f"artiq_client ls {self.path}").stdout
         experimentList = experimentList.split("\n")[:-1]  # The last one is always an empty string.
-        self.finished.emit(experimentList, self.widget)
+        self.fetched.emit(experimentList, self.widget)
 
 
 class ExplorerApp(qiwis.BaseApp):
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -34,11 +34,11 @@ class FileFinderThread(QThread):
     """QThread for finding the file list using a command line.
 
     Signals:
-        finished(experimentList, parent): Fetching the file list is finished.
+        finished(experimentList, widget): Fetching the file list is finished.
 
     Attributes:
         path: The path of the directory to search for experiment files.
-        parent: The widget corresponding to the path.
+        widget: The widget corresponding to the path.
     """
 
     finished = pyqtSignal(list, object)
@@ -46,7 +46,7 @@ class FileFinderThread(QThread):
     def __init__(
         self,
         path: str,
-        parent: Union[QTreeWidget, QTreeWidgetItem],
+        widget: Union[QTreeWidget, QTreeWidgetItem],
         callback: Callable[[List[str], Union[QTreeWidget, QTreeWidgetItem]], None]
     ):
         """
@@ -55,13 +55,13 @@ class FileFinderThread(QThread):
         """
         super().__init__()
         self.path = path
-        self.parent = parent
+        self.widget = widget
         self.finished.connect(callback)
 
     def run(self):
         """Fetches the file list using a command line.
 
-        Searches for only files in path, not in deeper path and adds them into parent.
+        Searches for only files in path, not in deeper path and adds them into the widget.
         After finished, the finished signal is emitted.
         """
         experimentList = cmdtools.run_command(f"artiq_client ls {self.path}").stdout
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -73,7 +73,7 @@ class ExplorerApp(qiwis.BaseApp):
         if experimentFileItem.childCount() != 1 or experimentFileItem.child(0).columnCount() != 0:
             return
         # Remove the empty item of an unloaded directory.
-        experimentFileItem.removeChild(experimentFileItem.child(0))
+        experimentFileItem.takeChild(0)
         experimentPath = self.getFullPath(experimentFileItem)
         threading.Thread(
             target=lambda: self._addFile(experimentPath, experimentFileItem)
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -62,7 +62,8 @@ class ExplorerApp(qiwis.BaseApp):
 
     @pyqtSlot(QTreeWidgetItem)
     def lazyLoadFile(self, item: QTreeWidgetItem):
-        pass
+        if item.childCount() != 1 or item.child(0).columnCount() != 0:
+            return
 
     # pylint: disable=no-self-use
     def _addFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -118,7 +118,7 @@ class ExplorerApp(qiwis.BaseApp):
 
     def getFullPath(self, experimentFileItem: QTreeWidgetItem) -> str:
         """Finds the full path of the file item and returns it.
-        
+
         Args:
             experimentFileItem: The file item to get its full path.
         """
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -49,9 +49,11 @@ class FileFinderThread(QThread):
         widget: Union[QTreeWidget, QTreeWidgetItem],
         callback: Callable[[List[str], Union[QTreeWidget, QTreeWidgetItem]], None]
     ):
-        """
+        """Extended.
+
         Args:
             callback: The callback method called after this thread is finished.
+            See the attributes section in FileFinderThread.
         """
         super().__init__()
         self.path = path
@@ -66,7 +68,7 @@ class FileFinderThread(QThread):
         """
         experimentList = cmdtools.run_command(f"artiq_client ls {self.path}").stdout
         experimentList = experimentList.split("\n")[:-1]  # The last one is always an empty string.
-        self.finished.emit(experimentList, self.parent)
+        self.finished.emit(experimentList, self.widget)
 
 
 class ExplorerApp(qiwis.BaseApp):
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -78,7 +78,6 @@ class ExplorerApp(qiwis.BaseApp):
             target=lambda: self._addFile(experimentPath, experimentFileItem)
         ).start()
 
-    # pylint: disable=no-self-use
     def _addFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
         """Searches only files in path, not in deeper path and adds them into parent.
 
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -37,7 +37,7 @@ class FileFinderThread(QThread):
         finished(experimentList, parent): Fetching the file list is finished.
 
     Attributes:
-        path: The path of the directory to search experiment files.
+        path: The path of the directory to search for experiment files.
         parent: The widget corresponding to the path.
     """
 
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -4,7 +4,7 @@ import threading
 import posixpath
 from typing import Optional, Tuple, Union
 
-from PyQt5.QtCore import QObject, pyqtSlot
+from PyQt5.QtCore import QObject, QThread, pyqtSlot
 from PyQt5.QtWidgets import (
     QPushButton, QTreeWidget, QTreeWidgetItem, QVBoxLayout, QWidget
 )
@@ -29,6 +29,10 @@ class ExplorerFrame(QWidget):
         layout.addWidget(self.reloadButton)
         layout.addWidget(self.fileTree)
         layout.addWidget(self.openButton)
+
+
+class FileFinderThread(QThread):
+    """QThread for finding the file list using a command line."""
 
 
 class ExplorerApp(qiwis.BaseApp):
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -32,9 +32,13 @@ class ExplorerFrame(QWidget):
 
 
 class FileFinderThread(QThread):
-    """QThread for finding the file list using a command line."""
+    """QThread for finding the file list using a command line.
+    
+    Signals:
+        finished(experimentList): Fetching the file list is finished.
+    """
 
-    finished = pyqtSignal()
+    finished = pyqtSignal(list)
 
 
 class ExplorerApp(qiwis.BaseApp):
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -4,7 +4,7 @@ import threading
 import posixpath
 from typing import Optional, Tuple, Union
 
-from PyQt5.QtCore import QObject, QThread, pyqtSlot
+from PyQt5.QtCore import QObject, QThread, pyqtSlot, pyqtSignal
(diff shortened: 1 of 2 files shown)


kangz12345:
FYI, you can use "resolve" as a closing keyword as well:
This PR _resolves_ #28.

iquip/apps/explorer.py
```diff
@@ -54,15 +56,36 @@ def loadFileTree(self):
...
 
     def _addFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
-        """Searches all files in path and adds them into parent.
+        """Searches only files in path, not in deeper path and adds them into parent.
 
         A file or directory which starts with "_" will be ignored, e.g. __pycache__/.
 
         Args:
             path: The path of the directory to search experiment files.
+            parent: The widget corresponding to the path.
```
kangz12345: Oops, nice work :)

iquip/apps/explorer.py
```diff
@@ -54,15 +56,36 @@ def loadFileTree(self):
...
+
+        This will be called when a directory item is expanded,
+        so it makes loading files lazy.
+
+        Args:
+            experimentFileItem: The expanded file item.
+        """
+        if experimentFileItem.childCount() != 1 or experimentFileItem.child(0).columnCount() != 0:
+            return
+        experimentFileItem.removeChild(experimentFileItem.child(0))
```
kangz12345: Please add a comment to explain that there is an empty item for an unloaded directory.

kangz12345: You can use [`takeChild(0)`](https://doc.qt.io/qt-6/qtreewidgetitem.html#takeChild).

iquip/apps/explorer.py
```diff
@@ -54,15 +56,36 @@ def loadFileTree(self):
...
+        experimentPath = self.getFullPath(experimentFileItem)
+        threading.Thread(
+            target=lambda: self._addFile(experimentPath, experimentFileItem)
+        ).start()
```
kangz12345: What a concise and clean code!

iquip/apps/explorer.py
```diff
@@ -54,15 +56,36 @@ def loadFileTree(self):
...
+
+        Args:
+            experimentFileItem: The expanded file item.
+        """
+        if experimentFileItem.childCount() != 1 or experimentFileItem.child(0).columnCount() != 0:
+            return
+        experimentFileItem.removeChild(experimentFileItem.child(0))
+        experimentPath = self.getFullPath(experimentFileItem)
+        threading.Thread(
+            target=lambda: self._addFile(experimentPath, experimentFileItem)
```
kangz12345: When you use a `lambda` expression, you should be careful about the variable scope and the closure.
For example, I guess the current code would make no problem but in the future, the code could be modified, so the `experimentPath` or `experimentFileItem` is used after the thread started.
In that case, since the variables are _shared_, some subtle concurrency issues might occur.

Therefore, I suggest to use `args` argument in [`threading.Thread()`](https://docs.python.org/3.7/library/threading.html#threading.Thread), so the variables are _passed_ over. Or, you can use [`functools.partial()`](https://docs.python.org/3.7/library/functools.html#functools.partial) as well.

iquip/apps/explorer.py
```diff
@@ -72,7 +95,8 @@ def _addFile(self, path: str, parent: Union[QTreeWidget, QTreeWidgetItem]):
             if experimentFile.endswith("/"):
                 experimentFileItem = QTreeWidgetItem(parent)
                 experimentFileItem.setText(0, experimentFile[:-1])
```
kangz12345, line 147: Since you are now run this codes in another thread, you have to be aware of the GUI event loop in the main thread.
If you update the GUI (e.g., `setText()`) in a thread other than the main thread, it could crash.
For safety, you should separate the _working_ part and the GUI part, and the GUI part must run on the main thread.
In order to accomplish this, you may use signal-slot mechanism or [`invokeMethod()`](https://doc.qt.io/qt-6/qmetaobject.html#invokeMethod-7). You might need `QThread` for these feature (I'm not sure!).
Or you could find a nice way to use `threading.Thread` instead.

iquip/apps/explorer.py
```diff
@@ -89,12 +113,20 @@ def openExperiment(self):
         experimentFileItem = self.explorerFrame.fileTree.currentItem()
         if experimentFileItem.childCount():
             return
-        experimentPath = experimentFileItem.text(0)
+        experimentPath = self.getFullPath(experimentFileItem)  # pylint: disable=unused-variable
```
kangz12345: Why is this line necessary?

iquip/apps/explorer.py
```diff
@@ -89,12 +113,20 @@ def openExperiment(self):
...
+
+    def getFullPath(self, experimentFileItem: QTreeWidgetItem) -> str:
```
kangz12345: In Qt's style, I would rather say `fullPath()`.

iquip/apps/explorer.py
```diff
@@ -89,12 +113,20 @@ def openExperiment(self):
...
+            experimentFileItem: The file item to get its full path.
+        """
+        path = experimentFileItem.text(0)
         while experimentFileItem.parent():
             experimentFileItem = experimentFileItem.parent()
-            experimentPath = posixpath.join(experimentFileItem.text(0), experimentPath)
-        experimentPath = posixpath.join(self.repositoryPath, experimentPath)
-
+            path = posixpath.join(experimentFileItem.text(0), path)
+        path = posixpath.join(self.repositoryPath, path)
```
kangz12345: How about using a `list` and extending it, and calling `join()` at the end just once?
I think `join` operation is more expansive since it repeatedly creates and copies strings.

setup.json
```diff
@@ -6,7 +6,7 @@
             "show": true,
             "pos": "left",
             "args": {
-                "masterPath": "/home/rabi/artiq/"
+                "masterPath": "/home/jaehun/quiqcl/artiq/examples/"
```
kangz12345: Maybe this is mistakenly staged here.
You can make another setup file (e.g. "setup_jaehun.json") and ignore it in your git repo.

kangz12345 (changes requested):

kangz12345 (changes requested):
In addition, consider the case where the user rapidly clicks a directory item several times, occurring the `itemExpanded` signal is emitted multiple times before the first thread is finished.

setup.json
BECATRUE: Oh, I made a mistake..
Thanks for offering a solution.

iquip/apps/explorer.py
BECATRUE: Oh, it looks simpler!

iquip/apps/explorer.py
BECATRUE: I think you told me about this issue before in the qiwis project.
I will fix it.

iquip/apps/explorer.py
BECATRUE: It is a preparation step for opening the experiment builder.
For submitting or getting the arguments, we need to know the full path, not only the file name.
So I've implemented up to here first.

iquip/apps/explorer.py
BECATRUE: Oh, it's better! I will apply it.

iquip/apps/explorer.py
BECATRUE: I found out a few solutions of setting a callback method, but they are a little complicated.
So I decide to replace it to `QThread`.

BECATRUE:
> In addition, consider the case where the user rapidly clicks a directory item several times, occurring the `itemExpanded` signal is emitted multiple times before the first thread is finished.

Oh, it's a keen analysis.

In the current system, however, we cannot clicks a directory item several times during the threading operation because before starting the thread, the child element is removed and the directory item has no child, so the toggle button would be removed.
Thus, I think we don't have to consider that situation.

What do you think of it?

kangz12345:
> In the current system, however, we cannot clicks a directory item several times during the threading operation because before starting the thread, the child element is removed and the directory item has no child, so the toggle button would be removed.
> 
> Thus, I think we don't have to consider that situation.
> 
> 
> 
> What do you think of it?

Oh, it makes sense. If it is confirmed that the toggle button is removed, then it would be okay!

iquip/apps/explorer.py
kangz12345: In that case, please leave a TODO comment describing what is this variable for and how and when it will be used.

BECATRUE: Umm.. But there is already a TODO comment in the docstring of the method, so I think an additional comment is not necessary.
How about remaining it as it is?

kangz12345: Oh, I did not see that. Then it seems alright!

kangz12345:
Please re-request the review after applying all the previous reviews.

BECATRUE:
> Please re-request the review after applying all the previous reviews.

Oh, I pushed the updated version before, but I didn't notify it..
It is already updated!

kangz12345:
You can _re-request_ a review in the _Reviewers_ section, as I just did! :)

iquip/apps/explorer.py
```diff
@@ -31,6 +30,45 @@ def __init__(self, parent: Optional[QWidget] = None):
...
 
 
+class FileFinderThread(QThread):
+    """QThread for finding the file list using a command line.
+
+    Signals:
+        finished(experimentList, parent): Fetching the file list is finished.
+
+    Attributes:
+        path: The path of the directory to search experiment files.
```
kangz12345: As far as I know:
Usually we use "for", e.g., "search for experiment files".
When we use "search A" without "for", then A is the place or object where the interested targets are in, e.g., "search the area for clues".

Please let me know if it has other usage!

iquip/apps/explorer.py
```diff
@@ -31,6 +30,45 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+        parent: Union[QTreeWidget, QTreeWidgetItem],
+        callback: Callable[[List[str], Union[QTreeWidget, QTreeWidgetItem]], None]
+    ):
+        """
+        Args:
+            callback: The callback method called after this thread is finished.
+        """
+        super().__init__()
+        self.path = path
+        self.parent = parent
```
kangz12345: I think this will overwrite the method [`QObject.parent()`](https://doc.qt.io/qt-5/qobject.html#parent), since `QThread` inherits `QObject`.

Please change its name to, e.g., `widget`, etc.

iquip/apps/explorer.py
```diff
@@ -31,6 +30,45 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+
+    finished = pyqtSignal(list, object)
+
+    def __init__(
+        self,
+        path: str,
+        parent: Union[QTreeWidget, QTreeWidgetItem],
+        callback: Callable[[List[str], Union[QTreeWidget, QTreeWidgetItem]], None]
+    ):
+        """
```
kangz12345: Since this overrides the `QThread.__init__()` method and calls `super().__init__()`, please write "Extended.".

kangz12345: In addition, write the docstrings for `path` and `parent`, since they are not in `QThread.__init__()` (`parent` seems confusing here as well).
You can just say "See the attributes section in FileFinderThread".

iquip/apps/explorer.py
```diff
@@ -31,6 +30,45 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+        Args:
+            callback: The callback method called after this thread is finished.
+        """
+        super().__init__()
+        self.path = path
+        self.parent = parent
+        self.finished.connect(callback)
+
+    def run(self):
```
kangz12345, line 64: You are overriding [`QThread.run()`](https://doc.qt.io/qt-5/qthread.html#run), so please write "Overridden.".

iquip/apps/explorer.py
```diff
@@ -31,6 +30,45 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+        parent: The widget corresponding to the path.
+    """
+
+    finished = pyqtSignal(list, object)
+
+    def __init__(
+        self,
+        path: str,
+        parent: Union[QTreeWidget, QTreeWidgetItem],
+        callback: Callable[[List[str], Union[QTreeWidget, QTreeWidgetItem]], None]
```
kangz12345: I think the return type could be `Any`.

iquip/apps/explorer.py
```diff
@@ -31,6 +30,45 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+    ):
+        """
+        Args:
+            callback: The callback method called after this thread is finished.
+        """
+        super().__init__()
+        self.path = path
+        self.parent = parent
+        self.finished.connect(callback)
```
kangz12345: Is there a reason to connect the callback here in `FileFinderThread`? I think it can be connected outside, i.e., by the _user_ of this class.

iquip/apps/explorer.py
```diff
@@ -31,6 +30,45 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+
+    Signals:
+        finished(experimentList, parent): Fetching the file list is finished.
+
+    Attributes:
+        path: The path of the directory to search experiment files.
+        parent: The widget corresponding to the path.
+    """
+
+    finished = pyqtSignal(list, object)
```
kangz12345: There is already a signal with the same name: [`QThread.finished()`](https://doc.qt.io/qt-5/qthread.html#finished)

iquip/apps/explorer.py
```diff
@@ -56,9 +94,12 @@ def loadFileTree(self):
...
         """
         self.explorerFrame.fileTree.clear()
-        threading.Thread(
-            target=lambda: self._addFile(self.repositoryPath, self.explorerFrame.fileTree)
-        ).start()
+        self.thread = FileFinderThread(
+            self.repositoryPath,
+            self.explorerFrame.fileTree,
+            self._addFile
+        )
```
kangz12345, line 107: Please add the `parent=self` argument for a `QObject`.
See also: https://doc.qt.io/qt-5/qthread.html#details

iquip/apps/explorer.py
```diff
@@ -113,19 +156,20 @@ def openExperiment(self):
         experimentFileItem = self.explorerFrame.fileTree.currentItem()
         if experimentFileItem.childCount():
```
kangz12345: Hmm.. this becomes now a problem, since during a directory is loaded, the widget item will have no child.

iquip/apps/explorer.py
```diff
@@ -113,19 +156,20 @@ def openExperiment(self):
...
-        path = experimentFileItem.text(0)
+        path = [experimentFileItem.text(0)]
         while experimentFileItem.parent():
             experimentFileItem = experimentFileItem.parent()
-            path = posixpath.join(experimentFileItem.text(0), path)
-        path = posixpath.join(self.repositoryPath, path)
+            path.append(experimentFileItem.text(0))
+        path.append(self.repositoryPath)
+        path = posixpath.join(*reversed(path))
         return path
```
kangz12345: Actually I prefer not to bind a new value of different types to a variable 😅
Could you just return it directly or change the variable name?
For example:
```python
return posixpath.join(*reversed(path))
```
or
```python
paths = [...]
...
path = ...
return path
```

kangz12345 (changes requested):

kangz12345:
Oh, and GitHub only detects A if you write "This closes A, B and C".
See also: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

BECATRUE:
> Oh, and GitHub only detects A if you write "This closes A, B and C". See also: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Wow, thanks for notifying it! I will update the PR description.

iquip/apps/explorer.py
BECATRUE: Oh, I looked it up and I think you're right!
I'm not good at English.. so please understand me😂

iquip/apps/explorer.py
BECATRUE: Currently, the `callback` is always set as `_addFile()` and it does not return anything, so I set the return type as `None`.
Nevertheless, if you still want to fix it, I will do so.

iquip/apps/explorer.py
BECATRUE: I thought about this part, too.
The reason why I did so is that `FileFinderThread` instances are created twice and I wanted to remove the duplicated parts as much as possible.

iquip/apps/explorer.py
BECATRUE: Yes! Actually, I intended it.

Probably, the `finished` signal was created with the aim of announcing the end of the thread.
I needed the similar signal which has additional arguments, so I named it the same.

But, if you want to remain the original signal, I will rename it lik `fetched`.

iquip/apps/explorer.py
BECATRUE: Oh, I was completely missing it. I will add the argument.

iquip/apps/explorer.py
BECATRUE: Ah😃 Then, I will return it directly.
Also, I think it would be better to change the former `path` to `paths` since it contains lots of the paths (even if parts).

iquip/apps/explorer.py
BECATRUE: Umm... it is a critical issue.

Then, how about not checking at all?
Actually, even if we completely check that the selected element is a file, we should re-check whether the file is an ARTIQ experiment file later.
This will be processed through an ARTIQ command line (probably, customed), so why don't we postpone it to this time?

BECATRUE: If you want to check it in this branch, we can do this through comparing argumetns of `self.thread` to `experimentFileItem`.

iquip/apps/explorer.py
kangz12345: Oh, if it is only for the internal use, it would be okay. However, since you named it as a public class, I think it should be considered as a part of the public API.
For the public API, please make it general as possible!

Or you can make it an internal class `_FileFinderThread`.

iquip/apps/explorer.py
kangz12345: Oh I see.
A better design would be separating the `FileFinderThread` instantiation part as a method and calling it twice, not doing it in the `FileFinderThread`.

However, if you think it is only for the internal use, you can name it e.g. `_FileFinderThread` to tell that it is an internal class.

iquip/apps/explorer.py
kangz12345: Hmm, however, `finished` signal is emitted when a `QThread` finishes executing, after stopping the event loop.
I am not sure if it is safe to overwrite such a private signal which already has its own role (since Python does not support overloading).
If we don't have to name it `finished`, I would prefer to use another name!

iquip/apps/explorer.py
kangz12345: As you said, later we have to check more things. I think it is fine not to check at all for now!

iquip/apps/explorer.py
BECATRUE: I intended to use the thread class only internally, so I will make it private.

iquip/apps/explorer.py
BECATRUE: Okay! Then, I will change the name into another one.

iquip/apps/explorer.py
```diff
@@ -30,43 +30,48 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+        widget: Union[QTreeWidget, QTreeWidgetItem],
+        callback: Callable[[List[str], Union[QTreeWidget, QTreeWidgetItem]], None],
+        parent: Optional[QObject] = None
     ):
-        """
+        """Extended.
+
         Args:
             callback: The callback method called after this thread is finished.
+            See the attributes section in _FileFinderThread.
```
kangz12345: To be clear, please write it like:
```python
"""
...
path, widget: See the ...
callback: ...
"""

File: iquip/apps/explorer.py [hunk 21]
```diff
@@ -130,7 +137,7 @@ def _addFile(self, experimentList: List[str], parent: Union[QTreeWidget, QTreeWi
 
         Args:
             experimentList: The list of files under the parent path.
-            parent: See FileFinderThread class.
+            parent: See _FileFinderThread class.
```
kangz12345: Since we changed `parent` -> `widget`, it should be applied here as well.

iquip/apps/explorer.py
```diff
@@ -30,43 +30,48 @@ def __init__(self, parent: Optional[QWidget] = None):
...
             callback: The callback method called after this thread is finished.
+            See the attributes section in _FileFinderThread.
         """
-        super().__init__()
+        super().__init__(parent=parent)
         self.path = path
-        self.parent = parent
-        self.finished.connect(callback)
+        self.widget = widget
+        self.fetched.connect(callback)
```
kangz12345: To be clear, let's give an explicit connection option:
```suggestion
        self.fetched.connect(callback, type=Qt.QueuedConnection)
```

kangz12345 (changes requested):

iquip/apps/explorer.py
BECATRUE: Oh, I thought remaining it as `parent` doesn't make a confusion in this method, so I left it as it is.
But if you want to apply it, I will do that.

iquip/apps/explorer.py
BECATRUE: It seems to show our purpose well!

iquip/apps/explorer.py
kangz12345: It makes sense, but you should write "See _FileFinderThread.widget." since its `parent` refers to a different thing.
However, now you changed the argument name, so it doesn't matter!

kangz12345 (approved):
That was a lot of work, cheers!

PR #35, opened 2023-06-08
Unit tests for explorer

### LINKED ISSUE #29: Unit tests for `apps.explorer`
Please make unit tests for `apps.explorer`.

BECATRUE:
Thie closes #29.

I implemented the test codes for `iquip.apps.explorer` module.
Even if `ExplorerApp._addFile()` and `_FileFinderThread` are private, they play an important role so I tested them.

For test, enter the following command line.
```shell
$ coverage run -m unittest tests/test_explorer.py
$ coverage report -m iquip/apps/explorer.py
```


tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
+"""Unit tests for explorer module."""
+
+import unittest
+from unittest import mock
+from importlib import reload
```
kangz12345: Please do not import a function directly:
```python
import importlib

importlib.reload(...)
```

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+
+from PyQt5.QtCore import QObject, Qt
+from PyQt5.QtWidgets import QApplication, QTreeWidgetItem
+from PyQt5.QtTest import QTest
+
+from iquip.apps import explorer
+
+class ExplorerFrameTest(unittest.TestCase):
```
kangz12345: I just realized that `ExplorerFrame` has no attribute docstring! Could you make it in this PR?

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+    """Unit tests for ExplorerFrame class."""
+
+    def setUp(self):
+        self.qapp = QApplication([])
+        explorer._FileFinderThread = mock.MagicMock()
+
+    def tearDown(self):
+        del self.qapp
+        reload(explorer)  # Restore explorer module.
```
kangz12345: Actually, you can use the [`patcher`](https://docs.python.org/3.7/library/unittest.mock.html#patch-methods-start-and-stop) object.
For example, 
```python
def setUp(self):
    ...
    self.patcher = mock.patch("explorer._FileFinderThread")
    self.MockedFileFinderThread = self.patcher.start()

def tearDown(self):
    ...
    self.patcher.stop()
```
Or, as you can see in the reference example, using [`addCleanup()`](https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.addCleanup) seems safer:
```python
def setUp(self):
    ...
    patcher = mock.patch("explorer._FileFinderThread")
    self.MockedFileFinderThread = patcher.start()
    self.addCleanup(patcher.stop)
```

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def test_file_tree_item_expanded(self):
+        explorer.ExplorerApp.lazyLoadFile = mock.MagicMock()
```
kangz12345: When you temporarily substitute an attribute or a method, please use patchers instead.
1. If you should patch it throughout the test method, use the decorator.
2. If you can patch it only in a part of codes, use the context manager.

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        directoryItem = QTreeWidgetItem(app.explorerFrame.fileTree)
+        directoryItem.setText(0, "directory")
+        QTreeWidgetItem(directoryItem)  # Add an empty item to an unloaded directory.
+        directoryItem.setExpanded(True)
+        app.lazyLoadFile.assert_called_once()  # pylint: disable=no-member
```
kangz12345: I think this is not a unit test for `ExplorerFrame`.
To be honest, I think `ExplorerFrame` is too simple so that we don't need any unit test for it 😂

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        QTest.mouseClick(app.explorerFrame.reloadButton, Qt.LeftButton)
+        # Once when the app is created, once explicitly.
+        self.assertEqual(app.loadFileTree.call_count, 2)  # pylint: disable=no-member
+
+    def test_open_button_clicked(self):
+        explorer.ExplorerApp.openExperiment = mock.MagicMock()
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        QTest.mouseClick(app.explorerFrame.openButton, Qt.LeftButton)
+        app.openExperiment.assert_called_once()  # pylint: disable=no-member
```
kangz12345: These seems not unit tests as well, but it could be good to have such GUI-behavior tests though..

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def tearDown(self):
+        del self.qapp
+        reload(explorer)  # Restore explorer module.
+
+    def test_init_thread(self):
+        widget = QTreeWidgetItem()
+        callback = mock.MagicMock()
+        parent = QObject()
+        thread = explorer._FileFinderThread(path="path", widget=widget, \
```
kangz12345: I think you don't need that backslash.

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+                                            callback=callback, parent=parent)
+        self.assertEqual(thread.path, "path")
+        self.assertEqual(thread.widget, widget)
```
kangz12345: You can use the patcher to check if the `connect()` is called with the `callback` as well:
```python
with mock.patch("explorer._FileFinderThread.fetched") as mocked_fetched:
    thread = ...
    mocked_fetched.connect.assert_called_once_with(callback, type=Qt.QueuedConnection)
```
Maybe it will work, but I'm not quite sure... 😄

tests/test_explorer.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def test_run(self):
+        explorer._FileFinderThread.fetched = mock.MagicMock()
+        explorer.cmdtools.run_command = mock.MagicMock()
```
kangz12345: Please use the patch features instead!

kangz12345 (changes requested):
Please use patch features first. I will continue reviewing after it!
Moreover, how about adding such non-unit tests to check behaviors?
Or, one can say those are also unit tests since it tests a single module..

BECATRUE:
> Please use patch features first. I will continue reviewing after it!

What do you want to patch?

I also want to use a patch block such as `with mock.patch.multiple()`, but in my case, most mocking must be done before an instance is created (e.g., `explorer.ExplorerApp.lazyLoadFile()` in `test_file_tree_item_expanded()`), so I think it cannot use a patch.

If there's anything I don't know, feel free to let me know.

BECATRUE:
> > Please use patch features first. I will continue reviewing after it!
> 
> What do you want to patch?
> 
> I also want to use a patch block such as `with mock.patch.multiple()`, but in my case, most mocking must be done before an instance is created (e.g., `explorer.ExplorerApp.lazyLoadFile()` in `test_file_tree_item_expanded()`), so I think it cannot use a patch.
> 
> If there's anything I don't know, feel free to let me know.

If you meant just to mock everything using a patch, I'll change it as it is.

BECATRUE:
> Moreover, how about adding such non-unit tests to check behaviors?
Or, one can say those are also unit tests since it tests a single module..

What do you want to test? Please give me an example!

tests/test_explorer.py
BECATRUE: Wow, you are so meticulous! I will do so😃

tests/test_explorer.py
BECATRUE: Oh, you already gave me a solution in here! Please ignore the below question about how to use patch..
I will modify the mocking parts referred to your suggestion.

tests/test_explorer.py
BECATRUE: As you said above, I will apply `patch`!

tests/test_explorer.py
BECATRUE: Umm.. You are right. Then, how about including tests for the frame in the app?
The frame is independent to the app, but in order to run well, the app is necessary.
So I want to move several tests such as expanding a directory into the app test.

Please leave your opinion!

tests/test_explorer.py
BECATRUE: Oh, you're right! I modified as you said and it runs correctly.

BECATRUE:
I applied your reviews, but I left the part of testing `ExplorerFrame`.
After receiving your opinion, I will modify it.

tests/test_explorer.py
kangz12345: But the problem is that tests like `test_file_tree_item_expanded()` are not _unit tests_ in our sense, since unit tests should only depend on the target function, method, class, etc.

We can call those tests _functional tests_ and it seems quite important in a sense that we should verify the API specification.

I am not quite sure yet, whether to implement the functional tests along with the unit tests or to separate them.

tests/.pylintrc
```diff
@@ -3,7 +3,7 @@ extension-pkg-whitelist=PyQt5
 
 [BASIC]
 argument-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9]*)*))$
-attr-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9]*)*))$
+attr-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9]*)*)|(?:[A-Z][a-z0-9]+(?:[A-Z][a-z0-9]*)*))$
```
kangz12345: 😂 Oh..
What would you prefer: `self.MockedThread` vs. `self.mocked_thread_cls`?

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        del self.qapp
+
+    @mock.patch("iquip.apps.explorer.ExplorerApp.lazyLoadFile")
+    def test_file_tree_item_expanded(self, mockedLazyLoadFile):
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        directoryItem = QTreeWidgetItem(app.explorerFrame.fileTree)
+        directoryItem.setText(0, "directory")
+        QTreeWidgetItem(directoryItem)  # Add an empty item to an unloaded directory.
+        directoryItem.setExpanded(True)
+        mockedLazyLoadFile.assert_called_once()  # pylint: disable=no-member
```
kangz12345: Now `mockedLazyLoadFile` is a mock object, so you can remove the pylint comment.

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        self.qapp = QApplication([])
+        patcher = mock.patch("iquip.apps.explorer._FileFinderThread")
+        self.MockedFileFinderThread = patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def tearDown(self):
+        del self.qapp
+
+    @mock.patch("iquip.apps.explorer.ExplorerApp.lazyLoadFile")
+    def test_file_tree_item_expanded(self, mockedLazyLoadFile):
```
kangz12345: We have to decide where to put these _functional tests_.

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        with mock.patch("iquip.apps.explorer._FileFinderThread.fetched") as mockedFetched:
+            thread = explorer._FileFinderThread(path="path", widget=widget,
+                                                callback=mock.MagicMock(), parent=parent)
+            thread.run()
+            thread.wait()
+            mocked_run_command.assert_called_once_with("artiq_client ls path")
+            mockedFetched.emit.assert_called_once_with(["path1", "path2"], widget)
+
+
+class ExplorerAppTest(unittest.TestCase):
```
kangz12345, line 46: I found that `ExplorerApp` also does not have the docstring for attributes. Please add it!

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        app = explorer.ExplorerApp(name="name")
+        self.assertEqual(app.repositoryPath, "./repository")
+
+    def test_load_file_tree(self):
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        QTreeWidgetItem(app.explorerFrame.fileTree)  # Add a dummy item to the file tree.
+        app.loadFileTree()
+        self.assertEqual(app.explorerFrame.fileTree.topLevelItemCount(), 0)
+        # Once when the app is created, once explicitly.
+        self.assertEqual(explorer._FileFinderThread.call_count, 2)
```
kangz12345: I would prefer to use `self.MockedFileFinderThread` instead.

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        fileItem = QTreeWidgetItem(app.explorerFrame.fileTree)
+        fileItem.setText(0, "file")
+        app.lazyLoadFile(fileItem)
+        explorer._FileFinderThread.assert_called_once()  # Once when the app is created.
+
+    def test_add_file(self):
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        parent = QTreeWidgetItem(app.explorerFrame.fileTree)
+        experimentList = ["directory/", "_hidden_directory/", \
```
kangz12345: Backslash is not necessary.

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+                          "experiment_file.py", "_hidden_file.py", "file.dummy"]
+        app._addFile(experimentList, parent)
+        self.assertEqual(parent.childCount(), 2)
```
kangz12345: Please leave a comment about why it should be 2, e.g., `# "directory/" and "experiment_file.py"`.

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        directoryItem = parent.child(0)
+        self.assertEqual(directoryItem.text(0), "directory")
+        self.assertEqual(directoryItem.childCount(), 1)
+        fileItem = parent.child(1)
+        self.assertEqual(fileItem.text(0), "experiment_file.py")
```
kangz12345: Good check! How about testing if `fileItem.childCount()` is 0?

tests/test_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def test_open_experiment(self):
+        app = explorer.ExplorerApp(name="name", masterPath="masterPath", parent=QObject())
+        item = QTreeWidgetItem(app.explorerFrame.fileTree)
+        app.explorerFrame.fileTree.setCurrentItem(item)
+        with mock.patch.object(app, "fullPath") as mockedFullPath:
+            app.openExperiment()
+            mockedFullPath.assert_called_with(item)
+
+    def test_full_path(self):
```
kangz12345, line 123: 👍

kangz12345 (changes requested):

tests/test_explorer.py
BECATRUE: Ah, I understood what you said. I agree with you and can I remove all tests for `ExplorerFrame`?

tests/.pylintrc
BECATRUE: I inteded to meet the PyQt class name convention, but I don't like it more that I changed the regex rule..
Thus, I will adopt the latter!

tests/test_explorer.py
kangz12345: Okay, then let's remove them in this PR, and we can talk about the functional tests later.

tests/test_explorer.py
BECATRUE: Oh, thanks for notifying it! But the tests will be removed..😅

tests/test_explorer.py
BECATRUE: I looked for several solutions:
1. Use `unittest` similar to the current tests.
Even if the module name is __unittest__, but it can also tests functional tests.
2. Use [pytest](https://docs.pytest.org/en/7.3.x/)
It is one of the most powerful test packages. Probably it offers more test methods.
But I don't like using various kinds of packages...

Then, How about using `unittest` and making functional tests in another file such as `test_explorer_function.py`?

BECATRUE: Or I think implementing functional tests in this file is not bad..

tests/test_explorer.py
BECATRUE: Oh, I added only the docstring about `ExplorerFrame`.. I will modify it.

tests/test_explorer.py
BECATRUE: It looks good! I will add it.

kangz12345 (approved):
LGTM!

Issue #38, opened 2023-06-17
Functional tests for `apps.explorer`


BECATRUE:
We will implement the functional tests for the explorer GUI.

PR #40, opened 2023-06-20
Replace `cmdtools` module to `requests`

### LINKED ISSUE #37: Remove `cmdtools`
We decided to remove the `cmdtools` module as we are not going to use the artiq client directly.
The [snu-quiqcl/artiq-proxy](https://github.com/snu-quiqcl/artiq-proxy) repository will provide an artiq-independent interface for queries, as mentioned in #22.

BECATRUE:
This will close #37.

I removed the `cmdtools` module and use the `requests` module to request to the proxy server.
Even if the `requests` module is external, in the official [documentation](https://docs.python.org/3/library/http.client.html), it recommends this module, so I chose it.

Also, as snu-quiqcl/artiq-proxy#14, we added the `repository_path` to the proxy server internally, so I modified it.

Now, the unittest would be failed. I will modify it in #38.


iquip/apps/explorer.py
```diff
@@ -69,34 +68,34 @@ def __init__(
...
-        experimentList = experimentList.split("\n")[:-1]  # The last one is always an empty string.
+        try:
+            response = requests.get("http://127.0.0.1:8000/ls/",
+                                    params={"directory": self.path},
+                                    timeout=10)
+            response.raise_for_status()
+            experimentList = response.json()
+        except requests.exceptions.RequestException:
+            pass
         self.fetched.emit(experimentList, self.widget)
```
kangz12345, line 86: This code is executed regardless if the get request succeeds.

iquip/apps/explorer.py
```diff
@@ -69,34 +68,34 @@ def __init__(
...
-        experimentList = cmdtools.run_command(f"artiq_client ls {self.path}").stdout
-        experimentList = experimentList.split("\n")[:-1]  # The last one is always an empty string.
+        try:
+            response = requests.get("http://127.0.0.1:8000/ls/",
+                                    params={"directory": self.path},
+                                    timeout=10)
+            response.raise_for_status()
+            experimentList = response.json()
+        except requests.exceptions.RequestException:
+            pass
```
kangz12345: Even though we are not logging by now, but let's temporarily print something about the exception information.

kangz12345 (changes requested):

iquip/apps/explorer.py
BECATRUE: Oh, you're right! I will add a printed message and return here in the handling part.

kangz12345 (approved):
LGTM

Issue #42, opened 2023-06-21
Implement basic functions of the builder


BECATRUE:
### Feature you want to implement
I want to add basic functions of the builder app.

As I wrote in #6, there are 2 goals.
- Able to submit the experiment to the scheduler.
- Able to edit the build params.

Also, now we can open the selected experiment when one is clicked. (See #10)

In the artiq dashboard, there are several other functions such as setting a due date, priority, and pipeline, but these are not for this issue.

#### Submit the experiment to the scheduler
We already implemented submitting an experiment in snu-quiqcl/artiq-proxy#6.
Thus, we just need to request to the proxy server.

#### Edit the build params
There are 5 types of arguments: `PYONValue`, `BooleanValue`, `EnumerationValue`, `NumberValue`, and `StringValue`.
For details, please see [reference](https://m-labs.hk/artiq/manual/core_language_reference.html?highlight=string#module-artiq.language.environment) and [GitHub](https://github.com/m-labs/artiq/blob/master/artiq/language/environment.py).

In this issue, I will handle all types expect for `PYONValue`. (In fact, I don't know well where it is used.)
- `BooleanValue`: The parameter is only the default value, such as `BooleanValue(default=False)`.
This will be implemented using `QCheckBox`.
- `EnumerationValue`: It supports that we can choose only one of pre-defined candidates.
The parameters are the options and default value, such as `EnumerationValue(choices=["r", "g", "b"], default="r")`.
This will be implemented using `QComboBox`
- `NumberValue`: It supports that we can choose a value in pre-defined range and resolution.
There are lots of parameters, so please refer to [reference](https://m-labs.hk/artiq/manual/core_language_reference.html?highlight=string#artiq.language.environment.NumberValue).
- `StringValue`: The parameter is only the default value, similar to `BooleanValue`.
This will be implemented using `QLineEdit`.

We already implemented getting arguments from a specific experiment in snu-quiqlc/artiq-proxy#5.
Thus, we will use it to get the type and default value of each argument.

PR #43, opened 2023-06-22
Able to open the builder and submit the experiment

BECATRUE:
This is related to #42.

First, I'm sorry I have a lot of reviews😂

I implemented three things:
1. Open the builder of the selected experiment (`explorer.py`)
2. Submit the experiment _with the default arguments_ (`builder.py`)
3. For simplicity, define the experiment info format, similar to `ExperimentInfo` in artiq-proxy (`protocols.py`)

### Open the builder
First of all, we need to check if the selected item is a valid experiment file.
We can check it through a request to `/experiment/info/`.
If the item is a directory or a non-experiment file, it returns a 500 Server error or an empty directory, respectively.
This function is implemented in `apps.builder.ExperimentInfoThread`.

After checking it, the explorer app requests to open the builder using `qiwiscall`.

### Submit the experiment
The builder app knows all arguments from the `qiwiscall`, so it doesn't need to request to the proxy server.
To submit it, I implemented `apps.builder.ExperimentSubmitThread` and it shows the rid after submitting.

### Result
- Try to open the builder

- Submit the experiment

### Not implemented (TODO)
1. Show the basic info such as the experiment path and the class name.
2. Show the build arguments and edit them
4. Set the scheduling arguments such as a due date and a priority.

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -1,9 +1,10 @@
 """App module for editting the build arguments and submitting the experiment."""
 
+import json
 from typing import Any, Callable, Dict, Optional, Tuple
 
 import requests
-from PyQt5.QtCore import QObject, Qt, QThread, pyqtSignal
+from PyQt5.QtCore import QObject, Qt, QThread, pyqtSignal, pyqtSlot
 from PyQt5.QtWidgets import (
     QPushButton, QVBoxLayout, QWidget
 )
@@ -100,7 +101,7 @@ class ExperimentSubmitThread(QThread):
         self,
         experimentPath: str,
         experimentArgs: Dict[str, Any], 
-        callback,
+        callback: Callable[[int], None],
         parent: Optional[QObject] = None
     ):
         """Extended.
@@ -125,7 +126,7 @@ class ExperimentSubmitThread(QThread):
         try:
             params = {
                 "file": self.experimentPath,
-                "args": self.experimentArgs
+                "args": json.dumps(self.experimentArgs)
             }
             response = requests.get("http://127.0.0.1:8000/experiment/submit/",
                                     params=params,
@@ -150,7 +151,7 @@ class BuilderApp(qiwis.BaseApp):
         name: str,
         experimentPath: str,
         experimentClsName: str,
-        experimentInfo: ExperimentInfo,
+        experimentInfo: Dict[str, Any],
         parent: Optional[QObject] = None
     ):
         """Extended.
@@ -158,10 +159,32 @@ class BuilderApp(qiwis.BaseApp):
         Args:
             experimentPath: The path of the experiment file.
             experimentClsName: The class name of the experiment.
-            experimentInfo: The experiment information. See protocols.ExperimentInfo.
+            experimentInfo: The experiment information, a dictionary of protocols.ExperimentInfo.
         """
         super().__init__(name, parent=parent)
+        self.experimentPath = experimentPath
+        self.experimentClsName = experimentClsName
+        self.experimentInfo = ExperimentInfo(**experimentInfo)
         self.builderFrame = BuilderFrame()
+        # connect signals to slots
+        self.builderFrame.submitButton.clicked.connect(self.submit)
+
+    @pyqtSlot()
+    def submit(self):
+        experimentArgs = {
+            argName: argInfo[0]["default"]
+            for argName, argInfo in self.experimentInfo.arginfo.items()
+        }
+        self.thread = ExperimentSubmitThread(
+            self.experimentPath,
+            experimentArgs,
+            self.onSubmitted,
+            self
+        )
+        self.thread.start()
+
+    def onSubmitted(self, rid: int):
+        print(f"RID: {rid}")
 
     def frames(self) -> Tuple[BuilderFrame]:
         """Overridden."""
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -171,6 +171,12 @@ class BuilderApp(qiwis.BaseApp):
 
     @pyqtSlot()
     def submit(self):
+        """Submits the experiment with the build arguments.
+        
+        Once the submitButton is clicked, this is called.
+
+        TODO(BECATRUE): Apply the editted arguments. It will be implemented in Basic Runner project.
+        """
         experimentArgs = {
             argName: argInfo[0]["default"]
             for argName, argInfo in self.experimentInfo.arginfo.items()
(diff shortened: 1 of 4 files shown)


BECATRUE:
Currently, the unit tests will be failed becuase `openExperiment()` uses a new thread and I didn't handle it in the test.
Please ignore it now..

iquip/apps/builder.py
```diff
@@ -0,0 +1,206 @@
...
+from PyQt5.QtCore import QObject, Qt, QThread, pyqtSignal, pyqtSlot
+from PyQt5.QtWidgets import (
+    QPushButton, QVBoxLayout, QWidget
+)
+
+import qiwis
+from iquip.protocols import ExperimentInfo
+
+class BuilderFrame(QWidget):
+    """Frame for showing the build arguments and requesting to submit it."""
```
kangz12345: Attributes docstring!

iquip/apps/builder.py
```diff
@@ -0,0 +1,206 @@
...
+            print(err)
+            return
+        if data:
+            experimentClsName = next(iter(data))
+            experimentInfo = data[experimentClsName]
+            self.fetched.emit(
+                self.experimentPath,
+                experimentClsName,
+                ExperimentInfo(**experimentInfo)
+            )
```
kangz12345: `else:` clause would be helpful to print out that we got an empty dictionary.

iquip/apps/builder.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class ExperimentSubmitThread(QThread):
+    """QThread for submitting the experiment with its build arguments.
+    
+    Signals:
+        submitted(rid): The experiment is submitted.
```
kangz12345: Please explain what the `rid` parameter is.

iquip/apps/builder.py
```diff
@@ -0,0 +1,206 @@
...
+        self.experimentPath = experimentPath
+        self.experimentArgs = experimentArgs
+        self.submitted.connect(callback, type=Qt.QueuedConnection)
+
+    def run(self):
+        """Overridden.
+        
+        Submits the experiment to the proxy server.
+
+        Regardless of whether it is successful or not, the server returns the run identifier.
```
kangz12345: Whether _what_ is successful or not?

iquip/apps/builder.py
```diff
@@ -0,0 +1,206 @@
...
+        After finished, the submitted signal is emitted.
```
kangz12345: After _what_ finishes?

iquip/apps/builder.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        try:
+            params = {
+                "file": self.experimentPath,
+                "args": json.dumps(self.experimentArgs)
+            }
```
kangz12345, line 76: I think this should be outside this `try:` block.
Maybe you should use another `try:` block and catch the exception from `json.dumps()`.

iquip/apps/explorer.py
```diff
@@ -10,6 +10,8 @@
 )
 
 import qiwis
+from iquip.protocols import ExperimentInfo
+from iquip.apps.builder import ExperimentInfoThread
```
kangz12345: If it is used here, why did you define `ExperimentInfoThread` in `builder` module?

kangz12345 (changes requested):

iquip/apps/explorer.py
BECATRUE: I think that fetching the experiment info is a part of `builder` module becuase it is a step of build.
Of course, we will use the thread in `builder` to edit and update the arguments.

Thus, I added it in `builder` module, not `explorer` module.

iquip/apps/builder.py
BECATRUE: Oh, it means an experiment operation, not submission! I will modify it.

BECATRUE: For example, if there is an invalid build argument, the experiment will be failed but the submission works well.

iquip/apps/builder.py
BECATRUE: It is also submission. I will modify it.

iquip/apps/builder.py
BECATRUE: It looks better!

iquip/apps/builder.py
```diff
@@ -120,14 +126,18 @@ def run(self):
         
         Submits the experiment to the proxy server.
 
-        Regardless of whether it is successful or not, the server returns the run identifier.
-        After finished, the submitted signal is emitted.
+        If only the experiment was submitted well regardless of whether it runs successfully or not,
```
kangz12345: I think "Whenever ..." is more appropriate. We use "If only ..." for the assumption expressions (of hopes).

iquip/apps/builder.py
```diff
@@ -120,14 +126,18 @@ def run(self):
...
+        the server returns the run identifier.
+        After submitted, the submitted signal is emitted.
         """
         try:
             params = {
                 "file": self.experimentPath,
                 "args": json.dumps(self.experimentArgs)
             }
+        except TypeError:
+            print("Failed to convert the build arguments to a JSON string.")
```
kangz12345, line 78: Maybe it should return here.

iquip/apps/explorer.py
```diff
@@ -18,7 +18,7 @@ class ExplorerFrame(QWidget):
 
     Attributes:
         fileTree: The tree widget for showing the file structure.
-        reloadButton: The buttont for reloading the fileTree.
+        reloadButton: The button for reloading the fileTree.
```
kangz12345, line 21: 😂 👍

kangz12345 (changes requested):

iquip/apps/explorer.py
kangz12345: Hmm.. I think we should keep each app decoupled if possible.
How about passing the full path to the builder constructor and the builder runs the thread to fetch the arguments, not the explorer?

iquip/apps/builder.py
BECATRUE: It looks better! I will apply it.

BECATRUE:
Several commits are twisted while handling them.. I checked as much as possible and pushed forcely, but please check it😥

iquip/apps/explorer.py
BECATRUE: Umm... I originally planned as you suggested, but there is a problem.
If we try to open a non-experiment file, we can only check that it is invalid after finishing `ExperimentInfoThread`.
Thus, we implemented as you said, a builder would also be created for an invalid file, and a separate process must be carried out to remove it.

If you're worried about coupling, how about moving the thread to another file such as `iquip/apps/experiment_info_thread.py`?

kangz12345: Okay then, but could you explain why `builder` needs the thread? I think all the necessary information is passed when constructing a builder app.

BECATRUE: It is just for reloading the build arguements!
In the artiq dashboard, there is the function because the experiment file can be modified during the builder is open and it is inconvenient to close and reopen.

I will implement the function.

BECATRUE:
I moved the thread to another file `iquip/apps/thread.py`. @kangz12345

iquip/apps/explorer.py
kangz12345: I see, then the both apps need the same feature (information fetch).
For this PR, let's go with your solution.
Maybe we can develop the `thread` further to help such thread definitions.

kangz12345 (approved):
LGTM

PR #47, opened 2023-06-25
Add the build argument entries

### LINKED ISSUE #42: Implement basic functions of the builder

BECATRUE:
This PR will close #42.

First of all, please check and merge the PR #43.

I implmented the left parts of #42:
1. Create the build arguments entries.
2. Submit the experiment with the editted arguments.

### Create the build arguments entries
Each entry is an instance of `_BaseEntry` and has the `nameLabel` to show the argument name.
Each entry type is implemented as a subclass of `_BaseEntry` and it has an appropriate widget such as `QSpinBox`.

### Submit the experiment with the editted arguments
Now, we can submit it with the _actual_ arguments not the default values using `_BaseEntry.value()`.

### Result

If you want to test more various experiments, I can offer you some tips!

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -133,7 +133,7 @@ class _NumberEntry(_BaseEntry):
             step: The step between values changed by the up and down button.
             min: The minimum value. None for no minimum.
             max: The maximum value. None for no maximum.
-            ndecimals: The maximum number of decimals.
+            ndecimals: The number of displayed decimals.
             default: The default value. If None, it is set to the min value.
             scale, type: See the attributes section in _NumberEntry.
         """
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -50,18 +50,19 @@ class _BaseEntry(QWidget):
 
 
 class _BooleanEntry(_BaseEntry):
-    """Entry class for a boolean value."""
-
-    def __init__(self, name: str, default: bool = False, parent: Optional[QWidget] = None):
-        """Extended.
-        
-        Args:
+    """Entry class for a boolean value.
+    
+    Attributes:
+        argInfo: Each key and its value are:
             default: The default value.
-        """
-        super().__init__(name, parent=parent)
+    """
+
+    def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget] = None):
+        """Extended."""
+        super().__init__(name, argInfo, parent=parent)
         # widgets
         self.checkBox = QCheckBox(self)
-        self.checkBox.setChecked(default)
+        self.checkBox.setChecked(self.argInfo["default"])
         # layout
         self.layout.addWidget(self.checkBox)
 
@@ -312,6 +313,7 @@ class BuilderApp(qiwis.BaseApp):
                 "EnumerationValue": _EnumerationEntry,
                 "NumberValue": _NumberEntry
             }[argInfo.pop("ty")]
+            print(argName)
             widget = entryCls(argName, argInfo)
             item = QListWidgetItem(self.builderFrame.argsListWidget)
             item.setSizeHint(widget.sizeHint())
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -109,7 +109,7 @@ class _NumberEntry(_BaseEntry):
     """Entry class for a number value.
     
     Attributes:
-        scale: The scale factor that actually applies.
+        scale: The scale factor that is multiplied to the number value.
         type: The type of the value. If "int", value() returns an integer value.
     """
 
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -6,8 +6,8 @@ from typing import Any, Callable, Dict, List, Optional, Tuple
 import requests
 from PyQt5.QtCore import QObject, Qt, QThread, pyqtSignal, pyqtSlot
 from PyQt5.QtWidgets import (
-    QCheckBox, QComboBox, QHBoxLayout, QLabel, QLineEdit, QListWidget, QListWidgetItem,
-    QPushButton, QVBoxLayout, QWidget
+    QCheckBox, QComboBox, QDoubleSpinBox, QHBoxLayout, QLabel, QLineEdit, QListWidget,
+    QListWidgetItem, QPushButton, QVBoxLayout, QWidget
 )
 
 import qiwis
@@ -85,7 +85,7 @@ class _EnumerationEntry(_BaseEntry):
         
         Args:
             choices: The pre-defined candidates.
-            default: The default value. If it does not exist, it is set to None.
+            default: The default value. If it does not exist, it is set to the first candidate.
         """
         super().__init__(name, parent=parent)
         # widgets
@@ -96,6 +96,58 @@ class _EnumerationEntry(_BaseEntry):
             self.comboBox.setCurrentText(choices[0] if default is None else default)
         # layout
         self.layout.addWidget(self.comboBox)
+
+    def value(self) -> str:
+        """Overridden.
+        
+        Returns the value of the comboBox.
+        """
+        return self.comboBox.currentText()
+
+
+class _NumberEntry(_BaseEntry):
+    """Entry class for a number value.
+    
+    Attributes:
+        scale: The scale factor that actually applies.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        unit: str,
+        scale: float,
+        step: float,
+        min: float,  # pylint: disable=redefined-builtin
+        max: float,  # pylint: disable=redefined-builtin
+        ndecimals: int,
+        type: str,  # pylint: disable=unused-argument,redefined-builtin
+        default: Optional[float] = None,
+        parent: Optional[QWidget] = None
+    ):  # pylint: disable=too-many-arguments
+        """Extended.
+        
+        Args:
+            unit: The unit of the value.
+            scale: See the attributes section in _NumberEntry.
+            step: The step between values changed by the up and down button.
+            min: The minimum value.
+            max: The maximum value.
+            ndecimals: The maximum number of decimals.
+            _type: The type of the value. This is not used.
+            default: The default value. If it does not exist, it is set to the min value.
+        """
+        super().__init__(name, parent=parent)
+        self.scale = scale
+        # widgets
+        self.spinBox = QDoubleSpinBox(self)
+        self.spinBox.setSuffix(unit)
+        self.spinBox.setSingleStep(step / scale)
+        self.spinBox.setRange(min / scale, max / scale)
+        self.spinBox.setDecimals(ndecimals)
+        self.spinBox.setValue((min if default is None else default) / scale)
+        # layout
+        self.layout.addWidget(self.spinBox)
 
     def value(self) -> str:
         """Overridden.
@@ -257,7 +309,8 @@ class BuilderApp(qiwis.BaseApp):
             entryCls = {
                 "BooleanValue": _BooleanEntry,
                 "StringValue": _StringEntry,
-                "EnumerationValue": _EnumerationEntry
+                "EnumerationValue": _EnumerationEntry,
+                "NumberValue": _NumberEntry
             }[argInfo.pop("ty")]
             widget = entryCls(argName, **argInfo)
             item = QListWidgetItem(self.builderFrame.argsListWidget)
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -59,7 +59,7 @@ class _BooleanEntry(_BaseEntry):
         super().__init__(name, parent=parent)
         # widgets
         self.checkBox = QCheckBox(self)
-        self.checkBox.setCheckState(default)
+        self.checkBox.setChecked(default)
         # layout
         self.layout.addWidget(self.checkBox)
 
@@ -68,7 +68,7 @@ class _BooleanEntry(_BaseEntry):
         
         Returns the status of the checkBox.
         """
-        return self.checkBox.checkState()
+        return self.checkBox.isChecked()
 
 
 class _EnumerationEntry(_BaseEntry):
@@ -154,7 +154,7 @@ class _NumberEntry(_BaseEntry):
         
         Returns the value of the comboBox.
         """
-        return self.comboBox.currentText()
+        return self.spinBox.value()
 
 
 class _StringEntry(_BaseEntry):
@@ -323,13 +323,12 @@ class BuilderApp(qiwis.BaseApp):
         """Submits the experiment with the build arguments.
         
         Once the submitButton is clicked, this is called.
-
-        TODO(BECATRUE): Apply the editted arguments. It will be implemented in Basic Runner project.
         """
-        experimentArgs = {
-            argName: argInfo[0]["default"]
-            for argName, argInfo in self.experimentInfo.arginfo.items()
-        }
+        experimentArgs = {}
+        for row in range(self.builderFrame.argsListWidget.count()):
+            item = self.builderFrame.argsListWidget.item(row)
+            widget = self.builderFrame.argsListWidget.itemWidget(item)
+            experimentArgs[widget.name] = widget.value()
         self.thread = ExperimentSubmitThread(
             self.experimentPath,
             experimentArgs,
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -23,6 +23,7 @@ class _BaseEntry(QWidget):
         name: The argument name.
         argInfo: The dictionary with the argument options.
         nameLabel: The label for showing the argument name.
+        layout: The QHBoxLayout with the nameLabel.
     """
 
     def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget] = None):
@@ -39,7 +40,6 @@ class _BaseEntry(QWidget):
         # layout
         self.layout = QHBoxLayout(self)
         self.layout.addWidget(self.nameLabel)
-        self.setLayout(self.layout)
 
     def value(self) -> Any:
         """Returns the entered or selected value.
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -1,7 +1,7 @@
 """App module for editting the build arguments and submitting the experiment."""
 
 import json
-from typing import Any, Callable, Dict, List, Optional, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import requests
 from PyQt5.QtCore import QObject, Qt, QThread, pyqtSignal, pyqtSlot
@@ -110,6 +110,7 @@ class _NumberEntry(_BaseEntry):
     
     Attributes:
         scale: The scale factor that actually applies.
+        type: The type of the value. If "int", value() returns an integer value.
     """
 
     def __init__(
@@ -121,7 +122,7 @@ class _NumberEntry(_BaseEntry):
         min: float,  # pylint: disable=redefined-builtin
         max: float,  # pylint: disable=redefined-builtin
         ndecimals: int,
-        type: str,  # pylint: disable=unused-argument,redefined-builtin
+        type: str,  # pylint: disable=redefined-builtin
         default: Optional[float] = None,
         parent: Optional[QWidget] = None
     ):  # pylint: disable=too-many-arguments
@@ -129,16 +130,16 @@ class _NumberEntry(_BaseEntry):
         
         Args:
             unit: The unit of the value.
-            scale: See the attributes section in _NumberEntry.
             step: The step between values changed by the up and down button.
             min: The minimum value.
             max: The maximum value.
             ndecimals: The maximum number of decimals.
-            _type: The type of the value. This is not used.
             default: The default value. If it does not exist, it is set to the min value.
+            scale, type: See the attributes section in _NumberEntry.
         """
         super().__init__(name, parent=parent)
         self.scale = scale
+        self.type = type
         # widgets
         self.spinBox = QDoubleSpinBox(self)
         self.spinBox.setSuffix(unit)
@@ -149,12 +150,13 @@ class _NumberEntry(_BaseEntry):
         # layout
         self.layout.addWidget(self.spinBox)
 
-    def value(self) -> str:
+    def value(self) -> Union[int, float]:
         """Overridden.
         
         Returns the value of the comboBox.
         """
-        return self.spinBox.value()
+        typeCls = int if self.type == "int" else float
+        return typeCls(self.spinBox.value())
 
 
 class _StringEntry(_BaseEntry):
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -43,11 +43,10 @@ class _BooleanEntry(_BaseEntry, QCheckBox):
     If there is no default value, it is set to False.
     """
 
-    def __init__(self, name: str, parent: Optional[QWidget] = None, **kwargs: Any):
+    def __init__(self, name: str, parent: Optional[QWidget] = None, default: bool = False):
         """Extended."""
-        _BaseEntry.__init__(name=name)
-        QCheckBox.__init__(parent=parent)
-        default = kwargs["default", False]
+        _BaseEntry.__init__(self, name=name)
+        QCheckBox.__init__(self, parent=parent)
         self.initEntry(default)
 
     def initEntry(self, default: bool):
@@ -80,6 +79,7 @@ class BuilderFrame(QWidget):
         layout = QVBoxLayout(self)
         layout.addWidget(self.argsListWidget)
         layout.addWidget(self.submitButton)
+        self.setLayout(layout)
 
 
 class ExperimentSubmitThread(QThread):
@@ -185,7 +185,15 @@ class BuilderApp(qiwis.BaseApp):
         Args:
             experimentInfo: The experiment information.
         """
-        print(experimentInfo)
+        for argName, _argInfo in experimentInfo.arginfo.items():
+            argInfo = _argInfo[0]  # All the remaining elements are None.
+            entryCls = {
+                "BooleanValue": _BooleanEntry
+            }[argInfo.pop("ty")]
+            widget = entryCls(argName, self.builderFrame, **argInfo)
+            item = QListWidgetItem(self.builderFrame.argsListWidget)
+            self.builderFrame.argsListWidget.addItem(item)
+            self.builderFrame.argsListWidget.setItemWidget(item, widget)
 
     @pyqtSlot()
     def submit(self):
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -313,7 +313,6 @@ class BuilderApp(qiwis.BaseApp):
                 "EnumerationValue": _EnumerationEntry,
                 "NumberValue": _NumberEntry
             }[argInfo.pop("ty")]
-            print(argName)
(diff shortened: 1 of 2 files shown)


iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+
+
+class _BooleanEntry(_BaseEntry):
+    """Entry class for a boolean value."""
+
+    def __init__(self, name: str, default: bool = False, parent: Optional[QWidget] = None):
+        """Extended.
+        
+        Args:
+            default: The default value. If it does not exist, it is set to False.
```
kangz12345: The second sentence ("If it ...") is redundant because the code already tells you that.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        return self.checkBox.isChecked()
+
+
+class _EnumerationEntry(_BaseEntry):
+    """Entry class for an enumeration value."""
+
+    def __init__(
+        self,
+        name: str,
+        choices: List[str],
```
kangz12345: Please take abstract type for the parameters, e.g., `Iterable`.
However, of course, in this case, you used `if choices` so `Iterable` is not appplicable.
Nevertheless, what happens if the `choices` is empty? Is it a valid situation?
I think it should be considered as an invalid input.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+    """
+
+    def __init__(
+        self,
+        name: str,
+        unit: str,
+        scale: float,
+        step: float,
+        min: Optional[float],  # pylint: disable=redefined-builtin
+        max: Optional[float],  # pylint: disable=redefined-builtin
```
kangz12345: Let's not redefine the built-in identifiers. You could use `minimum` or `min_`.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        Returns the value of the comboBox.
+        """
+        return self.comboBox.currentText()
+
+
+class _NumberEntry(_BaseEntry):
+    """Entry class for a number value.
+    
+    Attributes:
+        scale: The scale factor that actually applies.
```
kangz12345: Please be more specific, e.g., "... that is multiplied to the number value".

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+    def __init__(
+        self,
+        name: str,
+        unit: str,
+        scale: float,
+        step: float,
+        min: Optional[float],  # pylint: disable=redefined-builtin
+        max: Optional[float],  # pylint: disable=redefined-builtin
+        ndecimals: int,
+        type: str,  # pylint: disable=redefined-builtin
```
kangz12345: If you take the type like this (by `str`), the input domain is too broad, because it only takes `"int"` or `"float"`.
I would rather use a boolean type parameter, e.g., `isInt: bool`.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+
+    def __init__(
+        self,
+        name: str,
+        unit: str,
+        scale: float,
+        step: float,
+        min: Optional[float],  # pylint: disable=redefined-builtin
+        max: Optional[float],  # pylint: disable=redefined-builtin
+        ndecimals: int,
```
kangz12345: Even though ARTIQ uses this name, we are going to use it directly as the `decimals` property of the `QDoubleSpinBox`.
Therefore, I would prefer to use `decimals: int` instead. Then the readers would guess that this parameter works just as the one in a `QDoubleSpinBox`.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        parent: Optional[QWidget] = None
+    ):  # pylint: disable=too-many-arguments
+        """Extended.
+        
+        Args:
+            unit: The unit of the value.
+            step: The step between values changed by the up and down button.
+            min: The minimum value. None for no minimum.
+            max: The maximum value. None for no maximum.
+            ndecimals: The maximum number of decimals.
```
kangz12345: As you see in [`QDoubleSpinBox.decimals`](https://doc.qt.io/qt-6/qdoublespinbox.html#decimals-prop), it is not the _maximum_ number of decimals, but the number of displayed decimals (fixed).

kangz12345: You can just say that it is the decimals of QDoubleSpinBox.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+            default: The default value. If it does not exist, it is set to the min value.
```
kangz12345: "If None, " would be more accurate.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        min: Optional[float],  # pylint: disable=redefined-builtin
+        max: Optional[float],  # pylint: disable=redefined-builtin
+        ndecimals: int,
+        type: str,  # pylint: disable=redefined-builtin
+        default: Optional[float] = None,
+        parent: Optional[QWidget] = None
+    ):  # pylint: disable=too-many-arguments
+        """Extended.
+        
+        Args:
```
kangz12345: I think you should clearly denote the unit or scale for each parameter (or write an explanation in this docstring). Remember that you were confused as well, and explain in more detail to prevent such confusion.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+            unit: The unit of the value.
+            step: The step between values changed by the up and down button.
+            min: The minimum value. None for no minimum.
+            max: The maximum value. None for no maximum.
```
kangz12345: Do you mean "no mini/maximum" by boundless, i.e., no limit?

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        self.scale = scale
+        self.type = type
+        # widgets
+        self.spinBox = QDoubleSpinBox(self)
+        self.spinBox.setSuffix(unit)
+        self.spinBox.setSingleStep(step / scale)
+        if min is not None:
+            self.spinBox.setMinimum(min / scale)
+        if max is not None:
+            self.spinBox.setMaximum(max / scale)
```
kangz12345: If you see [`minimum`](https://doc.qt.io/qt-6/qdoublespinbox.html#minimum-prop) and [`maximum`](https://doc.qt.io/qt-6/qdoublespinbox.html#maximum-prop), they are not boundless by default. You have to set them with `float("inf")`, for example.

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        
+        Args:
+            unit: The unit of the value.
+            step: The step between values changed by the up and down button.
+            min: The minimum value. None for no minimum.
+            max: The maximum value. None for no maximum.
+            ndecimals: The maximum number of decimals.
+            default: The default value. If it does not exist, it is set to the min value.
+            scale, type: See the attributes section in _NumberEntry.
+        """
```
kangz12345: Maybe we should check input validity here:

1. If a predefined unit and a scale are given simultaneously and do not match
2. If `min` > `max`
3. If `default` is not in [`min`, `max`]

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        super().__init__(name, parent=parent)
+        self.scale = scale
+        self.type = type
+        # widgets
+        self.spinBox = QDoubleSpinBox(self)
```
kangz12345, line 140: Are you going to use `QDoubleSpinBox` regardless of the type (int, float)?
Moreover, even if the type is int, is it okay to give non-zero `decimals`?

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+
+    def __init__(self, name: str, default: str = "", parent: Optional[QWidget] = None):
+        """Extended.
+        
+        Args:
+            default: The default value. If it does not exist, it is set to an empty string.
+        """
+        super().__init__(name, parent=parent)
+        # widgets
+        self.lineEdit = QLineEdit(self)
```
kangz12345, line 175: Attribute docstring please!

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+    
+    Attributes:
+        submitButton: The button for submitting the experiment.
+    """
+
+    def __init__(self, parent: Optional[QWidget] = None):
+        """Extended."""
+        super().__init__(parent=parent)
+        # widgets
+        self.argsListWidget = QListWidget(self)
```
kangz12345, line 200: Attribute docstring please!

iquip/apps/builder.py
```diff
@@ -0,0 +1,419 @@
...
+        # connect signals to slots
+        self.builderFrame.submitButton.clicked.connect(self.submit)
+
+    def initArgsEntry(self, experimentInfo: ExperimentInfo):
+        """Initializes the build arguments entry.
+        
+        Args:
+            experimentInfo: The experiment information.
+        """
+        for argName, _argInfo in experimentInfo.arginfo.items():
```
kangz12345: How about:

```suggestion
        for argName, (argInfo, *_) in experimentInfo.arginfo.items():
```

kangz12345 (changes requested):

iquip/apps/builder.py
BECATRUE: The case the `choices` is empty never happens because in that case fetching the build arguments raises an exception.
Nevertheless if you think we should handle it, I will add it.

BECATRUE: Also, the proxy server always offers the `choices` with a list type and should be the type modified to `Iterable`.

iquip/apps/builder.py
BECATRUE: I hope that, too... but I want to use the original parameter name which is returned by the proxy server.

I came up with three solutions.
1. Modify `initArgsEntry()` to replace `min` and `max` to other names of `argInfo`.
For this, we should handle a `NumberValue` separately.
2. Modify the artiq proxy to change arguments name.
Similar to the above solution, we should handle separately.
3. Receive the arguments as a dict not separately.
4. Keep it as it is.

I think solution 3 is best, how about you?

iquip/apps/builder.py
BECATRUE: As I said above, the arguments of the constructor would be set automatically, not explicitly.
Concretly, the response of the proxy server which is the return value of `artiq.master.experiments.ExperimentDB.examine()` is passed to the arguments.
So currently we cannot modify the names..

iquip/apps/builder.py
BECATRUE: It looks necessary! However, how about postponing it in the later issue?
I will implement a tooltip to show the real applied value and handle the pre-defined units.
I think the exact manual will be decided only after this is over.

iquip/apps/builder.py
BECATRUE: Yes! I inteded no limit.

BECATRUE:
I think we first should decide how to handle the build arguments of `argInfo` in `initArgsEntry()`.
Please leave your opinion freely!

iquip/apps/builder.py
kangz12345: > The case the `choices` is empty never happens because in that case fetching the build arguments raises an exception.
> Nevertheless if you think we should handle it, I will add it.

No, my point was: "is `if choices` check necessary?"
You can assume that `choices` are always not empty as you said.
Because of `if choices`, you cannot use `Iterable` type annotation.

However, as you said, I think it is okay to use `List` since this is an internal class.

Conclusion: You can keep or change the current code as you want.

iquip/apps/builder.py
kangz12345: I also think 3 is the best. Then you can keep the docstring like:
```python
"""...

Args:
    argInfo: ... handled arguments are:
        ...
        min: ... (default=0)
        ...
        type: 
"""
```

BECATRUE: Good! Then, is it okay to keep the key names such as `min`, `type`, and `ndecimals`?

iquip/apps/builder.py
BECATRUE: Oh, I see. Then, I will keep it as it is.

BECATRUE: Oh, I'm sorry that I said the wrong thing about `choices`..

First, `choices` is always passed, i.e. it is guaranteed that it isn't `None`.
However, it may be an empty list and the experiment __CANNOT__ be successful since the argument cannot be assigned to any value.
But I think this PR isn't responsible for this problem, so I just ignore the case.

iquip/apps/builder.py
BECATRUE: I planned to handle the first case in #49.
How about handling the left cases in there?

iquip/apps/builder.py
BECATRUE: > Are you going to use QDoubleSpinBox regardless of the type (int, float)?

Yes. I think it isn't a problem to use `QDoubleSpinBox` when the type is int.

BECATRUE: > Moreover, even if the type is int, is it okay to give non-zero decimals?

In that case, the proxy server would return an error because it isn't allowed in ARTIQ.

iquip/apps/builder.py
BECATRUE: Thanks for your suggestion! I applied it.

BECATRUE:
Please check the updated codes! @kangz12345

iquip/apps/builder.py
kangz12345: I don't understand what you said.
1. Why do you care if `choices` is `None`? You didn't set the default value for it.
2. Do you mean that an empty list can be given to `choices`?

Please explain more about this sentence:

> However, it may be an empty list and the experiment CANNOT be successful since the argument cannot be assigned to any value.

iquip/apps/builder.py
kangz12345: Yes, it is okay since they are just strings.

iquip/apps/builder.py
kangz12345: Okay then, please leave a TODO comment.

iquip/apps/builder.py
kangz12345: So if I set a number value with `type="int"` and `ndecimals=1` in an experiment class, what happens when I fetch the arguments of that experiment class?

iquip/apps/builder.py
BECATRUE: Okay! I said it wrong at first, so I corrected it below.

The `choices` is always given so I didn't set its default value.
But the problem is when it is set as an empty list.
In that case, a user cannot set it to any value because there is no allowed candidate.
Once the user submits it, it causes an exception that there is an empty argument.

iquip/apps/builder.py
BECATRUE: In that case, the proxy server causes the exception as below.



On the server side, the log is as follows.

image

iquip/apps/builder.py
kangz12345: Okay, then `choices` here is always not empty, right?

iquip/apps/builder.py
kangz12345: Thank you for the screenshots! That confirms that ARTIQ rejects such ill-formatted experiment codes.

iquip/apps/builder.py
BECATRUE: Nope! It may be an empty list! (Is it right that 'empty' means an emtpy list?)
If a user set `choices` as an empty list in code, the builder is open well but the execution would be failed.

kangz12345: Oh, now I understand. Okay then, please create an issue for handling an empty enumeration value.

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+    This is a wrapper of each entry.
+    In each subclass, value() must be implemented to return the selected value.
+
+    Attributes:
+        name: The argument name.
+        argInfo: The dictionary with the argument options.
+        nameLabel: The label for showing the argument name.
+    """
+
+    def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget] = None):
```
kangz12345, line 29: Why did you write this in a single line here and in multiple lines in `_EnumerationEntry` and `_NumberEntry`?

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+        """
+        raise NotImplementedError
+
+
+class _BooleanEntry(_BaseEntry):
+    """Entry class for a boolean value.
+    
+    Attributes:
+        argInfo: Each key and its value are:
+            default: The default value.
```
kangz12345: Let's write the type of each values.

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+        return self.checkBox.isChecked()
+
+
+class _EnumerationEntry(_BaseEntry):
+    """Entry class for an enumeration value.
+    
+    Attributes:
+        argInfo: Each key and its value are:
+            choices: The pre-defined candidates.
+            default: The default value. If it does not exist, it is set to the first candidate.
```
kangz12345: Based on your code, "If None" seems more accurate.

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+        super().__init__(name, argInfo, parent=parent)
+        # widgets
+        self.comboBox = QComboBox(self)
+        for choice in self.argInfo["choices"]:
+            self.comboBox.addItem(choice)
+        if self.argInfo["choices"]:
+            self.comboBox.setCurrentText(
+                self.argInfo["choices"][0] if self.argInfo["default"] is None
+                else self.argInfo["default"]
+            )
```
kangz12345: I think you use `self.argInfo["choices"]` a lot.
If you want to, you can define a variable for it:
```python
choices = self.argInfo["choices"]
```
Or, moreover
```python
choices, default = map(self.argInfo.get, ("choices", "default"))
```

Of course, this is not necessary and you can keep your current code!

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+    Attributes:
+        argInfo: Each key and its value are:
+            default: The default value. If None, it is set to the min value.
+            unit: The unit of the value.
+            scale: The scale factor that is multiplied to the number value.
+            step: The step between values changed by the up and down button.
+            min: The minimum value. (default=0.0)
+            max: The maximum value. (default=99.99)
+            ndecimals: The number of displayed decimals.
+            type: The type of the value. If "int", value() returns an integer value.
```
kangz12345: I think it should include something like "it is regarded as float otherwise".

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+        self.spinBox = QDoubleSpinBox(self)
+        self.spinBox.setSuffix(self.argInfo["unit"])
+        self.spinBox.setSingleStep(self.argInfo["step"] / self.argInfo["scale"])
+        if self.argInfo["min"] is not None:
+            self.spinBox.setMinimum(self.argInfo["min"] / self.argInfo["scale"])
+        if self.argInfo["max"] is not None:
+            self.spinBox.setMaximum(self.argInfo["max"] / self.argInfo["scale"])
+        self.spinBox.setDecimals(self.argInfo["ndecimals"])
+        self.spinBox.setValue((self.argInfo["min"] if self.argInfo["default"] is None \
+                               else self.argInfo["default"]) / self.argInfo["scale"])
```
kangz12345: 1. Is the backslash necessary?
2. The expression is too convoluted. You can extract the `if-else` expression as a variable.

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+        typeCls = int if self.argInfo["type"] == "int" else float
+        return typeCls(self.spinBox.value() * self.argInfo["scale"])
+
+
+class _StringEntry(_BaseEntry):
+    """Entry class for a string value.
+    
+    Attributes:
+        argInfo: Each key and its value are:
+            default: The default value. If it does not exist, it is set to an empty string.
```
kangz12345: Your code seems it does not care if it exists.

iquip/apps/builder.py
```diff
@@ -1,32 +1,212 @@
...
+        """
+        super().__init__(parent=parent)
+        self.name = name
+        self.argInfo = argInfo
+        # widgets
+        self.nameLabel = QLabel(name, self)
+        # layout
+        self.layout = QHBoxLayout(self)
+        self.layout.addWidget(self.nameLabel)
+        self.setLayout(self.layout)
```
kangz12345: Usually you did not use `setLayout()`, did you? Do you know what the difference is?

iquip/apps/builder.py
```diff
@@ -107,30 +295,47 @@ def __init__(
...
+        Args:
+            experimentInfo: The experiment information.
+        """
+        for argName, (argInfo, *_) in experimentInfo.arginfo.items():
+            entryCls = {
+                "BooleanValue": _BooleanEntry,
+                "StringValue": _StringEntry,
+                "EnumerationValue": _EnumerationEntry,
+                "NumberValue": _NumberEntry
+            }[argInfo.pop("ty")]
```
kangz12345, line 319: Aren't there any other types, e.g., `Scannable`?

kangz12345 (changes requested):

iquip/apps/builder.py
BECATRUE: It was just a mistake😅 I will change all into a single line.

iquip/apps/builder.py
BECATRUE: I realized that if there is no default value, `argInfo` doesn't contain the default element!
So I will change the code using `argInfo.get()` not `argInfo[...]` and modify the docstring.

iquip/apps/builder.py
BECATRUE: Oh, it looks simpler! I will apply your method😃

iquip/apps/builder.py
BECATRUE: It looks more concrete!

iquip/apps/builder.py
BECATRUE: Oh, you are right!
There may be other types, but I planned not to handle them and assumed that handled experiments don't have those types.

Then, how about adding TODO comment?

iquip/apps/builder.py
BECATRUE: Oh, I will handle it.

iquip/apps/builder.py
BECATRUE: I tried to remove it, but I forgot😅
As I understand, `setLayout()` is unnecessary because it is already done while `layout = QHBoxLayout(self)`

iquip/apps/builder.py
```diff
@@ -81,27 +82,22 @@ class _EnumerationEntry(_BaseEntry):
...
         self.comboBox = QComboBox(self)
-        for choice in self.argInfo["choices"]:
-            self.comboBox.addItem(choice)
-        if self.argInfo["choices"]:
-            self.comboBox.setCurrentText(
-                self.argInfo["choices"][0] if self.argInfo["default"] is None
-                else self.argInfo["default"]
-            )
+        self.comboBox.addItems(choices)
+        self.comboBox.setCurrentText(self.argInfo.get("default", choices[0]))
```
kangz12345, line 100: 👍

iquip/apps/builder.py
```diff
@@ -118,38 +114,38 @@ class _NumberEntry(_BaseEntry):
...
             max: The maximum value. (default=99.99)
             ndecimals: The number of displayed decimals.
-            type: The type of the value. If "int", value() returns an integer value.
+            type: The type of the value.
+              If "int", value() returns an integer value.
+              Otherwise, it is regarded as a float value.
         spinBox: The spinbox showing the number value.
     
     TODO(BECATRUE): The operations of unit and scale will be concretized in Basic Runner project.
+    TODO(BECATRUE): Handling the case where the default doesn't exist and the min is None
```
kangz12345, line 131: Then `default` does not exist and `min` is `None` when they are not given?

iquip/apps/builder.py
kangz12345: I think you forgot this again!

iquip/apps/builder.py
```diff
@@ -1,31 +1,207 @@
...
 import qiwis
 from iquip.protocols import ExperimentInfo
 
+class _BaseEntry(QWidget):
+    """Base class for all argument entries.
+
+    This is a wrapper of each entry.
+    In each subclass, value() must be implemented to return the selected value.
+
+    Attributes:
```
kangz12345, line 22: The `self.layout` is missing! I think you should explain that it is a `QHBoxLayout` and the label is already added, so the subclasses can just add the input widget (checkbox, spinbox, etc.).

kangz12345 (changes requested):

iquip/apps/builder.py
BECATRUE: Oh, it looks helpful! I will add it.

iquip/apps/builder.py
BECATRUE: Umm.. It is entirely up to us.
In fact, the artiq-dashboard doesn't handle the default value when it is not given.

However, I want to handle it to prevent the non-default error.
So I planned to set it just to zero.

kangz12345: Do you mean it depends on the implementation of `artiq-proxy`?

kangz12345 (approved):
LGTM

iquip/apps/builder.py
BECATRUE: Nope! In the artiq-dashboard, if the default value is not given, it left the empty entry.
In this case, if the user submits it without setting the value, it would result in the non-default error.

Thus, I want to handle entries of which default value is not given to prevent forgetting to set it from an user.

PR #53, opened 2023-07-05
Initialize the code viewer GUI

### LINKED ISSUE #52: Initialize the code viewer GUI
This is the first step for the code viewer GUI.

I will implment two functions:
1. Make the viewer structure as a `QTreeWidget`.
2. Add the code blocks under the condition there is only non-block statements.

- Each code block is a `QTreeWidgetItem` in the root `QTreeWidget`.
- Assuming there is only non-block statements, there is no item with the child items.
- To make debugging easy, a `visualize` button will be added in the explorer GUI.
It is just temporary, and it will be moved into the builder GUI.

BECATRUE:
This PR will close #52.

As I said in #52, I implemented two things.
1. Initialize the code viewer as a `QTreeWidget`.
2. Show all statements of run() just in series.

_NOT IMPLEMENTED_
- All control and loop statement such as `if` and `for`.
- Remove unnecessary statements.
- For control or loop statement, show only the first line, i.e. except for the body.

#### Code
```python
from artiq.experiment import *

class VisualizeTest(EnvExperiment):
    """Visualize Test"""

    def build(self):
        self.setattr_device("core")
        self.setattr_device("a")
        self.setattr_device("b")

    @kernel
    def run(self):
        self.a.pulse(1*MHz, 1*us)
        with parallel:
            self.b.pulse(1*MHz, 1*us)
            self.a.pulse(1*MHz, 1*us)
```

### DIFF
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -231,7 +231,11 @@ class ExplorerApp(qiwis.BaseApp):
                 module="iquip.apps.visualizer",
                 cls="VisualizerApp",
                 show=True,
-                pos="left"
+                pos="left",
+                args={
+                    "experimentPath": experimentPath,
+                    "experimentClsName": experimentClsName
+                }
             )
         )
 
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -39,6 +39,7 @@ class ExplorerFrame(QWidget):
         layout.addWidget(self.reloadButton)
         layout.addWidget(self.fileTree)
         layout.addWidget(self.openButton)
+        layout.addWidget(self.visualizeButton)
 
 
 class _FileFinderThread(QThread):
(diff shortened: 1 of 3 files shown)


BECATRUE:
To obtain the raw code text from the ast tree (column 2 of the viewer), I used [`ast.get_source_segment()`](https://docs.python.org/3/library/ast.html#ast.get_source_segment).
However, this is only supported after version 3.8, and our supporting lowest version is 3.6 so it makes GitHub action errors.

As @kangz12345 and I said briefly that the version 3.6 is too old, how about raise the lowest version to 3.8 or 3.9?

kangz12345:
Is it 3.6? I thought it is 3.7.
Anyway, I am also considering upgrading the lowest version

BECATRUE:
> Is it 3.6? I thought it is 3.7. Anyway, I am also considering upgrading the lowest version

Oh, you are right! Then, let's discuss it in that issue.

kangz12345:
Let's finish the Builder features first, I will close this PR for a while.

BECATRUE:
For smooth development for v0.3.0-alpha, I think we should reopen this PR now.

BECATRUE:
As @kangz12345 and I discussed offline, we decided to change the visualizer plan and will not make the code viewer the way it is.

PR #65, opened 2023-07-13
Set the scheduler options in the builder

BECATRUE:
I added the scheduler options in the builder app.

### Functions
There are three scheduler options.
- Pipeline: The pipeline to run the experiment in.
- Priority: Higher value means sooner scheduling.
- Timed: The due date for the experiment in ISO format. None for no due date.

I added these options using several `Entry` classes implemented in #47.

### Examples
image
image

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -403,7 +403,7 @@ class BuilderApp(qiwis.BaseApp):
             self.builderFrame.schedOptsListWidget.setItemWidget(item, widget)
 
     def argumentsFromListWidget(self, listWidget: QListWidget) -> Dict[str, Any]:
-        """Gets arguments from the given list widget and return them.
+        """Gets arguments from the given list widget and returns them.
         
         Args:
             listWidget: The QListWidget containing _BaseEntry instances.
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -216,6 +216,7 @@ class ExperimentSubmitThread(QThread):
     Attributes:
         experimentPath: The path of the experiment file.
         experimentArgs: The arguments of the experiment.
+        schedOpts: The scheduler options; pipeline, priority, and timed.
     """
 
     submitted = pyqtSignal(int)
@@ -224,18 +225,21 @@ class ExperimentSubmitThread(QThread):
         self,
         experimentPath: str,
         experimentArgs: Dict[str, Any],
+        schedOpts: Dict[str, Any],
         callback: Callable[[int], None],
         parent: Optional[QObject] = None
-    ):
+    ):  # pylint: disable=too-many-arguments
         """Extended.
         
         Args:
-            experimentPath, experimentArgs: See the attributes section in ExperimentSubmitThread.
+            experimentPath, experimentArgs, schedOpts:
+              See the attributes section in ExperimentSubmitThread.
             callback: The callback method called after this thread is finished.
         """
         super().__init__(parent=parent)
         self.experimentPath = experimentPath
         self.experimentArgs = experimentArgs
+        self.schedOpts = schedOpts
         self.submitted.connect(callback, type=Qt.QueuedConnection)
 
     def run(self):
@@ -255,6 +259,8 @@ class ExperimentSubmitThread(QThread):
         except TypeError:
             print("Failed to convert the build arguments to a JSON string.")
             return
+        for schedOptName, schedOptValue in self.schedOpts.items():
+            params[schedOptName] = schedOptValue
         try:
             response = requests.get("http://127.0.0.1:8000/experiment/submit/",
                                     params=params,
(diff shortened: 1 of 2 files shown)


BECATRUE:
I will implement the unittest for this issue in the next PR, so please ignore the unittest result.

Also, I modify the approving mechanism that at lease one of Jiyong and I should approve and the number of approving reviews is at least 2, but I'm not sure if it works well...

iquip/apps/builder.py
```diff
@@ -185,6 +185,49 @@ def value(self) -> str:
...
+        self.dateTimeEdit.setDisplayFormat("yyyy-MM-dd HH:mm:ss")
+        self.dateTimeEdit.setEnabled(False)
+        self.dateTimeEdit.setMinimumDateTime(currentDateTime)
+        # layout
+        self.layout.addWidget(self.checkBox)
+        self.layout.addWidget(self.dateTimeEdit)
+
+    @pyqtSlot()
+    def updateDateTimeEditState(self):
+        """Enables or disables the dateTimeEdit in according to the state of the checkBox.
```
kangz12345: You can use "according to" or "in accordance with".

iquip/apps/builder.py
```diff
@@ -185,6 +185,49 @@ def value(self) -> str:
...
+        """
+        self.dateTimeEdit.setEnabled(self.checkBox.isChecked())
+
+    def value(self) -> Optional[str]:
+        """Overridden.
+        
+        Returns the value of the dateTimeEdit in ISO format if it is enabled.
+        Otherwise returns None.
+        """
+        return self.dateTimeEdit.dateTime().toString(Qt.ISODate) if self.dateTimeEdit.isEnabled() \
```
kangz12345: Considering the GUI, how about checking `checkBox` instead of `dateTimeEdit`?
Not going to happen but `dateTimeEdit` can be enabled or disabled via codes and the users will see the checkbox to know if the "timed" option is enabled.

kangz12345: Please avoid using the backslash.

iquip/apps/builder.py
```diff
@@ -198,10 +241,12 @@ def __init__(self, parent: Optional[QWidget] = None):
         super().__init__(parent=parent)
         # widgets
         self.argsListWidget = QListWidget(self)
+        self.schedOptsListWidget = QListWidget(self)
```
kangz12345, line 245: Please write the attribute docstring for `schedOptsListWidget`.

iquip/apps/builder.py
```diff
@@ -253,6 +302,8 @@ def run(self):
         except TypeError:
             print("Failed to convert the build arguments to a JSON string.")
             return
+        for schedOptName, schedOptValue in self.schedOpts.items():
+            params[schedOptName] = schedOptValue
```
kangz12345: You can use [`dict.update()`](https://docs.python.org/3/library/stdtypes.html#dict.update): `params.update(self.schedOpts)`

iquip/apps/builder.py
```diff
@@ -323,20 +375,65 @@ def initArgsEntry(self, experimentInfo: ExperimentInfo):
...
+            _NumberEntry("priority", priorityInfo),
+            _DateTimeEntry("timed")
+        ):
+            item = QListWidgetItem(self.builderFrame.schedOptsListWidget)
+            item.setSizeHint(widget.sizeHint())
+            self.builderFrame.schedOptsListWidget.addItem(item)
+            self.builderFrame.schedOptsListWidget.setItemWidget(item, widget)
+
+    def argumentsFromListWidget(self, listWidget: QListWidget) -> Dict[str, Any]:
+        """Get arguments from the given list widget and return them.
```
kangz12345: Gets ... and returns them.

kangz12345 (changes requested):

iquip/apps/builder.py
BECATRUE: Then, do you mean using a parentheses?
I think we should make an explicit line continuation using a backslash or a parentheses.

BECATRUE: Oh, it looks more clear! I will apply it.

BECATRUE:
Umm.. I misused the rebase option, so I think the CODEOWNER file has been added.
But.. it seems not to make a problem, so please ignore it😅

iquip/apps/builder.py
kangz12345: If the line becomes too long, you have to separate the code into multiple lines.
For example, you should not use the if-else _expression_ (or, the ternary operator) in that case. You should rather use the if-else _statement_.

Backslashes are not that readable (although this case was quite simple) in general.
If you have to write a long line, parentheses are preferred as stated in [PEP8](https://peps.python.org/pep-0008/#maximum-line-length).

iquip/apps/builder.py
```diff
@@ -404,7 +403,7 @@ def initSchedOptsEntry(self):
             self.builderFrame.schedOptsListWidget.setItemWidget(item, widget)
 
     def argumentsFromListWidget(self, listWidget: QListWidget) -> Dict[str, Any]:
-        """Get arguments from the given list widget and return them.
+        """Gets arguments from the given list widget and return them.
```
kangz12345: returns!

iquip/apps/builder.py
```diff
@@ -221,18 +221,18 @@ def updateDateTimeEditState(self):
...
         
-        Returns the value of the dateTimeEdit in ISO format if it is enabled.
+        Returns the value of the dateTimeEdit in ISO format if the checkBox is enabled.
         Otherwise returns None.
         """
-        return self.dateTimeEdit.dateTime().toString(Qt.ISODate) if self.dateTimeEdit.isEnabled() \
-               else None
-
+        return (self.dateTimeEdit.dateTime().toString(Qt.ISODate) if self.checkBox.isChecked()
+                else None)
```
kangz12345: I think just using the `if-else` statement would look better!
```python
if self.checkBox.isChecked():
    return self.dateTimeEdit.dateTime().toString(Qt.ISODate)
return None
```
Or is there any reason you want to use the ternary operator?

kangz12345 (changes requested):

kangz12345:
Will #68 resolve the unit test errors?

BECATRUE:
> Will #68 resolve the unit test errors?

Yes! the error will be handled in PR #70 of the issue #68.

iquip/apps/builder.py
BECATRUE: Oh, I couldn't comp up with an idea to use `if-else` statement😅
I will apply it.

kangz12345 (approved):
LGTM

PR #73, opened 2023-07-20
Remove Python 3.7 and add 3.11 to GitHub actions

### LINKED ISSUE #54: Upgrade the Python version
As [Status of Python Versions](https://devguide.python.org/versions/) says, Python 3.7 became _end-of-life_ since 2023-06-27.
We use Python 3.7 in the current experiment PC, so I think we should upgrade it to the latest version Python 3.11.
To do so, we may upgrade it step by step as introduced in [an article](https://pythonspeed.com/articles/stop-using-python-3.7/):

> 1. Upgrade to 3.8.
> 2. Fix any bugs you find.
> 3. Upgrade to 3.9, fix any bugs.
> 4. Repeat until you hit Python Python 3.11.

Moreover, [What's New in Python](https://docs.python.org/3.11/whatsnew/index.html) motivates this issue!

### Original comment
To obtain the raw code text from the ast tree (column 2 of the viewer), I used [`ast.get_source_segment()`](https://docs.python.org/3/library/ast.html#ast.get_source_segment).
However, this is only supported after version 3.8, and our supporting lowest version is 3.6 so it makes GitHub action errors.

As @kangz12345 and I said briefly that the version 3.6 is too old, how about raise the lowest version to 3.8 or 3.9?

_Originally posted by @BECATRUE in https://github.com/snu-quiqcl/iquip/issues/53#issuecomment-1620932898_

kangz12345:
This closes #54.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


BECATRUE (approved):
LGTM

BECATRUE:
Please merge this with the code owner authority.

PR #75, opened 2023-07-20
Aijuh/74/implement-log-viewer

### LINKED ISSUE #74: Implement log viewer
### Feature you to implement
Implement a log viewer for iquip Apps.

1. Make a new app logger.py in iquip/apps.
2. By using root logger as in **snu.quiqcl/qiwis**, make iquip apps' log messages deliver to the logger in logger.py.
3. Make pyQt widget for displaying log message.

### Additional context
Design of log widget and exact log format could be determined later.

Aijuh:
Implemented logger app by using same code as qiwis' logger.py.
An image below is implemented logger app. We can see RID value message from builder.py. Position and log format should be determined.

Unfortunately, other app also using logging so unwanted log message printed out since we are using root logger.
I think we can handle it by making naming convention inside iquip app. If other app running by iquip's app wants to print log message, we may read it  inside that specific iquip's app. 
I didn't think all the details but give option (ex. bool variable) to the Baseapp to read all log message or not. If it is supposed to take all the log message, read it using root logger and send to its own logger.
I may missed some points. I will wait for further comments.
 This closes issue #74

image

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -1,6 +1,7 @@
 """App module for editting the build arguments and submitting the experiment."""
 
 import json
+import logging
 from typing import Any, Callable, Dict, Optional, Tuple, Union
 
 import requests
@@ -12,6 +13,9 @@ from PyQt5.QtWidgets import (
 
 import qiwis
 from iquip.protocols import ExperimentInfo
+
+logger = logging.getLogger(__name__)
+
 
 class _BaseEntry(QWidget):
     """Base class for all argument entries.
@@ -251,7 +255,7 @@ class ExperimentSubmitThread(QThread):
                 "args": json.dumps(self.experimentArgs)
             }
         except TypeError:
-            print("Failed to convert the build arguments to a JSON string.")
+            logger.exception("Failed to convert the build arguments to a JSON string.")
             return
         try:
             response = requests.get("http://127.0.0.1:8000/experiment/submit/",
@@ -260,7 +264,7 @@ class ExperimentSubmitThread(QThread):
             response.raise_for_status()
             rid = response.json()
         except requests.exceptions.RequestException as err:
-            print(err)
+            logger.exception(err)
             return
         self.submitted.emit(rid)
 
@@ -352,7 +356,7 @@ class BuilderApp(qiwis.BaseApp):
         
         TODO(BECATRUE): It will be developed in Log Viewer project.
         """
-        print(f"RID: {rid}")
+        logger.info("RID: %d", rid)
 
     def frames(self) -> Tuple[BuilderFrame]:
         """Overridden."""
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -263,8 +263,8 @@ class ExperimentSubmitThread(QThread):
                                     timeout=10)
             response.raise_for_status()
             rid = response.json()
-        except requests.exceptions.RequestException as err:
-            logger.exception(err)
+        except requests.exceptions.RequestException:
+            logger.exception("Failed to submit the experiment.")
             return
         self.submitted.emit(rid)
 
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -347,14 +347,12 @@ class BuilderApp(qiwis.BaseApp):
         self.thread.start()
 
     def onSubmitted(self, rid: int):
-        """Prints the rid after submitted.
+        """Sends the rid to the logger after submitted.
 
         This is the callback function of ExperimentSubmitThread.
 
         Args:
             rid: The run identifier of the submitted experiment.
-        
-        TODO(BECATRUE): It will be developed in Log Viewer project.
         """
         logger.info("RID: %d", rid)
 
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -1,6 +1,7 @@
 """App module for showing the experiment list and opening an experiment."""
 
 import posixpath
+import logging
 from typing import Callable, List, Optional, Tuple, Union
 
 import requests
@@ -12,6 +13,9 @@ from PyQt5.QtWidgets import (
 import qiwis
 from iquip.protocols import ExperimentInfo
 from iquip.apps.thread import ExperimentInfoThread
+
+logger = logging.getLogger(__name__)
+
 
 class ExplorerFrame(QWidget):
     """Frame for showing the experiment list and opening an experiment.
@@ -84,7 +88,7 @@ class _FileFinderThread(QThread):
             response.raise_for_status()
             experimentList = response.json()
         except requests.exceptions.RequestException as err:
-            print(err)
+            logger.exception(err)
             return
         self.fetched.emit(experimentList, self.widget)
 
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -87,8 +87,8 @@ class _FileFinderThread(QThread):
                                     timeout=10)
             response.raise_for_status()
             experimentList = response.json()
-        except requests.exceptions.RequestException as err:
-            logger.exception(err)
+        except requests.exceptions.RequestException:
+            logger.exception("Failed to fetch the file list.")
             return
         self.fetched.emit(experimentList, self.widget)
 
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -1,0 +1,201 @@
+"""Module for log viewer in apps."""
+
+import time
+import logging
+from typing import Any, Optional, Tuple, Callable
+
+from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
+from PyQt5.QtWidgets import (
+    QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox, QComboBox
+)
+
+import qiwis
+
+class _Signaller(QObject):
+    """Signal only for LoggingHandler.
+
+    Signals:
+        signal(log): A formatted log message is emitted. 
+    """
+
+    signal = pyqtSignal(str)
+
+
+class LoggingHandler(logging.Handler):
+    """Handler for logger.
+
+    Sends a log message to connected function using emit.   
+    """
+
+    def __init__(self, slotfunc: Callable[[str], Any]):
+        """Extended.
+
+        Connects the slotfunc to the signal.
+
+        Args:
+            slotfunc: A slot function which is called when a log record is emitted.
+        """
+        super().__init__()
+        self.signaller = _Signaller()
+        self.signaller.signal.connect(slotfunc)
+
+    def emit(self, record: logging.LogRecord):
+        """Overridden.
+        
+        Emits input signal to the connected function.
+        """
+        s = self.format(record)
+        self.signaller.signal.emit(s)
+
+
+class LoggerFrame(QWidget):
+    """Frame for logging.
+
+    Attributes:
+        logEdit: A textEdit which shows all logs.
+        clearButton: A button for clearing all logs.
+        levelBox: A comboBox for setting the logger's level.
+    """
+
+    def __init__(self, parent: Optional[QObject] = None):
+        """Extended."""
+        super().__init__(parent=parent)
+        # widgets
+        self.logEdit = QTextEdit(self)
+        self.logEdit.setReadOnly(True)
+        self.clearButton = QPushButton("Clear")
+        self.levelBox = QComboBox(self)
+        # layout
+        layout = QVBoxLayout(self)
+        layout.addWidget(self.logEdit)
+        layout.addWidget(self.clearButton)
+        layout.addWidget(self.levelBox)
+
+
+class ConfirmClearingFrame(QWidget):
+    """
+    A confirmation frame for log clearing.
+    """
+    confirmed = pyqtSignal()
+
+    def __init__(self, parent: Optional[QObject] = None):
+        """
+        Extended.
+        """
+        super().__init__(parent=parent)
+        # widgets
+        self.label = QLabel("Are you sure to clear?")
+        self.buttonBox = QDialogButtonBox()
+        self.buttonBox.addButton("OK", QDialogButtonBox.AcceptRole)
+        self.buttonBox.addButton("Cancel", QDialogButtonBox.RejectRole)
+        # connect signals
+        self.buttonBox.accepted.connect(self.buttonOKClicked)
+        self.buttonBox.rejected.connect(self.buttonCancelClicked)
+        # layouts
+        layout = QVBoxLayout()
+        self.setLayout(layout)
+        layout.addWidget(self.label)
+        layout.addWidget(self.buttonBox)
+
+    def buttonOKClicked(self):
+        """Clicks OK to clear log."""
+        self.confirmed.emit()
+        self.close()
+
+    def buttonCancelClicked(self):
+        """Clicks Cancel not to clear log."""
+        self.close()
+
+
+class LoggerApp(qiwis.BaseApp):
+    """App for logging.
+
+    Manages a logger frame.
+
+    Attributes:
+        loggerFrame: A frame that shows the logs.
+    """
+
+    def __init__(self, name: str, parent: Optional[QObject] = None):
+        """Extended.
+
+        Args:
+            name: Name of the App
+        """
+        super().__init__(name, parent=parent)
+        self.loggerFrame = LoggerFrame()
+        # connect signals to slots
+        self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
+        self.confirmFrame = ConfirmClearingFrame()
+        self.confirmFrame.confirmed.connect(self.clearLog)
+        self.handler = LoggingHandler(self.addLog)
+        # TODO(aijuh): Change the log format when it is determined.
+        fs ="%(name)s %(message)s"
+        formatter = logging.Formatter(fs)
+        self.handler.setFormatter(formatter)
+        logger = logging.getLogger()
+        logger.setLevel(logging.DEBUG)
+        logger.addHandler(self.handler)
+        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+
+    @pyqtSlot(str)
+    def setLevel(self, text: str):
+        """Responds to the setLevelBox widget and changes the handler's level.
+
+        Args:
+            text: Selected level in the level select box.
+                  It should be one of "DEBUG", "INFO", "WARNING", "ERROR" and "CRITICAL".
+                  It should be case-sensitive and any other input is ignored.
+        """
+        level = {
+            "DEBUG": logging.DEBUG,
+            "INFO": logging.INFO,
+            "WARNING": logging.WARNING,
+            "ERROR": logging.ERROR,
+            "CRITICAL": logging.CRITICAL
+        }
+        if text in level:
+            self.handler.setLevel(level[text])
+
+    def frames(self) -> Tuple[LoggerFrame]:
+        """Overridden."""
+        return (self.loggerFrame,)
+
+    @pyqtSlot(str)
+    def addLog(self, content: str):
+        """Adds a channel name and log message.
+
+        Args:
+            content: Received log message.
+        """
+        timeString = time.strftime("%Y-%m-%d %H:%M:%S", time.localtime(time.time()))
+        self.loggerFrame.logEdit.insertPlainText(f"{timeString}: {content}\n")
+
+    def receivedSlot(self, channelName: str, content: Any):
+        """Overridden.
+
+        Possible channels are as follows.
+
+        "log": Log channel.
+            See self.addLog().
+        """
+        if channelName == "log":
+            if isinstance(content, str):
+                self.addLog(content)
+            else:
+                (logging.getLogger(__name__)).error("The message for the channel log should be a string.")
+        else:
+            (logging.getLogger(__name__)).error("The message was ignored because "
+                             "the treatment for the channel %s is not implemented.", channelName)
+
+    @pyqtSlot()
+    def checkToClear(self):
+        """Shows a confirmation frame for log clearing."""
+        self.broadcast("log", "Clicked to clear logs")
+        self.confirmFrame.show()
+
+    @pyqtSlot()
+    def clearLog(self):
+        """Clears the log text edit."""
+        self.loggerFrame.logEdit.clear()
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -66,7 +66,7 @@ class LoggerFrame(QWidget):
         # widgets
         self.logEdit = QTextEdit(self)
         self.logEdit.setReadOnly(True)
-        self.clearButton = QPushButton("Clear")
+        self.clearButton = QPushButton("Clear", self)
         self.levelBox = QComboBox(self)
         # layout
         layout = QVBoxLayout(self)
@@ -76,18 +76,15 @@ class LoggerFrame(QWidget):
 
 
 class ConfirmClearingFrame(QWidget):
-    """
-    A confirmation frame for log clearing.
-    """
+    """A confirmation frame for log clearing."""
+
     confirmed = pyqtSignal()
 
     def __init__(self, parent: Optional[QObject] = None):
-        """
-        Extended.
-        """
+        """Extended."""
         super().__init__(parent=parent)
         # widgets
-        self.label = QLabel("Are you sure to clear?")
+        self.label = QLabel("Are you sure to clear?", self)
         self.buttonBox = QDialogButtonBox()
         self.buttonBox.addButton("OK", QDialogButtonBox.AcceptRole)
         self.buttonBox.addButton("Cancel", QDialogButtonBox.RejectRole)
@@ -101,12 +98,12 @@ class ConfirmClearingFrame(QWidget):
         layout.addWidget(self.buttonBox)
 
     def buttonOKClicked(self):
-        """Clicks OK to clear log."""
+        """Called when the OK button is clicked."""
         self.confirmed.emit()
         self.close()
 
     def buttonCancelClicked(self):
-        """Clicks Cancel not to clear log."""
+        """Clicked Cancel not to clear log."""
         self.close()
 
 
@@ -117,6 +114,8 @@ class LoggerApp(qiwis.BaseApp):
 
     Attributes:
         loggerFrame: A frame that shows the logs.
+        confirmFame: A frame that asks to whether clear logs.
+        handler: A handler for adding logs to GUI. 
     """
 
     def __init__(self, name: str, parent: Optional[QObject] = None):
@@ -150,8 +149,8 @@ class LoggerApp(qiwis.BaseApp):
 
         Args:
             text: Selected level in the level select box.
-                  It should be one of "DEBUG", "INFO", "WARNING", "ERROR" and "CRITICAL".
-                  It should be case-sensitive and any other input is ignored.
+              It should be one of "DEBUG", "INFO", "WARNING", "ERROR" and "CRITICAL".
+              It should be case-sensitive and any other input is ignored.
         """
         level = {
             "DEBUG": logging.DEBUG,
@@ -162,6 +161,7 @@ class LoggerApp(qiwis.BaseApp):
         }
         if text in level:
             self.handler.setLevel(level[text])
+            (logging.getLogger()).setLevel(level[text])  
 
     def frames(self) -> Tuple[LoggerFrame]:
         """Overridden."""
@@ -180,7 +180,7 @@ class LoggerApp(qiwis.BaseApp):
     @pyqtSlot()
     def checkToClear(self):
         """Shows a confirmation frame for log clearing."""
-        logger.info("Clicked to clear logs")
+        logger.info("Clear button is clicked to clear logs")
         self.confirmFrame.show()
 
     @pyqtSlot()
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -83,7 +83,9 @@ class ConfirmClearingFrame(QWidget):
     Attributes:
         label: The label for displaying a confirmation message to clear logs in the LoggerFrame.
         buttonBox: The buttonBox with OK and Cancel button to check whether to clear logs.
-        confirmed: A pyqtSignal that emits signal when Ok button clicked. 
+
+    Signals:
+        confirmed: A pyqtSignal that emits signal when Ok button is clicked. 
     """
 
     confirmed = pyqtSignal()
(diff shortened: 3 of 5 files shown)


kangz12345:
> Unfortunately, other app also using logging so unwanted log message printed out since we are using root logger.

Good! I intended that. That's why we used the root logger.
We can use the filter options for the logger of the handler to filtering the logs.
You can implement that feature later as well!

iquip/apps/logger.py
```diff
@@ -0,0 +1,202 @@
...
+        self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
+        self.confirmFrame = ConfirmClearingFrame()
+        self.confirmFrame.confirmed.connect(self.clearLog)
+        self.handler = LoggingHandler(self.addLog)
+        # TODO(aijuh): Change the log format when it is determined.
+        fs ="%(name)s %(message)s"
+        formatter = logging.Formatter(fs)
+        self.handler.setFormatter(formatter)
+        logger = logging.getLogger()
+        logger.setLevel(logging.DEBUG)
```
kangz12345: In my opinion, it is more efficient to change the level when the user select the level via the combobox.
For example, when the selected level is `WARNING`, many useless log records will be passed to the root logger, which is quite inefficient.

Aijuh: I will set root logger's level as `WARNING` and change the code of level selector.

Aijuh:
I finished setting initial log handler's level as `WARNING`. Initial GUI window is shown as below.

image

Aijuh:
Also I add a filter inside log handler to pass only designated name of logger.
In the below image, even if the level is `DEBUG` only log message from iquip app printed.
For now, allowed log channel is : `["iquip.apps.builder", "iquip.apps.explorer","iquip.apps.logger", "iquip.apps.thread"]`
image

Aijuh:
I remember that I didn't get exact feedbacks about log filter.
Is it okay to remove the filter part in this PR and just leave log viewer part? If that is the case, this PR will be only about log viewer.
I think we should think more about filter option. I will add option to write log to the text file in next PR.

kangz12345:
> I remember that I didn't get exact feedbacks about log filter.
> 
> Is it okay to remove the filter part in this PR and just leave log viewer part? If that is the case, this PR will be only about log viewer.
> 
> I think we should think more about filter option. I will add option to write log to the text file in next PR.


I'm sorry for the late reply.
Yes, I also think that the filter should be implemented as another feature.
Please remove the filter in this PR and design and implement the filter feature in another PR.

Aijuh:
Remove the filter part of the code.

iquip/apps/builder.py
```diff
@@ -260,7 +264,7 @@ def run(self):
             response.raise_for_status()
             rid = response.json()
         except requests.exceptions.RequestException as err:
-            print(err)
+            logger.exception(err)
```
kangz12345: `logger.exception("Failed to submit the experiment.")`

iquip/apps/explorer.py
```diff
@@ -84,7 +88,7 @@ def run(self):
             response.raise_for_status()
             experimentList = response.json()
         except requests.exceptions.RequestException as err:
-            print(err)
+            logger.exception(err)
```
kangz12345: `logger.exception("Failed to fetch the file list.")`

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        layout = QVBoxLayout(self)
+        layout.addWidget(self.logEdit)
+        layout.addWidget(self.clearButton)
+        layout.addWidget(self.levelBox)
+
+
+class ConfirmClearingFrame(QWidget):
+    """
+    A confirmation frame for log clearing.
+    """
```
kangz12345: You can write it in a single line.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+    confirmed = pyqtSignal()
```
kangz12345, line 91: 1 blank line before this.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __init__(self, parent: Optional[QObject] = None):
+        """
+        Extended.
+        """
```
kangz12345: Please write it in a single line.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        super().__init__(parent=parent)
+        # widgets
+        self.label = QLabel("Are you sure to clear?")
+        self.buttonBox = QDialogButtonBox()
```
kangz12345: Please set the parents: `QLabel("...", self)`.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        levelBox: A comboBox for setting the logger's level.
+    """
+
+    def __init__(self, parent: Optional[QObject] = None):
+        """Extended."""
+        super().__init__(parent=parent)
+        # widgets
+        self.logEdit = QTextEdit(self)
+        self.logEdit.setReadOnly(True)
+        self.clearButton = QPushButton("Clear")
```
kangz12345: `QPushButton("Clear", self)`

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        self.label = QLabel("Are you sure to clear?")
+        self.buttonBox = QDialogButtonBox()
+        self.buttonBox.addButton("OK", QDialogButtonBox.AcceptRole)
+        self.buttonBox.addButton("Cancel", QDialogButtonBox.RejectRole)
+        # connect signals
+        self.buttonBox.accepted.connect(self.buttonOKClicked)
+        self.buttonBox.rejected.connect(self.buttonCancelClicked)
+        # layouts
+        layout = QVBoxLayout()
+        self.setLayout(layout)
```
kangz12345: You can write `layout = QVBoxLayout(self)`.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        layout.addWidget(self.label)
+        layout.addWidget(self.buttonBox)
+
+    def buttonOKClicked(self):
+        """Clicks OK to clear log."""
```
kangz12345: It seems like this method is clicking the OK button.
How about "Called when the OK button is clicked."?

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        """Clicks Cancel not to clear log."""
+        self.close()
+
+
+class LoggerApp(qiwis.BaseApp):
+    """App for logging.
+
+    Manages a logger frame.
+
+    Attributes:
```
kangz12345, line 125: It has more attributes: `confirmFrame` and `handler`.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        rootLogger = logging.getLogger()
+        rootLogger.setLevel(logging.DEBUG)
+        rootLogger.addHandler(self.handler)
+        self.handler.setLevel(logging.WARNING)
+        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+        self.loggerFrame.levelBox.setCurrentText("WARNING")
+
+    @pyqtSlot(str)
+    def setLevel(self, text: str):
```
kangz12345: Can you change the root logger's level here as well?

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        """Responds to the setLevelBox widget and changes the handler's level.
+
+        Args:
+            text: Selected level in the level select box.
+                  It should be one of "DEBUG", "INFO", "WARNING", "ERROR" and "CRITICAL".
+                  It should be case-sensitive and any other input is ignored.
```
kangz12345: We put only 2 more spaces:
```
Args:
    text: ...
      It should be ...
      It should be ...
```

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        Args:
+            content: Received log message.
+        """
+        timeString = time.strftime("%Y-%m-%d %H:%M:%S", time.localtime(time.time()))
+        self.loggerFrame.logEdit.insertPlainText(f"{timeString}: {content}\n")
+
+    @pyqtSlot()
+    def checkToClear(self):
+        """Shows a confirmation frame for log clearing."""
+        logger.info("Clicked to clear logs")
```
kangz12345: Please be more specific: "Clear button is clicked to clear logs".

iquip/apps/thread.py
```diff
@@ -54,7 +58,7 @@ def run(self):
             response.raise_for_status()
             data = response.json()
         except requests.exceptions.RequestException as err:
-            print(err)
+            logger.exception(err)
```
kangz12345: `logger.exception("Failed to fetch the experiment information.")`

iquip/apps/thread.py
```diff
@@ -65,4 +69,4 @@ def run(self):
                 ExperimentInfo(**experimentInfo)
             )
         else:
-            print("The selected item is a non-experiment file.")
+            logger.error("The selected item is a non-experiment file.")
```
kangz12345: Although it is the original text, but I think "The selected item is not an experiment file." sounds better.

kangz12345 (changes requested):

iquip/apps/logger.py
Aijuh: It seems reasonable. I will also change same part in qiwis.

iquip/apps/builder.py
BECATRUE: You should print more detailed description in terms of what exception occurs. (Like Jiyong's suggestion!)

Previously, I just printed `err` because the logger was not implemented yet😅

iquip/apps/builder.py
```diff
@@ -352,7 +356,7 @@ def onSubmitted(self, rid: int):
         
         TODO(BECATRUE): It will be developed in Log Viewer project.
         """
-        print(f"RID: {rid}")
+        logger.info("RID: %d", rid)
```
BECATRUE, line 357: How about updating the method docstring as follows:
```
Sends the rid to the logger after submitted.
```

This is just an example, and if you think 'prints' is still ok, never mind!

BECATRUE: Also, please remove the above TODO comment because you implemented it instead of me.

iquip/apps/thread.py
BECATRUE: How about modifying the level as `logging.info()`?
I think the case where the selected item is not an experiment file is not a critical problem.

Aijuh:
Updated requested changes.

BECATRUE:
After merging this PR, please implement the unit tests for `logger.py`!

BECATRUE (changes requested):

Aijuh:
I don't know much about Unittest. I will read previous issues and PR and will implement it.

BECATRUE:
> I don't know much about Unittest. I will read previous issues and PR.

Okay! Actually, I hoped you will learn about Python unittest through this.

iquip/apps/logger.py
```diff
@@ -76,18 +76,15 @@ def __init__(self, parent: Optional[QObject] = None):
...
     def __init__(self, parent: Optional[QObject] = None):
-        """
-        Extended.
-        """
+        """Extended."""
         super().__init__(parent=parent)
         # widgets
-        self.label = QLabel("Are you sure to clear?")
+        self.label = QLabel("Are you sure to clear?", self)
         self.buttonBox = QDialogButtonBox()
```
kangz12345: Add a parent to this as well!

iquip/apps/logger.py
```diff
@@ -101,12 +98,12 @@ def __init__(self, parent: Optional[QObject] = None):
...
 
     def buttonOKClicked(self):
-        """Clicks OK to clear log."""
+        """Called when the OK button is clicked."""
         self.confirmed.emit()
         self.close()
 
     def buttonCancelClicked(self):
-        """Clicks Cancel not to clear log."""
+        """Clicked Cancel not to clear log."""
```
kangz12345: For consistency, please follow the form of `buttonOKClicked()` docstring.
"Called when the Cancel button is clicked."

iquip/apps/logger.py
```diff
@@ -117,6 +114,8 @@ class LoggerApp(qiwis.BaseApp):
 
     Attributes:
         loggerFrame: A frame that shows the logs.
+        confirmFame: A frame that asks to whether clear logs.
```
kangz12345: "... whether to clear logs" seems better.

kangz12345 (changes requested):

iquip/apps/logger.py
Aijuh: I had fixed it in this way, but since it is not requested part, I just moved to the original one.
I will change it.

iquip/apps/logger.py
```diff
@@ -85,14 +85,14 @@ def __init__(self, parent: Optional[QObject] = None):
...
+        self.buttonBox = QDialogButtonBox(self)
         self.buttonBox.addButton("OK", QDialogButtonBox.AcceptRole)
         self.buttonBox.addButton("Cancel", QDialogButtonBox.RejectRole)
         # connect signals
         self.buttonBox.accepted.connect(self.buttonOKClicked)
         self.buttonBox.rejected.connect(self.buttonCancelClicked)
         # layouts
-        layout = QVBoxLayout()
+        layout = QVBoxLayout(self)
         self.setLayout(layout)
```
kangz12345: You can remove this line since you set the parent of `layout`.

iquip/apps/logger.py
```diff
@@ -114,7 +114,7 @@ class LoggerApp(qiwis.BaseApp):
 
     Attributes:
         loggerFrame: A frame that shows the logs.
-        confirmFame: A frame that asks to whether clear logs.
+        confirmFame: A frame that asks to whether to clear logs.
```
kangz12345: I'm sorry for the confusion, "... asks whether to clear logs."

kangz12345 (changes requested):

iquip/apps/logger.py
BECATRUE: The `confirmFame` looks a typo!

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        level = {
+            "DEBUG": logging.DEBUG,
+            "INFO": logging.INFO,
+            "WARNING": logging.WARNING,
+            "ERROR": logging.ERROR,
+            "CRITICAL": logging.CRITICAL
+        }
+        if text in level:
+            self.handler.setLevel(level[text])
+            (logging.getLogger()).setLevel(level[text])
```
kangz12345: You don't need the parenthesis: `logging.getLogger().setLevel(...)`

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def buttonCancelClicked(self):
+        """Called when the Cancel button is clicked."""
+        self.close()
+
+
+class LoggerApp(qiwis.BaseApp):
+    """App for logging.
+
+    Manages a logger frame.
```
BECATRUE: Please write more explanation. It looks too vague.

BECATRUE: For example, there is another frame, confirmFrame!

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Attributes:
+        loggerFrame: A frame that shows the logs.
+        confirmFame: A frame that asks to whether to clear logs.
+        handler: A handler for adding logs to GUI.
```
BECATRUE: How about "A handler to add logs to the loggerFrame`?

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        loggerFrame: A frame that shows the logs.
+        confirmFame: A frame that asks to whether to clear logs.
+        handler: A handler for adding logs to GUI. 
+    """
+
+    def __init__(self, name: str, parent: Optional[QObject] = None):
+        """Extended.
+
+        Args:
+            name: Name of the App
```
BECATRUE: Please remove this explanation because it is already written in `BaseApp` docstring!
"Extended" expression implies that arguments not shown here are already described in the super class.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        super().__init__(name, parent=parent)
+        self.loggerFrame = LoggerFrame()
+        # connect signals to slots
+        self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
+        self.confirmFrame = ConfirmClearingFrame()
```
BECATRUE: Please move this up to the above comment.
Several comments look necessary to show legibly.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        self.confirmFrame.confirmed.connect(self.clearLog)
+        self.handler = LoggingHandler(self.addLog)
+        # TODO(aijuh): Change the log format when it is determined.
+        fs ="%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s "
```
BECATRUE: The variable name looks vague, but it will be updated later, so you can ignore it now.

iquip/apps/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        super().__init__()
+        self.signaller = _Signaller()
+        self.signaller.signal.connect(slotfunc)
+
+    def emit(self, record: logging.LogRecord):
+        """Overridden.
+        
+        Emits input signal to the connected function.
+        """
+        s = self.format(record)
```
BECATRUE: I think we try not to use a single character name.
According to this [styleguide](https://google.github.io/styleguide/pyguide.html#3161-names-to-avoid), it is not recommended.

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+        rootLogger = logging.getLogger()
+        rootLogger.setLevel(logging.DEBUG)
+        rootLogger.addHandler(self.handler)
+        self.handler.setLevel(logging.WARNING)
+        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+        self.loggerFrame.levelBox.setCurrentText("WARNING")
+
+    @pyqtSlot(str)
+    def setLevel(self, text: str):
```
BECATRUE: How about `levelText`? `text` looks ambiguous.

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+        level = {
+            "DEBUG": logging.DEBUG,
+            "INFO": logging.INFO,
+            "WARNING": logging.WARNING,
+            "ERROR": logging.ERROR,
+            "CRITICAL": logging.CRITICAL
+        }
+        if text in level:
+            self.handler.setLevel(level[text])
+            logging.getLogger().setLevel(level[text])
```
BECATRUE: @kangz12345 Did you intend this?

If so, in `__init__()`, the root logger level and the handler level are set differently, and it looks too awkward.

BECATRUE: Also, in `__init__()`, you can just call `self.setLevel("WARNING")`.

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+        return (self.loggerFrame,)
+
+    @pyqtSlot(str)
+    def addLog(self, content: str):
+        """Adds a channel name and log message.
+
+        Args:
+            content: Received log message.
+        """
+        timeString = time.strftime("%Y-%m-%d %H:%M:%S", time.localtime(time.time()))
```
BECATRUE: It is just an option, and you can use `QDateTime.currentDateTime()` to obtain the current time.
I used this in `builder.py` and it may look more unified.

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+        self.loggerFrame.logEdit.insertPlainText(f"{timeString}: {content}\n")
+
+    @pyqtSlot()
+    def checkToClear(self):
+        """Shows a confirmation frame for log clearing."""
```
BECATRUE: Isn't "... for clearing logs" more right?

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+        logger.info("Clear button is clicked to clear logs")
```
BECATRUE: The statement is exactly correct, but it looks a little confused because "clear" is repeated.

How about "Tried to clear logs"?

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
+"""Module for log viewer in apps."""
```
BECATRUE: How about "App module for showing logs."?

BECATRUE (changes requested):

iquip/apps/logger.py
Aijuh: I will make two handlers and two formats when implement option to add log to file. I will fix name at that PR.

iquip/apps/logger.py
kangz12345: 1. The root logger level
2. The handler level
3. The combobox level text

should always coincide.

BECATRUE: Okay! I think it is okay to remain the root looger level as the lowest level, i.e. DEBUG, but keeping them all the same looks good, too.

kangz12345: I think they should _always_ be the same, so please unify those levels in `__init__()` as well.

Aijuh: I intended to prevent possible problems by changing root logger's level.
 But as I know there is no problem with handling root logger's level.

 I will change this part.

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+        signal(log): A formatted log message is emitted. 
+    """
+
+    signal = pyqtSignal(str)
+
+
+class LoggingHandler(logging.Handler):
+    """Handler for logger.
+
+    Sends a log message to connected function using emit.
```
BECATRUE: '... to the connected'

BECATRUE: Also, 'through a signal' looks better that 'using emit`.

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+        self.levelBox = QComboBox(self)
+        # layout
+        layout = QVBoxLayout(self)
+        layout.addWidget(self.logEdit)
+        layout.addWidget(self.clearButton)
+        layout.addWidget(self.levelBox)
+
+
+class ConfirmClearingFrame(QWidget):
+    """A confirmation frame for log clearing."""
```
BECATRUE: I realized just now that there is no any docstring about its attributes!

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+    def buttonCancelClicked(self):
+        """Called when the Cancel button is clicked."""
+        self.close()
+
+
+class LoggerApp(qiwis.BaseApp):
+    """App for logging.
+
+    Set a handler of root logger and manages loggerFrame to show log messages.
+    Give options to clear logs and select log level in the loggerFrame.
```
BECATRUE: 'Sets' and 'Gives'.

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+        self.confirmFrame.confirmed.connect(self.clearLog)
+        self.handler = LoggingHandler(self.addLog)
+        # TODO(aijuh): Change the log format when it is determined.
+        fs ="%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s "
+        formatter = logging.Formatter(fs)
+        self.handler.setFormatter(formatter)
+        rootLogger = logging.getLogger()
+        rootLogger.setLevel("WARNING")
+        rootLogger.addHandler(self.handler)
+        self.handler.setLevel("WARNING")
```
BECATRUE: Please use `self.setLevel("WARNING")`.

BECATRUE: And does `self.loggerFrame.levelBox.setCurrentText("WARNING")` call `self.setLevel("WARNING")`?
If so, I think you can just remove these two lines.

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+        self.loggerFrame.levelBox.setCurrentText("WARNING")
+
+    @pyqtSlot(str)
+    def setLevel(self, levelText: str):
+        """Responds to the setLevelBox widget and changes the handler's level.
+
+        Args:
+            text: Selected level in the level select box.
```
BECATRUE: Please update the argument name.

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+              It should be one of "DEBUG", "INFO", "WARNING", "ERROR" and "CRITICAL".
+              It should be case-sensitive and any other input is ignored.
+        """
+        level = {
```
BECATRUE: `levels` looks more appropriate because it is not only one level.

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+        return (self.loggerFrame,)
+
+    @pyqtSlot(str)
+    def addLog(self, content: str):
+        """Adds a channel name and log message.
+
+        Args:
+            content: Received log message.
+        """
+        timeString = QDateTime.currentDateTime().toString("yyyy-MM-dd HH:mm:ss")
```
BECATRUE, line 182: 👍

iquip/apps/logger.py
```diff
@@ -0,0 +1,184 @@
...
+        rootLogger.setLevel("WARNING")
+        rootLogger.addHandler(self.handler)
+        self.handler.setLevel("WARNING")
+        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+        self.loggerFrame.levelBox.setCurrentText("WARNING")
+
+    @pyqtSlot(str)
+    def setLevel(self, levelText: str):
+        """Responds to the setLevelBox widget and changes the handler's level.
```
BECATRUE: Isn't the levelBox correct?

BECATRUE (changes requested):
Maybe almost done!

iquip/apps/logger.py
Aijuh: My bad. Next time I change variable name, I will change all the comments using that name.

Aijuh:
I updated codes and got a question. 
I will implement feature for saving logs to file for next PR.
Before, I set root logger as "DEBUG" and change two different handlers' level. (ex. set handler for the loggerFrame as "WARNING", handler for the filesaving as "INFO").
But since level of root logger and LoggerFrame handler change at same time,  file saving handler's level may be ignored.
So which way is better?

1. Two handlers and root logger has all same level.
2.  Set root logger as DEBUG and control logs by handler's level.

kangz12345:
Good question.
In that case:
1. GUI handler level = GUI handler level combobox
2. File handler level = File handler level combobox
3. Root logger level = minimum of 1 and 2 (note that [logger levels](https://docs.python.org/3/library/logging.html#logging-levels) have numeric values so you can use `min()` to get the lower level)

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(self, slotfunc: Callable[[str], Any]):
+        """Extended.
+
+        Connects the slotfunc to the signal.
+
+        Args:
+            slotfunc: A slot function which is called when a log record is emitted.
+        """
+        super().__init__()
+        self.signaller = _Signaller()
```
BECATRUE, line 44: Please write down the docstring about `signaller` attribute.

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+        layout.addWidget(self.clearButton)
+        layout.addWidget(self.levelBox)
+
+
+class ConfirmClearingFrame(QWidget):
+    """A confirmation frame for log clearing in the LoggerFrame.
+    
+    Attributes:
+        label: Displays a confirmation message to clear logs in the LoggerFrame.
+        buttonBox: Contains OK and Cancel button to check whether to clear logs.
```
BECATRUE: Please write in the format:
```
The label for displaying ...
The button box with OK and ...
```

iquip/apps/logger.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+
+    confirmed = pyqtSignal()
```
BECATRUE, line 91: Please add the docstring about this signal.

BECATRUE (changes requested):
This may be the final review!

After done, please request Jiyong for a review.

iquip/apps/logger.py
kangz12345: You should add "Signals:" section in the docstring.

iquip/apps/logger.py
```diff
@@ -0,0 +1,192 @@
...
+        """Extended."""
+        super().__init__(name, parent=parent)
+        self.loggerFrame = LoggerFrame()
+        self.confirmFrame = ConfirmClearingFrame()
+        # connect signals to slots
+        self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
+        self.confirmFrame.confirmed.connect(self.clearLog)
+        self.handler = LoggingHandler(self.addLog)
+        # TODO(aijuh): Change the log format when it is determined.
+        fs ="%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s "
```
kangz12345: Oh, why isn't there a space between `=` and `"`, and why is there a space at the end?
In other words, shouldn't it be `fs = "%(levelname)s ... %(message)s"`?

kangz12345 (changes requested):

iquip/apps/logger.py
Aijuh: My mistake. I will check details.

Aijuh:
Fix comments.

iquip/apps/logger.py
```diff
@@ -172,21 +174,21 @@ def frames(self) -> Tuple[LoggerFrame]:
 
     @pyqtSlot(str)
     def addLog(self, content: str):
-        """Adds a channel name and log message.
+        """Adds a received log message to the LoggerFrame.
 
         Args:
-            content: Received log message.
+            content: A received log message.
```
kangz12345: You don't need "A" here since it is just a noun and describes the meaning of `content`.
If you want to use an article, I think "the" is more appropriate because it refers to the specific log message that is passed as the `content` parameter.

kangz12345: I think these are quite subtle and you only have to be careful when you write sentences.

Aijuh: I will go back to previous one.

kangz12345 (approved):
LGTM

BECATRUE (approved):
LGTM!

PR #69, opened 2023-07-17
Giwon2004/63/scheduler

giwon2004:
The `scheduler.py` is implemented in a _local_ sense, which does not interact with `artiq-proxy` in a moment, and will be implemented afterward. 

The list `expList` is implemented through `QListView`, which holds `ExperimentView` instances.
The model `ExperimentView` inside `expList` also provides a drag & drop feature.

### Examples
+ Here, I had no connection through `artiq-proxy` for now, so I simply added some dummy experiments before starting.

image
+ Current experiment is at the top, while others are on the list.
   - Other information besides their names is displayed under/beside the names.

### Problems
+ EDIT button above is not properly working, and I expect this as a problem with `QAbstractItemDelegate`.
+ Drag & drop feature may not be compatible with the practical use of `artiq-proxy`.

This closes #63.

PR #82, opened 2023-07-27
Implement additional functions of the builder

### LINKED ISSUE #49: Implement additional functions of the builder
Several additional functions of the builder will be implemented.
1. Show the experiment info such as the name and class name.
2. Enable to reload the build arguments from the experiment file.
3. Alert if a `_NumberEntry` uses an atypical scale factor.
4. Show the actual value of a `_NumberEntry` through `QToolTip`.
5. Handle when `min` is greater than `max` in a `_NumberEntry`.
In that case, just swap `min` and `max` each other.
6. Handle when both `min` and `default` are not given.

BECATRUE:
This PR will close #49.

As I wrote in #49, I implemented several additional functions in the builder.
- Show the experiment info such as the name and class name.
- Enable to reload the build arguments from the experiment file.
- Alert if a _NumberEntry uses an atypical scale factor.
- Show the actual value of a _NumberEntry through QToolTip.
- Handle when min is greater than max in a _NumberEntry.
In that case, just swap min and max each other.
- Handle when both min and default are not given.

#### Results
image
In the `time` argument, the scale is set to 0.1 explicitly, so it causes an warning.

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -14,6 +14,18 @@ import qiwis
 from iquip.protocols import ExperimentInfo
 from iquip.apps.thread import ExperimentInfoThread
 
+def compute_unit(unit: str) -> Optional[float]:
+    """Computes the unit scale of the given unit string based on ARTIQ units.
+    
+    Args:
+        unit: The unit string e.g., "ns", "kHz".
+
+    Returns:
+        The scale of the given unit. For example, the scale of "ms" is 0.001.
+        If the unit is not defined in ARTIQ, it returns None.
+    """
+
+
 class _BaseEntry(QWidget):
     """Base class for all argument entries.
 
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -137,6 +137,7 @@ class _NumberEntry(_BaseEntry):
         scale, minValue, maxValue = map(self.argInfo.get, ("scale", "min", "max"))
         # widgets
         self.spinBox = QDoubleSpinBox(self)
+        self.spinBox.valueChanged.connect(self.updateToolTip)
         self.spinBox.setSuffix(self.argInfo["unit"])
         self.spinBox.setSingleStep(self.argInfo["step"] / scale)
         # TODO(BECATRUE): A WARNING log will be added after implementing the logger app.
@@ -166,6 +167,14 @@ class _NumberEntry(_BaseEntry):
         """
         typeCls = int if self.argInfo["type"] == "int" else float
         return typeCls(self.spinBox.value() * self.argInfo["scale"])
+
+    @pyqtSlot()
+    def updateToolTip(self):
+        """Updates the text in the tooltip.
+        
+        Once the value of the spinBox is changed, this is called.
+        """
+        self.setToolTip(str(self.value()))
 
 
 class _StringEntry(_BaseEntry):
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -403,6 +403,8 @@ class BuilderApp(qiwis.BaseApp):
     Attributes:
         builderFrame: The frame that shows the build arguments and requests to submit it.
         experimentPath: The path of the experiment file.
+        experimentSubmitThread: The most recently executed ExperimentSubmitThread instance.
+        experimentInfoThread: The most recently executed ExperimentInfoThread instance.
     """
 
     def __init__(
@@ -422,6 +424,8 @@ class BuilderApp(qiwis.BaseApp):
         """
         super().__init__(name, parent=parent)
         self.experimentPath = experimentPath
+        self.experimentSubmitThread = None
+        self.experimentInfoThread = None
         self.builderFrame = BuilderFrame(experimentInfo["name"], experimentClsName)
         self.initArgsEntry(ExperimentInfo(**experimentInfo))
         self.initSchedOptsEntry()
@@ -484,8 +488,8 @@ class BuilderApp(qiwis.BaseApp):
         
         Once the reloadArgsButton is clicked, this is called.
         """
-        self.thread = ExperimentInfoThread(self.experimentPath, self.onReloaded, self)
-        self.thread.start()
+        self.experimentInfoThread = ExperimentInfoThread(self.experimentPath, self.onReloaded, self)
+        self.experimentInfoThread.start()
 
     def onReloaded(
         self,
@@ -532,14 +536,14 @@ class BuilderApp(qiwis.BaseApp):
         """
         experimentArgs = self.argumentsFromListWidget(self.builderFrame.argsListWidget)
         schedOpts = self.argumentsFromListWidget(self.builderFrame.schedOptsListWidget)
-        self.thread = ExperimentSubmitThread(
+        self.experimentSubmitThread = ExperimentSubmitThread(
             self.experimentPath,
             experimentArgs,
             schedOpts,
             self.onSubmitted,
             self
         )
-        self.thread.start()
+        self.experimentSubmitThread.start()
 
     def onSubmitted(self, rid: int):
         """Prints the rid after submitted.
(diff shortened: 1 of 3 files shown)


Aijuh (approved):
As far as I can check, there is no problem.
Learned how to compute unit from your code. 👍

iquip/apps/builder.py
```diff
@@ -12,6 +12,41 @@
 
 import qiwis
 from iquip.protocols import ExperimentInfo
+from iquip.apps.thread import ExperimentInfoThread
+
+def compute_unit(unit: str) -> Optional[float]:
```
kangz12345: I think `compute_scale()` would be better since it returns the scale of the given `unit`.

iquip/apps/builder.py
```diff
@@ -12,6 +12,41 @@
...
+    
+    Args:
+        unit: The unit string e.g., "ns", "kHz".
+
+    Returns:
+        The scale of the given unit. For example, the scale of "ms" is 0.001.
+        If the unit is not defined in ARTIQ, it returns None.
+    """
+    base_prefixes = "pnum_kMG"
+    unit_specs = {  # Units in ARTIQ
```
kangz12345: You can leave a link of the reference: https://github.com/m-labs/artiq/blob/master/artiq/language/units.py

iquip/apps/builder.py
```diff
@@ -126,28 +161,43 @@ class _NumberEntry(_BaseEntry):
               If "int", value() returns an integer value.
               Otherwise, it is regarded as a float value.
         spinBox: The spinbox showing the number value.
-    
-    TODO(BECATRUE): The operations of unit and scale will be concretized in Basic Runner project.
-    TODO(BECATRUE): Handling the case where the default doesn't exist and the min is None
-      will be implemented in Basic Runner project.
+        warningLabel: The label showing a warning.
+            If the given scale is not typical for the unit, it shows a warning.
```
kangz12345: Indentation!
Indent 4 spaces only for a block after a colon `:`.

iquip/apps/builder.py
```diff
@@ -126,28 +161,43 @@ class _NumberEntry(_BaseEntry):
...
+        unit, scale, minValue, maxValue = map(self.argInfo.get, ("unit", "scale", "min", "max"))
         # widgets
         self.spinBox = QDoubleSpinBox(self)
-        self.spinBox.setSuffix(self.argInfo["unit"])
+        self.spinBox.valueChanged.connect(self.updateToolTip)
+        self.spinBox.setSuffix(unit)
         self.spinBox.setSingleStep(self.argInfo["step"] / scale)
+        # TODO(BECATRUE): A WARNING log will be added after implementing the logger app.
+        if minValue is not None and maxValue is not None and minValue > maxValue:
+            minValue, maxValue = maxValue, minValue
```
kangz12345, line 186: Please document such policy for the inputs, e.g., "if min > max, then they are swapped."

iquip/apps/builder.py
```diff
@@ -126,28 +161,43 @@ class _NumberEntry(_BaseEntry):
...
         if minValue is not None:
             self.spinBox.setMinimum(minValue / scale)
         if maxValue is not None:
             self.spinBox.setMaximum(maxValue / scale)
```
kangz12345: How about setting the `minValue` and `maxValue` to 0 and 99.99, respectively, when they are `None`?
Then you don't have to check if it is `None` afterwards.

iquip/apps/builder.py
```diff
@@ -157,6 +207,14 @@ def value(self) -> Union[int, float]:
         typeCls = int if self.argInfo["type"] == "int" else float
         return typeCls(self.spinBox.value() * self.argInfo["scale"])
 
+    @pyqtSlot()
+    def updateToolTip(self):
+        """Updates the text in the tooltip.
```
kangz12345: Please be more specific: Updates the text to what?

iquip/apps/builder.py
```diff
@@ -238,16 +296,32 @@ class BuilderFrame(QWidget):
...
+        Args:
+            experimentName: The experiment name, the name field of protocols.ExperimentInfo.
+            experimentClsName: The class name of the experiment.
+        """
         super().__init__(parent=parent)
         # widgets
+        self.experimentNameLabel = QLabel(f"Name: {experimentName}", self)
+        self.experimentClsNameLabel = QLabel(f"Class: {experimentClsName}", self)
         self.argsListWidget = QListWidget(self)
+        self.reloadArgsButton = QPushButton("Reload", self)
```
kangz12345, line 316: Attribute docstring please!

iquip/apps/builder.py
```diff
@@ -404,6 +478,33 @@ def initSchedOptsEntry(self):
             self.builderFrame.schedOptsListWidget.addItem(item)
             self.builderFrame.schedOptsListWidget.setItemWidget(item, widget)
 
+    @pyqtSlot()
+    def reloadArgs(self):
+        """Reloads the build arguments.
+        
+        Once the reloadArgsButton is clicked, this is called.
+        """
+        self.thread = ExperimentInfoThread(self.experimentPath, self.onReloaded, self)
```
kangz12345: All the public attributes should be defined in `__init__()` and documented.

iquip/apps/builder.py
```diff
@@ -404,6 +478,33 @@ def initSchedOptsEntry(self):
...
+        self.thread.start()
+
+    def onReloaded(
+        self,
+        _experimentPath: str,
```
kangz12345, line 496: Oh did we use an underscore to avoid the "unused argument" warning?
I think we should agree about the convention, because we used the pylint disable comment in #69.

kangz12345 (changes requested):

iquip/apps/builder.py
BECATRUE: Oh, I agree with you! I will update it.

iquip/apps/builder.py
BECATRUE: Wow, it makes much simpler! I will apply it.

iquip/apps/builder.py
BECATRUE: I added the url according to the [styleguide](https://google.github.io/styleguide/pyguide.html#32-line-length).

iquip/apps/builder.py
BECATRUE: Oh, we didn't document `self.thread` in `explorer.py`, too.

Can I do this here, or is it better to make it as another issue?

iquip/apps/builder.py
BECATRUE: Umm.. I think if the method is overridden, it looks better to use the original variable name like #49.
But, in this case where the method isn't overridden but we can't remove the unused variable, in according to [styleguide](https://google.github.io/styleguide/pyguide.html#214-decision), I want to use an underline.

Or, how about defining only `_`, not `_experimentPath`?

kangz12345: What do you mean by "like #49"?

As you said, since it is not overridden, we don't have to use the original name.
However, to show the "place-holding role" of the unused variable, I think your current naming is good enough.

iquip/apps/builder.py
kangz12345: I think you can do it here!

iquip/apps/builder.py
BECATRUE: > What do you mean by "like https://github.com/snu-quiqcl/iquip/issues/49"?

Oh, sorry. I intended #69 you metioned.

kangz12345: Oh I see.

iquip/apps/builder.py
```diff
@@ -422,6 +424,8 @@ def __init__(
         """
         super().__init__(name, parent=parent)
         self.experimentPath = experimentPath
+        self.experimentSubmitThread = None
+        self.experimentInfoThread = None
```
kangz12345: Let's annotate this like:
```python
self.AThread: Optional[A] = None
```

kangz12345 (changes requested):

kangz12345 (approved):
LGTM

PR #84, opened 2023-07-28
Aijuh/76/save logs to file

### LINKED ISSUE #76: Add an option to save logs to file
### Feature to implement
Save generated logs to file.

1. Add a new text file in iquip directory. (I will set it "log_record" for now.)
2. Make another handler for root logger in logger.py. (log format, log level can be different with the handler for screen.)

### Additional context
We can add new widget for saving logger file. But it would be to complicated, so I am planning to record all the log with `INFO` level.
Details can be fixed later.

Aijuh:
### Added feature for saving logs to file. 
This PR closes issue #76 
Below are images of current loggerFrame and saved log format.

All logs are first saved at temp file by using file handler. I first think that level for fileHandler should be different for frameHandler. But since user chooses desired level, we can just save that level of logs to file. 

 When user presses `Save log to file` button, contents at temp file saved to log record file and temp file is cleared.
 When user presses `Clear log file` button, only log record file cleared.

The time format of logs in loggerFrame and file is different because we specify format for loggerFrame but use `%asctime` of format in fileHandler. 

All the details can be changed.  

log_frame
file_log

### DIFF
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -196,7 +196,7 @@ class LoggerApp(qiwis.BaseApp):
     Attributes:
         loggerFrame: A frame that shows the logs.
         confirmFrame: A frame that asks whether to clear logs.
-        handler: A handler for adding logs to the loggerFrame.
+        frameHandler: A handler for adding logs to the loggerFrame.
         fileHandler: A handler for saving logs to file.
     """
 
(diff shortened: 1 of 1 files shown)


Aijuh:
There is pylint error.
 I also tested pylint in my local, but it says `Too many instance attributes (8/7)` at `class LoggerApp(qiwis.BaseApp):`.
It seems there are 8 attributes in this file (7 classes and 1 logger), Should we delete one class or ignore this error message?

iquip/apps/logger.py
```diff
@@ -135,28 +196,47 @@ class LoggerApp(qiwis.BaseApp):
     Attributes:
         loggerFrame: A frame that shows the logs.
         confirmFrame: A frame that asks whether to clear logs.
-        handler: A handler for adding logs to the loggerFrame. 
+        frameHandler: A handler for adding logs to the loggerFrame.
+        fileHandler: A handler for saving logs to file.
     """
 
     def __init__(self, name: str, parent: Optional[QObject] = None):
```
BECATRUE, line 203: The constructor looks too long!
Please separate this into several methods.

BECATRUE (changes requested):

BECATRUE:
I have several questions.

1. Handling the log file
I think it is natural to save logs to a file consistently in the background without clicking a button.
In addition, I think there are too many button in the `loggerFrame` so that it looks convoluted.
But, I wonder How @kangz12345 think of it.

2. Setting the handler level
Due to the same reason above, I think we should set the file logger level to DEBUG or INFO constantly.

Aijuh:
I am sorry that I didn't update quickly, but I should delete this PR and create new branch.
This PR's branch didn't stem from current branch. It stems from previous error-branch.
I will create new branch and new PR.

PR #86, opened 2023-07-29
Implement the left unit tests for builder app

### LINKED ISSUE #68: Implement the left unit tests for builder app
I will implement the left unit tests for builder app.

BECATRUE:
This PR will close #68.

I implemented the left __unit__ tests for builder app, not functional tests.

So I __DID NOT__ implement tests about the followings.
- All entry classes except for `BaseEntry`
- `BuilderApp.onSubmitted()`

Also, all `print()` statements will be converted to logging, so we can test it using [`assertLogs()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs) later.

### Results
image

### DIFF
--- a/tests/.pylintrc
+++ b/tests/.pylintrc
@@ -9,6 +9,5 @@ variable-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9
 
 [MESSAGES CONTROL]
 disable=
-    duplicate-code,
     missing-docstring,
     protected-access,
--- a/tests/.pylintrc
+++ b/tests/.pylintrc
@@ -9,5 +9,6 @@ variable-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9
 
 [MESSAGES CONTROL]
 disable=
+    duplicate-code,
     missing-docstring,
     protected-access,
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -5,6 +5,7 @@ import json
 import unittest
 from unittest import mock
 
+import requests
 from PyQt5.QtCore import QObject, Qt
 from PyQt5.QtWidgets import QApplication, QListWidget, QListWidgetItem, QWidget
 
@@ -108,6 +109,34 @@ class _ExperimentSubmitThreadTest(unittest.TestCase):
                                                  params=params,
                                                  timeout=10)
         mocked_submitted.emit.assert_called_once_with(100)
+
+    def test_run_request_exception(self):
+        """Tests when a requests.exceptions.RequestException occurs."""
+        self.mocked_response.raise_for_status.side_effect = requests.exceptions.RequestException()
+        experimentArgs = {"arg1": "arg_value1", "arg2": "arg_value2"}
+        schedOpts = {"opt1": "opt_value1", "opt2": "opt_value2"}
+        callback = mock.MagicMock()
+        parent = QObject()
+        with mock.patch("iquip.apps.builder.ExperimentSubmitThread.submitted") as mocked_submitted:
+            thread = builder.ExperimentSubmitThread(
+                experimentPath="experiment_path",
+                experimentArgs=experimentArgs,
+                schedOpts=schedOpts,
+                callback=callback,
+                parent=parent
+            )
+            thread.run()
+            thread.wait()
+        params = {
+            "file": "experiment_path",
+            "args": json.dumps(experimentArgs),
+            "opt1": "opt_value1",
+            "opt2": "opt_value2"
+        }
+        self.mocked_get.assert_called_once_with("http://127.0.0.1:8000/experiment/submit/",
+                                                 params=params,
+                                                 timeout=10)
+        mocked_submitted.emit.assert_not_called()
 
 
 class BuilderAppTest(unittest.TestCase):
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -3,7 +3,7 @@
 import copy
 import json
 import unittest
-from typing import Callable
+from typing import Callable, Dict, Optional
 from unittest import mock
 
 import requests
@@ -55,28 +55,8 @@ class _BaseEntryTest(unittest.TestCase):
                 entry.value()
 
 
-class _BooleanEntryTest(unittest.TestCase):
-    """Unit tests for _BooleanEntry class."""
-
-
-class _EnumerationEntryTest(unittest.TestCase):
-    """Unit tests for _EnumerationEntry class."""
-
-
-class _NumberEntryTest(unittest.TestCase):
-    """Unit tests for _NumberEntry class."""
-
-
-class _StringEntryTest(unittest.TestCase):
-    """Unit tests for _StringEntry class."""
-
-
-class _DateTimeEntryTest(unittest.TestCase):
-    """Unit tests for _DateTimeEntry class."""
-
-
-class _ExperimentSubmitThreadTest(unittest.TestCase):
-    """Unit tests for _ExperimentSubmitThread class."""
+class ExperimentSubmitThreadTest(unittest.TestCase):
+    """Unit tests for ExperimentSubmitThread class."""
 
     def setUp(self):
         self.qapp = QApplication([])
@@ -135,20 +115,39 @@ class _ExperimentSubmitThreadTest(unittest.TestCase):
                                                  timeout=10)
         mocked_submitted.emit.assert_not_called()
 
+    def test_run_type_error(self):
+        """Tests when a TypeError occurs."""
+        callback = mock.MagicMock()
+        parent = QObject()
+        with mock.patch("iquip.apps.builder.ExperimentSubmitThread.submitted") as mocked_submitted:
+            experimentArgs = {"arg1": lambda: None}  # Not JSONifiable.
+            thread = get_thread(callback, parent, experimentArgs)
+            thread.run()
+            thread.wait()
+        params = {
+            "file": EXPERIMENT_PATH,
+            "args": json.dumps(EXPERIMENT_ARGS),
+            **SCHED_OPTS
+        }
+        self.mocked_get.assert_not_called()
+        mocked_submitted.emit.assert_not_called()
+
 
 def get_thread(
         callback: Callable[[int], None],
-        parent: QObject
+        parent: QObject,
+        experimentArgs: Optional[Dict[str, str]] = EXPERIMENT_ARGS
     ) -> builder.ExperimentSubmitThread:
     """Returns an ExperimentSubmitThread instance.
     
     Args:
         callback: The function called after the thread is done.
         parent: The parent object.
+        experimentArgs: The arguments of the experiment.
     """
     return builder.ExperimentSubmitThread(
         experimentPath=EXPERIMENT_PATH,
-        experimentArgs=EXPERIMENT_ARGS,
+        experimentArgs=experimentArgs,
         schedOpts=SCHED_OPTS,
         callback=callback,
         parent=parent
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -58,6 +58,7 @@ class BaseEntryTest(unittest.TestCase):
 class ExperimentSubmitThreadTest(unittest.TestCase):
     """Unit tests for ExperimentSubmitThread class."""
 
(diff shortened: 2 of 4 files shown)


tests/test_builder.py
```diff
@@ -18,33 +27,128 @@
...
+            thread.wait()
+        self.mocked_get.assert_not_called()
+        mocked_submitted.emit.assert_not_called()
 
 
-class ExperimentSubmitThreadTest(unittest.TestCase):
-    """Unit tests for ExperimentSubmitThread class."""
+def get_thread(
+        callback: Callable[[int], None],
+        parent: QObject,
```
kangz12345: Isn't it `Optional[QObject]`?

tests/test_builder.py
```diff
@@ -54,19 +158,17 @@ def setUp(self):
...
+            for type_ in ("Boolean", "String", "Enumeration", "Number", "DateTime")
         }
-        experiment_submit_thread_patcher = mock.patch("iquip.apps.builder.ExperimentSubmitThread")
         entries_patcher = mock.patch.multiple(
             "iquip.apps.builder",
             _BooleanEntry=self.mocked_entries["BooleanValue"],
             _StringEntry=self.mocked_entries["StringValue"],
             _EnumerationEntry=self.mocked_entries["EnumerationValue"],
-            _NumberEntry=self.mocked_entries["NumberValue"]
+            _NumberEntry=self.mocked_entries["NumberValue"],
```
kangz12345, line 169: Because of these changes, I prefer to use the trailing commas 😄

tests/test_builder.py
```diff
@@ -82,6 +184,90 @@ def test_init_args_entry(self):
...
+            listWidget.setItemWidget(item, widget)
+        app = builder.BuilderApp(
+            name="name",
+            experimentPath="experimentPath",
+            experimentClsName="experimentClsName",
+            experimentInfo=copy.deepcopy(EMPTY_EXPERIMENT_INFO)
+        )
+        args = app.argumentsFromListWidget(listWidget)
+        self.assertEqual(args["name1"], "value1")
+        self.assertNotIn("name2", args)
```
kangz12345: Hmm, why does `argumentsFromListWidget()` exclude `None` values from the returned dictionary? I don't remember the context.
I see the `_DateTimeEntry` may return `None`. Is it right to exclude the return value?
If so, please document this (`None` values are not included in the returned dictionary) in `argumentsFromListWidget()`.

tests/test_builder.py
```diff
@@ -82,6 +184,90 @@ def test_init_args_entry(self):
...
+        experimentArgs = {"key1": "value1"}
+        schedOpts = {"key2": "value2"}
+        with mock.patch.multiple(
+            app,
+            argumentsFromListWidget=mock.DEFAULT,
+            onSubmitted=mock.DEFAULT
+        ) as mocked:
+            mocked_arguments_from_list_widget = mocked["argumentsFromListWidget"]
+            mocked_on_submitted = mocked["onSubmitted"]
+            mocked_arguments_from_list_widget.side_effect=[experimentArgs, schedOpts]
```
kangz12345: Please insert spaces around `=`.

tests/.pylintrc
```diff
@@ -9,5 +9,6 @@ variable-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9
 
 [MESSAGES CONTROL]
 disable=
+    duplicate-code,
```
kangz12345: Is it because of the instantiation of builder apps?
I think we should not disable this warning even for tests (although we did in `qiwis`).
Because we can define non-testing methods whose names do not begin with "test_".
Just as you did for `get_thread()`.

kangz12345 (changes requested):

tests/test_builder.py
BECATRUE: Umm.. isn't it always given?
I think it is unnecessary to set `Optional[QObject]`.

tests/test_builder.py
BECATRUE: > why does argumentsFromListWidget() exclude None values from the returned dictionary? I don't remember the context.

This is because I want to use the default value of the parameter in the proxy server.
For example, if `_DateTimeEntry` returns `None`, the submission query doesn't contain the `timed` argument, and according to [ARTIQ-PROXY](https://github.com/snu-quiqcl/artiq-proxy/blob/develop/main.py#L117), it will be set to `None`.

In fact, it is okay to send the `None` value explicitly, and it is just my coding style😅

tests/.pylintrc
BECATRUE: I added this to ignore the PyQt warning in terms of the following code.
```python
def setUp(self):
        self.qapp = QApplication([])
        patcher = mock.patch("requests.get")
        self.mocked_get = patcher.start()
        self.mocked_response = self.mocked_get.return_value
        self.addCleanup(patcher.stop)
```

This method is located in both `test_explorer.py` and `test_builder.py`, thus it makes the warning.
Which solution is better?

tests/test_builder.py
kangz12345: Well, if you want to force to pass the `parent`, you can.
However, in this code, `parent` is only used as the `parent` parameter of `ExperimentSubmitThread`.
In its constructor, `parent` is `Optional[QObject]`.

tests/test_builder.py
kangz12345: In my opinion, that looks little bit weird because in `_DateTimeEntry`, `None` means that the checkbox is not checked, which is a valid and intended state.
If we remove this and rely on the default behavior of the proxy server, then we are ignoring user's intension.
Of course in this case, the default value is `None` so it would not make any problem, but this seems logically awkward.
Consider the case where the default behavior of the proxy is not equivalent to the `None` value after some updates (although I cannot imagine it, conceptually it is possible).

Therefore, I think we should pass `None` explicitly unless it has some downsides.

tests/.pylintrc
kangz12345: Oh, that was the problem.
Then how about adding pylint disable comments instead?
I think ignoring the `duplicate-code` warning at all seems too strong.

tests/test_builder.py
BECATRUE: Then, I agree with you that setting `parent` as `Optional[QObject]` is more natural.
I will apply it.

tests/test_builder.py
BECATRUE: Umm.. I think it's better to hear you say. I will apply it.

BECATRUE:
For resolving conflicts with `develop` branch, I pushed forcely.

BECATRUE:
I'm having a similar problem. It works fine on the server right now, so let's ignore it.

kangz12345:
> For resolving conflicts with `develop` branch, I pushed forcely.

I think in this case, resolving the merge conflicts was the better option.
We rebase when we need the updated main branch feature is required for my branch development.

Rebasing refreshes the commit history, so it is hard to track the code changes (I lose the point where I reviewed previously).

tests/test_builder.py
```diff
@@ -18,33 +27,129 @@
...
+            thread.run()
+            thread.wait()
+        self.mocked_get.assert_not_called()
+        mocked_submitted.emit.assert_not_called()
+
+
+def get_thread(
+        callback: Callable[[int], None],
+        parent: Optional[QObject] = None,
+        experimentArgs: Optional[Dict[str, str]] = None
```
kangz12345: I think it should be `Optional[Dict[str, Any]]`!

tests/test_builder.py
```diff
@@ -82,6 +185,90 @@ def test_init_args_entry(self):
...
+        self.mocked_entries["StringValue"].assert_any_call("pipeline", pipelineInfo)
+        self.mocked_entries["NumberValue"].assert_any_call("priority", priorityInfo)
+        self.mocked_entries["DateTimeValue"].assert_any_call("timed")
+
+    def test_arguments_from_list_widget(self):
+        listWidget = QListWidget()
+        for name, value in (
+            ("name1", "value1"),
+            ("name2", None),
+        ):
```
kangz12345: Doesn't it fit in a single line?

kangz12345 (changes requested):

tests/test_builder.py
BECATRUE: Oh, I wrote it in multiple lines just for legibility.
If you think it is okay to fit in a signle line, I will modify it.

kangz12345 (approved):
LGTM

giwon2004:
I will review this soon, quite a long code to read 😂

giwon2004:
Why does `$ coverage run --source=iquip/ -m unittest discover` gives

```
.........
.Failed to convert the build arguments to a JSON string.
...............
.............
----------------------------------------------------------------------
Ran 38 tests in 0.314s
```

rather than containing only 38 `.`s?

tests/test_builder.py
```diff
@@ -18,33 +27,129 @@
...
+        mocked_submitted.emit.assert_not_called()
 
-class _ExperimentSubmitThreadTest(unittest.TestCase):
-    """Unit tests for _ExperimentSubmitThread class."""
+    def test_run_type_error(self):
+        """Tests when a TypeError occurs."""
+        callback = mock.MagicMock()
+        parent = QObject()
+        with mock.patch("iquip.apps.builder._ExperimentSubmitThread.submitted") as mocked_submitted:
+            experimentArgs = {"arg1": lambda: None}  # Not JSONifiable.
```
giwon2004, line 124: This gives

```
.........
.Failed to convert the build arguments to a JSON string.
...............
.............
----------------------------------------------------------------------
Ran 38 tests in 0.314s
```

rather than containing only 38 `.`s when `$ coverage run --source=iquip/ -m unittest discover` is run.

To be neat for future unit tests, I guess there must be a way to redirect the `sys.output` here.

tests/test_builder.py
```diff
@@ -18,33 +27,129 @@
...
+            thread = get_thread(callback, parent, experimentArgs)
+            thread.run()
+            thread.wait()
+        self.mocked_get.assert_not_called()
+        mocked_submitted.emit.assert_not_called()
+
+
+def get_thread(
```
giwon2004, line 132: Why is this function located between two unit test classes? Is it because only `ExperimentSubmitThreadTest` uses this?

Maybe I am too used to compiled languages, but isn't it better to be above the class?

tests/test_explorer.py
```diff
@@ -15,6 +15,7 @@
 class FileFinderThreadTest(unittest.TestCase):
     """Unit tests for _FileFinderThread class."""
 
+    # pylint: disable=duplicate-code
```
giwon2004, line 18: Deleting this does not cause any problems in pylint. What is this comment for?

giwon2004 (changes requested):

tests/test_builder.py
BECATRUE: Then, did you mean that it would be better to mock `sys.output`?

tests/test_builder.py
BECATRUE: Umm we didn't decide this format explicitly, but I think it is our convention!

The `test_monitor.py` also located `get_ttl_monitor_widget_with()` at the end of the file and in `qiwis.py` in QIWIS repo, most functions are located at the end.

tests/test_explorer.py
BECATRUE: Really?

I added the comments in both `test_explorer.py` and `test_builder.py` because the `setUp()` methods are the same and caused the Pylint warning.

BECATRUE:
Oh, I sent @kangz12345 an unnecessary request. Sorry😂

tests/test_builder.py
kangz12345: https://docs.python.org/3/library/unittest.html#cmdoption-unittest-b

Will this resolve it?

BECATRUE: Wow, it may be a solution!

I think modifying the unittest yml file should be handled in another issue, thus please never mind it in this PR.

tests/test_builder.py
giwon2004: Okay, I will follow this convention too!

tests/test_explorer.py
giwon2004: Oh, really? I do not have any pylint disable comment in my branch for [PR#89](https://github.com/snu-quiqcl/iquip/pull/89), but it did not give any pylint warning.

BECATRUE: It is probably because `setUp()` in #89 is different from this one!

giwon2004 (approved):
LGTM!

PR #88, opened 2023-07-30
Aijuh/76/logs to file

### LINKED ISSUE #76: Add an option to save logs to file

Aijuh:
### Added feature for saving logs to file.

This is same PR as [previous](https://github.com/snu-quiqcl/iquip/pull/84). @BECATRUE leaved some comments on it.

This PR closes https://github.com/snu-quiqcl/iquip/issues/76
Below are images of current loggerFrame and saved log format.

All logs are first saved at temp file by using file handler. I first think that level for fileHandler should be different for frameHandler. But since user chooses desired level, we can just save that level of logs to file.

When user presses Save log to file button, contents at temp file saved to log record file and temp file is cleared.
When user presses Clear log file button, only log record file cleared.

The time format of logs in loggerFrame and file is different because we specify format for loggerFrame but use %asctime of format in fileHandler.

All the details can be changed.

image
image

### DIFF
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -158,8 +158,8 @@ class LoggerApp(qiwis.BaseApp):
                            logging.ERROR: "ERROR", logging.CRITICAL: "CRITICAL"}
         self.frameHandler = LoggingHandler(self.addLog)
         logFileName = os.path.join(path,
-                                   QDateTime.currentDateTime().toString("log_yyMMdd-HHmmss"))
-        self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName+".log", when="midnight"
+                                   QDateTime.currentDateTime().toString("log_yyMMdd-HHmmss.log"))
+        self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName, when="midnight"
                                                              ,encoding="utf-8")
         self.initLogger()
         # set loggerFrame's frameLevelBox
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -146,7 +146,7 @@ class LoggerApp(qiwis.BaseApp):
         """Extended.
         
         Args:
-            path: The path of the log record file.
+            logFilePath: The path of the log record file.
         """
         super().__init__(name, parent=parent)
         self.loggerFrame = LoggerFrame()
@@ -175,11 +175,7 @@ class LoggerApp(qiwis.BaseApp):
         fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
 
     def initLogger(self):
-        """Initializes the root logger and handlers.
-        
-        Args:
-            path: The path of the log record file.    
-        """
+        """Initializes the root logger and handlers."""
         def namer(name):
             return name.replace(".log", "") + ".log"
         self.fileHandler.suffix = "(%Y%m%d)"
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -200,6 +200,7 @@ class LoggerApp(qiwis.BaseApp):
         fileHandler: A handler for saving logs to file.
     """
 
+    # pylint: disable=too-many-instance-attributes
     def __init__(self, name: str, parent: Optional[QObject] = None):
         """Extended."""
         super().__init__(name, parent=parent)
@@ -226,7 +227,7 @@ class LoggerApp(qiwis.BaseApp):
         self.loggerFrame.levelBox.addItems(levels_dict.values())
         self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
         self.loggerFrame.levelBox.setCurrentText(levels_dict[self.frameHandler.level])
-    
+
     def initLogger(self):
         """Initializes the root logger and handlers for constructor."""
         # initialize handlers
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -154,32 +154,36 @@ class LoggerApp(qiwis.BaseApp):
         # connect signals to slots
         self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
         self.confirmFrame.confirmed.connect(self.clearLog)
-        self.initLogger(path)
-        # set loggerFrame's frameLevelBox
-        frameLevelBox = self.loggerFrame.frameLevelBox
-        frameLevelBox.addItems(self.levelsDict.values())
-        frameLevelBox.textActivated.connect(functools.partial(self.setLevel, self.frameHandler))
-        frameLevelBox.setCurrentText(self.levelsDict[self.frameHandler.level])
-        # set loggerFrame's fileLevelBox
-        fileLevelBox = self.loggerFrame.fileLevelBox
-        fileLevelBox.addItems(self.levelsDict.values())
-        fileLevelBox.textActivated.connect(functools.partial(self.setLevel, self.fileHandler))
-        fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
-
-    def initLogger(self, path: str):
-        """Initializes the root logger and handlers.
-        
-        Args:
-            path: The path of the log record file.    
-        """
         self.levelsDict = {logging.DEBUG: "DEBUG", logging.INFO: "INFO", logging.WARNING: "WARNING",
                            logging.ERROR: "ERROR", logging.CRITICAL: "CRITICAL"}
         self.frameHandler = LoggingHandler(self.addLog)
         logFileName = os.path.join(path,
                                    QDateTime.currentDateTime().toString("log_yyMMdd-HHmmss"))
-        self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName, when="midnight",
-                                                             encoding="utf-8")
-        self.fileHandler.suffix = "(%Y%m%d).log"
+        self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName+".log", when="M",
+                                                             interval = 1, encoding="utf-8")
+        self.initLogger()
+        # set loggerFrame's frameLevelBox
+        frameLevelBox = self.loggerFrame.frameLevelBox
+        frameLevelBox.addItems(self.levelsDict.values())
+        frameLevelBox.currentTextChanged.connect(functools.partial(self.setLevel,
+                                                                   self.frameHandler))
+        frameLevelBox.setCurrentText(self.levelsDict[self.frameHandler.level])
+        # set loggerFrame's fileLevelBox
+        fileLevelBox = self.loggerFrame.fileLevelBox
+        fileLevelBox.addItems(self.levelsDict.values())
+        fileLevelBox.currentTextChanged.connect(functools.partial(self.setLevel, self.fileHandler))
+        fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
+
+    def initLogger(self):
+        """Initializes the root logger and handlers.
+        
+        Args:
+            path: The path of the log record file.    
+        """
+        def namer(name):
+            return name.replace(".log", "") + ".log"
+        self.fileHandler.suffix = "(%Y%m%d)"
+        self.fileHandler.namer = namer
         shortFormat = "[%(name)s] %(message)s"
         longFormat = "%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
         self.frameHandler.setFormatter(logging.Formatter(shortFormat))
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -175,10 +175,11 @@ class LoggerApp(qiwis.BaseApp):
         self.levelsDict = {logging.DEBUG: "DEBUG", logging.INFO: "INFO", logging.WARNING: "WARNING",
                            logging.ERROR: "ERROR", logging.CRITICAL: "CRITICAL"}
         self.frameHandler = LoggingHandler(self.addLog)
-        logFileName = os.path.join(path, QDateTime.currentDateTime().toString("log_yyMMdd-HHmmss.log"))
+        logFileName = os.path.join(path, 
+                                   QDateTime.currentDateTime().toString("log_yyMMdd-HHmmss"))
         self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName, when="midnight",
                                                              interval=1, encoding="utf-8")
-        self.fileHandler.suffix = "(%Y%m%d)"
+        self.fileHandler.suffix = "(%Y%m%d).log"
         shortFormat = "[%(name)s] %(message)s"
         longFormat = "%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
         self.frameHandler.setFormatter(logging.Formatter(shortFormat))
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -219,6 +219,16 @@ class LoggerApp(qiwis.BaseApp):
         self.dirLogFIle = "log_record.txt"
         with open(self.dirTempLogFile, mode = "w", encoding = "utf-8"):
             pass
+        #initialize logger
+        self.initLogger()
+        # set loggerFrame's levelBox
+        levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
+        self.loggerFrame.levelBox.addItems(levels_dict.values())
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+        self.loggerFrame.levelBox.setCurrentText(levels_dict[self.frameHandler.level])
+    
+    def initLogger(self):
+        """Initializes the root logger and handlers for constructor."""
         # initialize handlers
         self.frameHandler = LoggingHandler(self.addLog)
         self.fileHandler = logging.FileHandler(self.dirTempLogFile)
@@ -232,11 +242,6 @@ class LoggerApp(qiwis.BaseApp):
         rootLogger.addHandler(self.frameHandler)
         rootLogger.addHandler(self.fileHandler)
         self.setLevel("WARNING")
-        # set loggerFrame's levelBox
-        levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
-        self.loggerFrame.levelBox.addItems(levels_dict.values())
-        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
-        self.loggerFrame.levelBox.setCurrentText(levels_dict[self.frameHandler.level])
 
     @pyqtSlot(str)
     def setLevel(self, levelText: str):
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -178,7 +178,7 @@ class LoggerApp(qiwis.BaseApp):
         logFileName = os.path.join(path,
                                    QDateTime.currentDateTime().toString("log_yyMMdd-HHmmss"))
         self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName, when="midnight",
-                                                             interval=1, encoding="utf-8")
+                                                             encoding="utf-8")
         self.fileHandler.suffix = "(%Y%m%d).log"
         shortFormat = "[%(name)s] %(message)s"
         longFormat = "%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
@@ -197,7 +197,7 @@ class LoggerApp(qiwis.BaseApp):
 
     @pyqtSlot(logging.Handler, str)
     def setLevel(self, handler_: logging.Handler, levelText: str):
-        """Changes the handler's level and update root logger's level.
+        """Changes the handler's level and updates root logger's level.
 
         Args:
             handler_: A Handler for which the level should be set. 
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -1,10 +1,10 @@
 """App module for log viewer in apps."""
 
+import os
+import functools
 import logging
 from logging import handlers
 from typing import Any, Optional, Tuple, Callable
-from functools import partial
-import os
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal, QDateTime
 from PyQt5.QtWidgets import (
@@ -146,7 +146,6 @@ class LoggerApp(qiwis.BaseApp):
         """Extended.
         
         Args:
-            name: Specifies name of the LoggerApp.
             path: Desribes file directory of log record file.
         """
         super().__init__(name, parent=parent)
@@ -166,12 +165,12 @@ class LoggerApp(qiwis.BaseApp):
         # set loggerFrame's frameLevelBox
         frameLevelBox = self.loggerFrame.frameLevelBox
         frameLevelBox.addItems(self.levelsDict.values())
-        frameLevelBox.textActivated.connect(partial(self.setLevel, self.frameHandler))
+        frameLevelBox.textActivated.connect(functools.partial(self.setLevel, self.frameHandler))
         frameLevelBox.setCurrentText(self.levelsDict[self.frameHandler.level])
         # set loggerFrame's fileLevelBox
         fileLevelBox = self.loggerFrame.fileLevelBox
         fileLevelBox.addItems(self.levelsDict.values())
-        fileLevelBox.textActivated.connect(partial(self.setLevel, self.fileHandler))
+        fileLevelBox.textActivated.connect(functools.partial(self.setLevel, self.fileHandler))
         fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
 
     def initLogger(self):
@@ -185,7 +184,7 @@ class LoggerApp(qiwis.BaseApp):
         rootLogger = logging.getLogger()
         rootLogger.addHandler(self.frameHandler)
         rootLogger.addHandler(self.fileHandler)
-        self.frameHandler.setLevel("WARNING")
+        self.setLevel(self.frameHandler, "WARNING")
         self.setLevel(self.fileHandler, "WARNING")
 
     def frames(self) -> Tuple[LoggerFrame]:
@@ -194,7 +193,7 @@ class LoggerApp(qiwis.BaseApp):
 
     @pyqtSlot(logging.Handler, str)
     def setLevel(self, handler_: logging.Handler, levelText: str):
-        """Responds to the loggerFrame's fileLevelBox widgets and changes the handler's level.
+        """Changes the handler's level and update root logger's level.
 
         Args:
             handler_: A Handler for which the level should be set. 
@@ -205,7 +204,7 @@ class LoggerApp(qiwis.BaseApp):
         if levelText in self.levelsDict.values():
             handler_.setLevel(levelText)
             lowerLevel = min(self.frameHandler.level, self.fileHandler.level)
-            logging.getLogger().setLevel(self.levelsDict[lowerLevel])
+            logging.getLogger().setLevel(lowerLevel)
 
     @pyqtSlot(str)
     def addLog(self, content: str):
--- a/iquip/apps/logger.py
+++ b/iquip/apps/logger.py
@@ -1,6 +1,7 @@
 """App module for log viewer in apps."""
 
 import logging
+import logging.handlers
 from typing import Any, Optional, Tuple, Callable
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal, QDateTime
@@ -59,10 +60,10 @@ class LoggerFrame(QWidget):
     Attributes:
         logEdit: A textEdit which shows all logs.
         clearButton: A button for clearing all logs.
-        levelBoxLabel: A label for describing levelBox.
-        levelBox: A comboBox for setting the logger's level.
-        fileWriteButton: A button for wrtie logs to file.
-        fileClearButton: A button for clear logs in file.
+        frameLevelBoxLabel: A label for describing frameLevelBox.
+        frameLevelBox: A comboBox for setting the logger's level for displaying logs to Loggerframe.
+        fileLevelBoxLabel: A label for describing fileLevelBox.
+        fileLevelBox: A comboBox for setting the file logger's level for saving logs to file.
     """
 
     def __init__(self, parent: Optional[QObject] = None):
@@ -72,18 +73,18 @@ class LoggerFrame(QWidget):
         self.logEdit = QTextEdit(self)
         self.logEdit.setReadOnly(True)
         self.clearButton = QPushButton("Clear", self)
-        self.levelBoxLabel = QLabel("Select log's level:")
-        self.levelBox = QComboBox(self)
-        self.fileWriteButton = QPushButton("Save log to file")
-        self.fileClearButton = QPushButton("Clear log file")
+        self.frameLevelBoxLabel = QLabel("Select screen log's level:")
+        self.frameLevelBox = QComboBox(self)
+        self.fileLevelBoxLabel = QLabel("Select file log's level:")
+        self.fileLevelBox = QComboBox(self)
         # layout
         layout = QGridLayout(self)
         layout.addWidget(self.logEdit, 0, 0, 1, 6)
         layout.addWidget(self.clearButton,1, 0, 1, 6)
-        layout.addWidget(self.levelBoxLabel, 2, 0, 1, 2)
-        layout.addWidget(self.levelBox, 2, 2, 1, 4)
-        layout.addWidget(self.fileWriteButton, 3, 0, 1, 3)
-        layout.addWidget(self.fileClearButton, 3, 3, 1, 3)
+        layout.addWidget(self.frameLevelBoxLabel, 2, 0, 1, 2)
+        layout.addWidget(self.frameLevelBox, 2, 2, 1, 4)
+        layout.addWidget(self.fileLevelBoxLabel, 3, 0, 1, 2)
+        layout.addWidget(self.fileLevelBox, 3, 2, 1, 4)
 
 
 class ConfirmClearingFrame(QWidget):
@@ -125,68 +126,6 @@ class ConfirmClearingFrame(QWidget):
         self.close()
 
 
-class ConfirmFileClearFrame(QWidget):
-    """A confirmation frame for clearing the log record file's log messages."""
-    confirmed = pyqtSignal()
-
-    def __init__(self, parent: Optional[QObject] = None):
-        """Extended."""
-        super().__init__(parent=parent)
-        # widgets
-        self.label = QLabel("Are you sure to clear log file?")
-        self.buttonBox = QDialogButtonBox()
-        self.buttonBox.addButton("OK", QDialogButtonBox.AcceptRole)
-        self.buttonBox.addButton("Cancel", QDialogButtonBox.RejectRole)
-        # connect signals
-        self.buttonBox.accepted.connect(self.buttonOKClicked)
-        self.buttonBox.rejected.connect(self.buttonCancelClicked)
-        # layouts
-        layout = QVBoxLayout()
-        self.setLayout(layout)
-        layout.addWidget(self.label)
-        layout.addWidget(self.buttonBox)
-
-    def buttonOKClicked(self):
-        """Clicked OK button to clear the log record file."""
-        self.confirmed.emit()
-        self.close()
-
-    def buttonCancelClicked(self):
-        """Clicked Cancel button not to clear the log record file."""
-        self.close()
-
-
-class ConfirmFileWriteFrame(QWidget):
-    """A confirmation frame for writing log messages to the log record file."""
-    confirmed = pyqtSignal()
-
-    def __init__(self, parent: Optional[QObject] = None):
-        """Extended."""
-        super().__init__(parent=parent)
-        # widgets
-        self.label = QLabel("Are you sure to write log messages to log file?")
-        self.buttonBox = QDialogButtonBox()
-        self.buttonBox.addButton("OK", QDialogButtonBox.AcceptRole)
-        self.buttonBox.addButton("Cancel", QDialogButtonBox.RejectRole)
-        # connect signals
-        self.buttonBox.accepted.connect(self.buttonOKClicked)
-        self.buttonBox.rejected.connect(self.buttonCancelClicked)
-        # layouts
-        layout = QVBoxLayout()
-        self.setLayout(layout)
-        layout.addWidget(self.label)
-        layout.addWidget(self.buttonBox)
-
-    def buttonOKClicked(self):
-        """Clicked OK button to wrtie log messages to the log record file."""
-        self.confirmed.emit()
-        self.close()
-
-    def buttonCancelClicked(self):
-        """Clicked Cancel button not to write log messages to the log record file."""
-        self.close()
-
-
 class LoggerApp(qiwis.BaseApp):
     """App for logging.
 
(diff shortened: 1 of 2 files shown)


BECATRUE:
Please fix the Pylint warnings.

You can use `# pylint: disable=too-many-arguments` comment to ignore that warning.
Please see `ExperimentSubmitThread.__init__()` in `builder.py`.

Aijuh:
I added `# pylint: disable=too-many-instance-attributes` above `__init__` of LoggerFrame app to solve pylint error. I am not sure whether it is gramatically correct, but it prevents pylint error.

kangz12345:
Hmm.. I think we did not agree enough with each other about the goal. I am sorry for checking the implementation plan thoroughly.

I thought the file logger logs unconditionally, all the time. Therefore, its file path must be unique for every program run (for example, using the date time).
This is very important because the program might crash and terminate unexpectedly, and the user may re-run the program immediately. Then the program will overwrite the logs on "temp.txt" so the previous logs will disappear.
So in my opinion, the following policy would be great:
1. Creates a directory "yyyy/mm/dd" (3-depth structure)
2. Log into that directory using [`TimedRotatingFileHandler`](https://docs.python.org/3/library/logging.handlers.html#timedrotatingfilehandler) since we usually do not turn off the GUI program. After the midnight, we want to log into the next day's directory.

Let's discuss this further to clarify what to implement. Please ask me if you find something not certain.

kangz12345:
In addition, I expected two combo boxes to set the handler levels.

BECATRUE:
I have several questions.

1. Handling the log file
I think it is natural to save logs to a file consistently in the background without clicking a button.
In addition, I think there are too many button in the `loggerFrame` so that it looks convoluted.
But, I wonder How @kangz12345 think of it.

2. Setting the handler level
Due to the same reason above, I think we should set the file logger level to DEBUG or INFO constantly.

Aijuh:
As I understand, I will implement followings.

1. At the start of LoggerApp, make file directory `yyyy/mm/dd` using today's date. ( Will use` os.mkdir("./log-record/y/m/d")`.)
2. Unconditionally append all the logs to the today's file directory. 
3. Whenever date changes, make new file directory and move to that directory.
4. Delete button for write and clear logs in file and add option for select log level for file saving.

But for short search, TimedRotatingFileHandler only suffices designated format when given interval ends. Therefore it is easy to save many text file as "log-y-m-d" in one file directory. But we cannot check whether date changes outside that handler to save at file directory "/y/m/d". I will first do more search, but I wonder does @kangz12345 have any idea of log in to today's date file using TimedRotatingFileHandler.

kangz12345:
> 1. Handling the log file

In my opinion, there should be 2 comboboxes for selecting the levels and 1 button for clearing the logs.
However, since now we log to files as well, I think we can just clear the log without asking when the button is clicked.

> 2. Setting the handler level

In my opinion, as I said before, it should be set to `min(guiLoggerLevel, fileLoggerLevel)`.

kangz12345:
> 1. At the start of LoggerApp, make file directory `yyyy/mm/dd` using today's date. ( Will use` os.mkdir("./log-record/y/m/d")`.)

Instead of using the current directory ".", it is better to parameterize the log directory, so it can be set using a config file, for example.
If you do it with ".", the log directory can be changed depending on where the user runs the program.

> 2. Unconditionally append all the logs to the today's file directory.

Yes, but the user can still set the handler's level.

> 3. Whenever date changes, make new file directory and move to that directory.

Yes, I thought the `TimedRotatingFileHandler` would provide that feature, but it seems not as you said.

> 4. Delete button for write and clear logs in file and add option for select log level for file saving.

Exactly!

> But for short search, TimedRotatingFileHandler only suffices designate format when given interval ends. Therefore it is easy to save many text file as "log-y-m-d" in one file directory. But we cannot check whether date changes outside that handler to save at file direction "/y/m/d". I will first do more search, but I wonder does @kangz12345 have any idea of log in to today's date file using TimedRotatingFileHandler.

As you said, it only provides the suffix. Then we can just give up the directory structure and put all the logs in a flat directory.
However, to prevent the log flood, we can give the `backupCount` parameter to, e.g., `90`.
If some logs are important, users will copy it to prevent from vanishing.

Aijuh:
Updated codes.

 I didn't check it for midnight, but if I set interval of `TimedRotatingFileHandler` as minute, I checked that after one minute previous file **log** get suffix **-yyyymmdd** and log is saved to original file **log**.
After two minutes, contents of **log** file saves to **log-yyyymmdd** and previous content of **log-yyyymmdd** removed.

I think it is good to set file format as text file. But when we set original file name as **log.txt** , and suffix to **-yyyymmdd** the file after midnight becomes **log.txt-yyyymmdd** which is not the form of text file. It doesn't seem natural. 
So I think we should not set original file name as **txt** format and add **txt** in suffix. (For example original name as **log** and suffix name as **-yyyy/mm/dd.txt**.)

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        self.loggerFrame.fileLevelBox.textActivated.connect(self.setFileLevel)
+        self.loggerFrame.fileLevelBox.setCurrentText(self.levels_dict[self.fileHandler.level])
+
+    def initLogger(self):
+        """Initializes the root logger and handlers for constructor."""
+        self.levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
+        # initialize handlers
+        self.frameHandler = LoggingHandler(self.addLog)
+        self.fileHandler = handlers.TimedRotatingFileHandler(filename = "log", when = "midnight",
+                                                             interval = 1, encoding = 'utf-8')
```
kangz12345: All the declarations (initial assignments) must reside in `__init__()` for readability.
Moreover, they must be documented in the class docstring (`Attributes:` section).

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
         self.confirmFrame.confirmed.connect(self.clearLog)
-        self.handler = LoggingHandler(self.addLog)
-        # TODO(aijuh): Change the log format when it is determined.
-        fs = "%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
-        formatter = logging.Formatter(fs)
-        self.handler.setFormatter(formatter)
+        #initialize logger
+        self.initLogger()
+        # set loggerFrame's frameLevelBox
+        self.loggerFrame.frameLevelBox.addItems(self.levels_dict.values())
```
kangz12345: Since `levels_dict` is defined in `initLogger()`, code readers cannot find it unless they read the `initLogger()`.

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        self.loggerFrame.frameLevelBox.textActivated.connect(self.setFrameLevel)
+        self.loggerFrame.frameLevelBox.setCurrentText(self.levels_dict[self.frameHandler.level])
+        # set loggerFrame's fileLevelBox
+        self.loggerFrame.fileLevelBox.addItems(self.levels_dict.values())
+        self.loggerFrame.fileLevelBox.textActivated.connect(self.setFileLevel)
+        self.loggerFrame.fileLevelBox.setCurrentText(self.levels_dict[self.fileHandler.level])
+
+    def initLogger(self):
+        """Initializes the root logger and handlers for constructor."""
+        self.levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
```
kangz12345: Instead of using the bare number values, please use the constants in `logging` module. For example, `{logging.DEBUG: "DEBUG", ...}`.
Of coarse the constant values won't be changed, but it is still more robust to module updates and more intuitive (code readers who do not know about `logging` module would be confused about the reason why these numbers are used).

iquip/apps/logger.py
kangz12345: When you write keyword arguments, you don't put spaces besides `=`.
For example, `f(a=0, b=1)` instead of `f(a = 0, b = 1)`.
In addition, please use double quotes `"` instead of single quotes `'`.

kangz12345: I requested to use `backupCount` option, but I think we don't have to! Never mind that option.

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        # initialize handlers
+        self.frameHandler = LoggingHandler(self.addLog)
+        self.fileHandler = handlers.TimedRotatingFileHandler(filename = "log", when = "midnight",
+                                                             interval = 1, encoding = 'utf-8')
+        self.fileHandler.suffix = '-%Y%m%d'
```
kangz12345: Good! Please use double quotes `"`.

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        self.loggerFrame.fileLevelBox.addItems(self.levels_dict.values())
+        self.loggerFrame.fileLevelBox.textActivated.connect(self.setFileLevel)
+        self.loggerFrame.fileLevelBox.setCurrentText(self.levels_dict[self.fileHandler.level])
+
+    def initLogger(self):
+        """Initializes the root logger and handlers for constructor."""
+        self.levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
+        # initialize handlers
+        self.frameHandler = LoggingHandler(self.addLog)
+        self.fileHandler = handlers.TimedRotatingFileHandler(filename = "log", when = "midnight",
```
kangz12345: I would like to use the filename as follows:
1. Take the log directory path by adding a parameter in `__init__()`. For example, `LoggerApp(name="logger", path="logs", parent=None)`.
2. Include the date time when the program starts (the log file is created).

Therefore, an example filename would be `logs/230801-100230.log`, which means the log started at 2023.08.01 10:02:30.
Maybe the suffix would be the date, so the next day's rollover would be `logs/230801-100230(230802).log`.

iquip/apps/logger.py
kangz12345: How about the form `"(%Y%m%d)"`?

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+                                                             interval = 1, encoding = 'utf-8')
+        self.fileHandler.suffix = '-%Y%m%d'
+        simpleFormat = "[%(name)s] %(message)s"
+        complexFormat = "%(asctime)s %(levelname)s [%(name)s]"\
+                        " [%(filename)s:%(lineno)d] %(message)s"
```
kangz12345: 1. Please avoid using back slashes.
2. In this case, you can just write it in a single line. I checked that it takes 101 columns. 1 column over is okay here since it is more readable!
3. Or you can use shorter variable names, e.g., `shortFormat` and `longFormat`.

Please choose one between 2 and 3.

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
-        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
-        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
-        self.loggerFrame.levelBox.setCurrentText("WARNING")
+        rootLogger.addHandler(self.frameHandler)
+        rootLogger.addHandler(self.fileHandler)
+        self.fileHandler.setLevel("WARNING")
+        self.setFrameLevel("WARNING")
+        self.setFileLevel("WARNING")
+
+
```
kangz12345: Just a single blank line between methods.

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        # set rootLogger
         rootLogger = logging.getLogger()
-        rootLogger.addHandler(self.handler)
-        self.setLevel("WARNING")
-        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
-        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
-        self.loggerFrame.levelBox.setCurrentText("WARNING")
+        rootLogger.addHandler(self.frameHandler)
+        rootLogger.addHandler(self.fileHandler)
+        self.fileHandler.setLevel("WARNING")
```
kangz12345: I think you should set the `rootLogger` level here since you set the others below.

iquip/apps/logger.py
kangz12345: Please use the consistent naming convention: `levelsDict`.
I think just `levels` or `levelNames` is fine, if it is okay with you.

iquip/apps/logger.py
```diff
@@ -136,41 +147,74 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
-        }
-        if levelText in levels:
-            self.handler.setLevel(levels[levelText])
-            logging.getLogger().setLevel(levels[levelText])
+        if levelText in self.levels_dict.values():
+            self.frameHandler.setLevel(levelText)
+            if self.frameHandler.level < self.fileHandler.level:
+                logging.getLogger().setLevel(levelText)
+            else:
+                logging.getLogger().setLevel(self.levels_dict[self.fileHandler.level])
```
kangz12345: How about:
```python
lowerLevel = min(self.frameHandler.level, self.fileHandler.level)
logging.getLogger().setLevel(self.levels_dict[lowerLevel])
```

kangz12345: If you do this, `setFrameLevel()` and `setFileLevel()` will become almost identical, except for the handler (`frameHandler` and `fileHandler`, respectively).
Then you can merge them as (for example):
```python
def setLevel(self, handler_: logging.Handler, levelText: str):
    ...
```
and then use [`functools.partial()`](https://docs.python.org/3/library/functools.html#functools.partial) when you connect them as slots.

iquip/apps/logger.py
```diff
@@ -184,11 +228,11 @@ def addLog(self, content: str):
 
     @pyqtSlot()
     def checkToClear(self):
-        """Shows a confirmation frame for clearing logs."""
+        """Shows a confirmation frame for clearing logs in the loggerFrame."""
         logger.info("Tried to clear logs by clicking clear button")
         self.confirmFrame.show()
```
kangz12345, line 228: I think we don't need this anymore since we log to the files.

kangz12345 (changes requested):

iquip/apps/logger.py
Aijuh: I didn't understand your request at that time and forgot about it.
I understand that as maximum number of files that handler create.
Maybe we can add more feature later.

iquip/apps/logger.py
Aijuh: You mean not using `confirmFrame`?

Aijuh: I understand that we can clear logs in screen right away since we are writing at file.
But I want to make sure before delete that feature.

iquip/apps/logger.py
```diff
@@ -135,86 +136,71 @@ class LoggerApp(qiwis.BaseApp):
     Attributes:
         loggerFrame: A frame that shows the logs.
         confirmFrame: A frame that asks whether to clear logs.
+        levelsDict: A dictionary that matches int level to string level.
         frameHandler: A handler for adding logs to the loggerFrame.
         fileHandler: A handler for saving logs to file.
     """
 
-    def __init__(self, name: str, parent: Optional[QObject] = None):
+    def __init__(self, name: str = "logger", path: str = "logs", parent: Optional[QObject] = None):
```
kangz12345: I think `name` and `path` should not have default values.

iquip/apps/logger.py
```diff
@@ -135,86 +136,71 @@ class LoggerApp(qiwis.BaseApp):
...
         """Extended."""
```
kangz12345: Docstring for `path` argument.

iquip/apps/logger.py
kangz12345: Hmm, yeah but I think it is out of this PR's scope! Let's focus on the file-logging system for now.

iquip/apps/logger.py
```diff
@@ -125,52 +136,71 @@ class LoggerApp(qiwis.BaseApp):
...
-        self.handler = LoggingHandler(self.addLog)
-        # TODO(aijuh): Change the log format when it is determined.
-        fs = "%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
-        formatter = logging.Formatter(fs)
-        self.handler.setFormatter(formatter)
+        # initialize handlers
+        self.levelsDict = {logging.DEBUG: "DEBUG", logging.INFO: "INFO", logging.WARNING: "WARNING",
+                           logging.ERROR: "ERROR", logging.CRITICAL: "CRITICAL"}
+        self.frameHandler = LoggingHandler(self.addLog)
+        logFileName = path + QDateTime.currentDateTime().toString("yyMMdd-HHmmss")
```
kangz12345: Instead of `+`, please use [`pathlib`](https://docs.python.org/3/library/pathlib.html) or [`os.path.join()`](https://docs.python.org/3/library/os.path.html#os.path.join).
`path` could be like `"dir1/dir2/"` or `"dir1/dir2"`.

iquip/apps/logger.py
```diff
@@ -125,52 +136,71 @@ class LoggerApp(qiwis.BaseApp):
...
+        frameLevelBox.setCurrentText(self.levelsDict[self.frameHandler.level])
+        # set loggerFrame's fileLevelBox
+        fileLevelBox = self.loggerFrame.fileLevelBox
+        fileLevelBox.addItems(self.levelsDict.values())
+        fileLevelBox.textActivated.connect(partial(self.setLevel, self.fileHandler))
+        fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
+
+    def initLogger(self):
+        """Initializes the root logger and handlers for constructor."""
+        self.fileHandler.suffix = '-(%Y%m%d)'
```
kangz12345: 1. Double quotes `"`
2. Isn't it better to remove `-`?

iquip/apps/logger.py
```diff
@@ -125,52 +136,71 @@ class LoggerApp(qiwis.BaseApp):
...
-        rootLogger.addHandler(self.handler)
-        self.setLevel("WARNING")
-        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
-        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
-        self.loggerFrame.levelBox.setCurrentText("WARNING")
+        rootLogger.addHandler(self.frameHandler)
+        rootLogger.addHandler(self.fileHandler)
+        self.frameHandler.setLevel("WARNING")
+        self.fileHandler.setLevel("WARNING")
+        rootLogger.setLevel("WARNING")
```
kangz12345: I am sorry, you don't have to set its level since `setLevel()` sets the root logger's level as well.

iquip/apps/logger.py
```diff
@@ -125,52 +136,71 @@ class LoggerApp(qiwis.BaseApp):
...
 
-    @pyqtSlot(str)
-    def setLevel(self, levelText: str):
-        """Responds to the loggerFrame's levelBox widget and changes the handler's level.
+    def frames(self) -> Tuple[LoggerFrame]:
+        """Overridden."""
+        return (self.loggerFrame,)
+
+    @pyqtSlot(logging.Handler, str)
```
kangz12345, line 199: Does this work? I am just curious. No error occurs?

kangz12345 (changes requested):

Aijuh:
If we does not set default value of path in line 145, the program does not work.
It is because qiwis.py calls it without path variable. I think we  should update qiwis.py first. 

I tested path format of line 162 and it worked well.

iquip/apps/logger.py
Aijuh: Yes It seems there is no problem.

kangz12345:
> If we does not set default value of path in line 145, the program does not work. It is because qiwis.py calls it without path variable. I think we should update qiwis.py first.

You can modify `setup.json` file to give the parameters.
Please refer to [`qiwis/examples/config.json`](https://github.com/snu-quiqcl/qiwis/blob/develop/examples/config.json)

Aijuh:
I will check it.

kangz12345:
Oh I didn't see this comment:

> I didn't check it for midnight, but if I set interval of `TimedRotatingFileHandler` as minute, I checked that after one minute previous file **log** get suffix **-yyyymmdd** and log is saved to original file **log**. After two minutes, contents of **log** file saves to **log-yyyymmdd** and previous content of **log-yyyymmdd** removed.

Is the previous log file removed? What was the `backupCount` value?

> I think it is good to set file format as text file. But when we set original file name as **log.txt** , and suffix to **-yyyymmdd** the file after midnight becomes **log.txt-yyyymmdd** which is not the form of text file. It doesn't seem natural. So I think we should not set original file name as **txt** format and add **txt** in suffix. (For example original name as **log** and suffix name as **-yyyy/mm/dd.txt**.)

That's hilarious. Then I think there are two workarounds:
1. Do not use the file extension
2. I think people use `.log` extension for logs, so use `.log` and do some work with [`namer`](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.BaseRotatingHandler.namer) attribute of the handler. See also: [cookbook](https://docs.python.org/3/howto/logging-cookbook.html#using-a-rotator-and-namer-to-customize-log-rotation-processing).

Aijuh:
The log file is removed because name collision occures. If we set file name as interval of day and update file in minute, name collision occurs and previous file deleted. If we update file at midnight, this would not happen. 

Then I will just use **.log** instead of **.txt**.

kangz12345:
> The log file is removed because name collision occures. If we set file name as interval of day and update file in minute, name collision occurs and previous file deleted. If we update file at midnight, this would not happen.

Oh I see. That's good.

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
+        fileHandler: A handler for saving logs to file.
     """
 
-    def __init__(self, name: str, parent: Optional[QObject] = None):
-        """Extended."""
+    def __init__(self, name: str, path: str, parent: Optional[QObject] = None):
+        """Extended.
+        
+        Args:
+            name: Specifies name of the LoggerApp.
```
kangz12345: Please omit the inherited argument `name`, since its role did not change.
See also: snu-quiqcl/qiwis#44.

iquip/apps/logger.py
```diff
@@ -1,11 +1,14 @@
 """App module for log viewer in apps."""
 
 import logging
+from logging import handlers
 from typing import Any, Optional, Tuple, Callable
+from functools import partial
```
kangz12345: In most cases, importing a function instead of its module should be avoided:
```python
import functools

functools.partial(...)
```
Although sometimes we use it for some popular functions for convenience.
Of course `partial()` is popular enough, but let's keep our convention.

iquip/apps/logger.py
```diff
@@ -1,11 +1,14 @@
...
+import os
```
kangz12345: Please put `import` lines before `from ... import` lines.

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
+        # set rootLogger
         rootLogger = logging.getLogger()
-        rootLogger.addHandler(self.handler)
-        self.setLevel("WARNING")
-        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
-        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
-        self.loggerFrame.levelBox.setCurrentText("WARNING")
+        rootLogger.addHandler(self.frameHandler)
+        rootLogger.addHandler(self.fileHandler)
+        self.frameHandler.setLevel("WARNING")
```
kangz12345: I think using `self.setLevel()` seems more consistent.

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
-    @pyqtSlot(str)
-    def setLevel(self, levelText: str):
-        """Responds to the loggerFrame's levelBox widget and changes the handler's level.
+    def frames(self) -> Tuple[LoggerFrame]:
+        """Overridden."""
+        return (self.loggerFrame,)
+
+    @pyqtSlot(logging.Handler, str)
+    def setLevel(self, handler_: logging.Handler, levelText: str):
+        """Responds to the loggerFrame's fileLevelBox widgets and changes the handler's level.
```
kangz12345: The summary does not have to include "Responds to ...".
The essence is that it "Changes the handler's level".

kangz12345: Note that this is called in `__init__()` as well. As you can see here, we usually define _what_ a function or method does, rather than _when_ it is called.

kangz12345 (changes requested):
Did you found how to change the file extension? Or, did you decide not to use `.log`?

iquip/apps/logger.py
```diff
@@ -26,7 +29,7 @@ class _Signaller(QObject):
 class LoggingHandler(logging.Handler):
     """Handler for logger.
 
-    Sends a log message to the connected function through a signal. 
+    Sends a log message to the connected function through a signal.
```
BECATRUE, line 32: It is just a question! Is there a trailing space originally?

BECATRUE:
> This PR closes issue https://github.com/snu-quiqcl/iquip/issues/76

I think this statment will not close that issue automatically.

Please write "close" and the corresponding issue in a row, referred to [GitHub](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
         confirmFrame: A frame that asks whether to clear logs.
-        handler: A handler for adding logs to the loggerFrame. 
+        levelsDict: A dictionary that matches int level to string level.
+        frameHandler: A handler for adding logs to the loggerFrame.
+        fileHandler: A handler for saving logs to file.
     """
 
-    def __init__(self, name: str, parent: Optional[QObject] = None):
-        """Extended."""
+    def __init__(self, name: str, path: str, parent: Optional[QObject] = None):
```
BECATRUE: I think it is better that the `path` has a default value such as ".".

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
+        """Extended.
+        
+        Args:
+            name: Specifies name of the LoggerApp.
+            path: Desribes file directory of log record file.
```
BECATRUE: How about "The path of log recodrd file".
I recommend to write it in noun form.

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
+        frameLevelBox.textActivated.connect(partial(self.setLevel, self.frameHandler))
+        frameLevelBox.setCurrentText(self.levelsDict[self.frameHandler.level])
+        # set loggerFrame's fileLevelBox
+        fileLevelBox = self.loggerFrame.fileLevelBox
+        fileLevelBox.addItems(self.levelsDict.values())
+        fileLevelBox.textActivated.connect(partial(self.setLevel, self.fileHandler))
+        fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
+
+    def initLogger(self):
+        """Initializes the root logger and handlers for constructor."""
```
BECATRUE: I think "for constructor" looks unnecessary.

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
-        self.handler = LoggingHandler(self.addLog)
-        # TODO(aijuh): Change the log format when it is determined.
-        fs = "%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
-        formatter = logging.Formatter(fs)
-        self.handler.setFormatter(formatter)
+        # initialize handlers
+        self.levelsDict = {logging.DEBUG: "DEBUG", logging.INFO: "INFO", logging.WARNING: "WARNING",
+                           logging.ERROR: "ERROR", logging.CRITICAL: "CRITICAL"}
+        self.frameHandler = LoggingHandler(self.addLog)
+        logFileName = os.path.join(path, QDateTime.currentDateTime().toString("logyyMMdd-HHmmss"))
```
BECATRUE: How about adding "_" between "log" and the datetime?

BECATRUE: Also, please use the extension ".log".

BECATRUE: Also, you set the `when` parameter of the file handler to "midnight", so isn't it unnecessary the time part in file name.
Isn't "log_yyMMdd.log" enough?

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
+        self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName, when="midnight",
+                                                             interval=1, encoding="utf-8")
```
BECATRUE: The default value of `interval" is 1, but did you intend to write it down explicitly?

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
+        self.initLogger()
+        # set loggerFrame's frameLevelBox
+        frameLevelBox = self.loggerFrame.frameLevelBox
+        frameLevelBox.addItems(self.levelsDict.values())
+        frameLevelBox.textActivated.connect(partial(self.setLevel, self.frameHandler))
```
BECATRUE: I think `currentTextChanged` is more appropriate that `textActivated`.
Referred to [PyQt docs](https://doc.qt.io/qt-5/qcombobox.html#textActivated), please choose the best one.

iquip/apps/logger.py
```diff
@@ -125,52 +137,75 @@ class LoggerApp(qiwis.BaseApp):
...
+        frameLevelBox.setCurrentText(self.levelsDict[self.frameHandler.level])
+        # set loggerFrame's fileLevelBox
+        fileLevelBox = self.loggerFrame.fileLevelBox
+        fileLevelBox.addItems(self.levelsDict.values())
+        fileLevelBox.textActivated.connect(partial(self.setLevel, self.fileHandler))
+        fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
+
+    def initLogger(self):
```
BECATRUE, line 177: If you use `initLogger` method, then it looks better to move the parts initializing loggers in the constructor into this method.

BECATRUE (changes requested):

iquip/apps/logger.py
Aijuh: I cannot remember whether I actually fix this part. I am not sure..

iquip/apps/logger.py
Aijuh: I am not sure about the naming convention about time.
As I turn on and off qiwis while testing, it seems helpful to recognize current file.
But in real application, I understand that qiwis is keep running and not turn off. In this case time part seems unnecessary.

iquip/apps/logger.py
Aijuh: I tested for multiple options of time unit and used that part. I will discard when the format is fixed.

iquip/apps/logger.py
kangz12345: I requested to remove the default value because the "current directory" depends on where the program is executed.

iquip/apps/logger.py
kangz12345: The file name is created when the program starts, i.e., the filename shows the start time. Note that we can run the program several times in a single day.

Aijuh:
I moved all handler code from constructor to `initLogger()` method in `LoggerApp` class.

Then, `fileHandler` should know path of log record file and `initLogger()` should take path as an argument.
Therefore comments for argument duplicated in LoggerApp and initLogger. I am not sure whether it is desirable.

iquip/apps/logger.py
```diff
@@ -194,7 +197,7 @@ def frames(self) -> Tuple[LoggerFrame]:
 
     @pyqtSlot(logging.Handler, str)
     def setLevel(self, handler_: logging.Handler, levelText: str):
-        """Responds to the loggerFrame's fileLevelBox widgets and changes the handler's level.
+        """Changes the handler's level and update root logger's level.
```
kangz12345: ... updates ...

iquip/apps/logger.py
kangz12345: Is `interval=1` necessary?

iquip/apps/logger.py
kangz12345: I requested to define all the attributes in `__init__()`.
@Aijuh You can discuss why you did it that way or ask me about this.

In addition, "initialize" and "instantiate" are different.

iquip/apps/logger.py
```diff
@@ -205,7 +208,7 @@ def setLevel(self, handler_: logging.Handler, levelText: str):
         if levelText in self.levelsDict.values():
             handler_.setLevel(levelText)
             lowerLevel = min(self.frameHandler.level, self.fileHandler.level)
-            logging.getLogger().setLevel(self.levelsDict[lowerLevel])
+            logging.getLogger().setLevel(lowerLevel)
```
kangz12345, line 212: 👍

kangz12345 (changes requested):

kangz12345:
We had an issue about the file extension:
- if we set the extension in the filename, the rollover becomes like "~.log(yyMMdd)"
- if we set the extension in the suffix, the main log file becomes like "~" (without any extension)

What was your decision to resolve this?

iquip/apps/logger.py
Aijuh: I separated constructor into two parts because constructor looked to long. 
@kangz12345 is it okay to move all the codes into constructor?

In addition, since log and handler class became an specific object, instantiate was right in this case.

Aijuh:
What we can do is setting main log file as "\~" and setting suffix as  "~(yyMMdd).log". 

But my original thought is to get signal when TimedRotateHandler roll over occured, and update manually at designated time. 
For example at midnight, file name changed from "log.log" to "log.(yymmdd).log".

But I cannot find any signal in TimedRotateHandler or additional option. So I am considering to make alarm clock and update at specific time. But it requires continuous check and extra work. 

So option is as follows.

1. Just set main file as "\~" and suffix as "~(yymmdd).log"
2. Add alarm to update file name manually.

Which one do you prefer?

iquip/apps/logger.py
BECATRUE: > Note that we can run the program several times in a single day.

@kangz12345 
Then, why do we set `when` parameter of `TimedRotatingFileHandler` to "midnight"?
If the log file is created everytime the program starts, I think the `TimedRotatingFileHandler` is unnecessary.

iquip/apps/logger.py
BECATRUE: Please check this!

Aijuh: Sorry I had many issues and forgot it. I will check it.

iquip/apps/logger.py
kangz12345: Because we usually don't terminate the GUI program. Then a single log file would become too large, which makes it difficult to find the logs for specific date and time.

kangz12345:
> > I think it is good to set file format as text file. But when we set original file name as **log.txt** , and suffix to **-yyyymmdd** the file after midnight becomes **log.txt-yyyymmdd** which is not the form of text file. It doesn't seem natural. So I think we should not set original file name as **txt** format and add **txt** in suffix. (For example original name as **log** and suffix name as **-yyyy/mm/dd.txt**.)
> 
> That's hilarious. Then I think there are two workarounds:
> 
> 1. Do not use the file extension
> 2. I think people use `.log` extension for logs, so use `.log` and do some work with [`namer`](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.BaseRotatingHandler.namer) attribute of the handler. See also: [cookbook](https://docs.python.org/3/howto/logging-cookbook.html#using-a-rotator-and-namer-to-customize-log-rotation-processing).

Did you check this out?

iquip/apps/logger.py
Aijuh: `currentTextChanged` is more appropriate because it emits signal only when current item is changed. I will fix it in this way.

iquip/apps/logger.py
Aijuh: I think it is not. I deleted it.

Aijuh:
I first thought that `namer` is not different from using suffix because it appends string to file name.
I will further check options inside handler.

kangz12345:
One suggestion:
1. Set the main name as "~.log"
2. Define `namer(name)` to return `name.replace(".log", "") + ".log"`

iquip/apps/logger.py
BECATRUE: Okay! Now I see.

BECATRUE:
Could you leave a screenshot of the latest version?

I can check it in my local, but It would be nice to post it on GitHub.

iquip/apps/logger.py
```diff
@@ -69,12 +75,18 @@ def __init__(self, parent: Optional[QObject] = None):
         self.logEdit = QTextEdit(self)
         self.logEdit.setReadOnly(True)
         self.clearButton = QPushButton("Clear", self)
-        self.levelBox = QComboBox(self)
+        self.frameLevelBoxLabel = QLabel("Select screen log's level:")
```
BECATRUE: Don't forget to set its parent to `self`!

iquip/apps/logger.py
```diff
@@ -125,52 +137,78 @@ class LoggerApp(qiwis.BaseApp):
...
+        fileLevelBox.textActivated.connect(functools.partial(self.setLevel, self.fileHandler))
+        fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
+
+    def initLogger(self, path: str):
+        """Initializes the root logger and handlers.
+        
+        Args:
+            path: The path of the log record file.    
+        """
+        self.levelsDict = {logging.DEBUG: "DEBUG", logging.INFO: "INFO", logging.WARNING: "WARNING",
```
BECATRUE: What happened to the discussion that each attribute must be defined in `__init__()`.

iquip/apps/logger.py
```diff
@@ -125,52 +137,78 @@ class LoggerApp(qiwis.BaseApp):
...
-        # TODO(aijuh): Change the log format when it is determined.
-        fs = "%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
-        formatter = logging.Formatter(fs)
-        self.handler.setFormatter(formatter)
+        self.initLogger(path)
+        # set loggerFrame's frameLevelBox
+        frameLevelBox = self.loggerFrame.frameLevelBox
+        frameLevelBox.addItems(self.levelsDict.values())
+        frameLevelBox.textActivated.connect(functools.partial(self.setLevel, self.frameHandler))
+        frameLevelBox.setCurrentText(self.levelsDict[self.frameHandler.level])
```
BECATRUE, line 170: How about setting the default logging level (here, `WARNING`) as a variable in the constructor?
Then you can pass the value to `initLogger` and `setCurrentText()`.
Also, in the future, we may set the default level in the `config.json`.

But, this is just a coding style, so if you prefer to the original method, please never mind.

BECATRUE (changes requested):
I think it will be enough if the file name format is decided.

iquip/apps/logger.py
Aijuh: I updated it right now!

Aijuh:
As @kangz12345 suggested, I used namer.

> Define namer(name) to return name.replace(".log", "") + ".log"



Also I tested saved file format as below by setting roll over interval as minute.

Since I used suffix of rotateHandler, when roll over occured, suffixed time is not the current time, but it is previous time.
For example when I press log button at `A`, and after two minutes I press log button at `B`, `A` is suffixed at the end of file.
I think we can distinguish which file is currently using and which one is the latest roll overed in this case. But There might be some problem that suffixed time is not matched to current time.

So I am considering not using suffix and add current time using `QdateTime` inside namer. Is it okay to do that?

image

Below is the GUI of logger viewer.


GUI

iquip/apps/logger.py
BECATRUE: Oh, I realize it now, but how about `logFilePath`?
I think `path` is little vague, because there is already `path` argument in `setup.json`.

iquip/apps/logger.py
```diff
@@ -125,52 +137,83 @@ class LoggerApp(qiwis.BaseApp):
...
-        fs = "%(levelname)s [%(name)s] [%(filename)s:%(lineno)d] %(message)s"
-        formatter = logging.Formatter(fs)
-        self.handler.setFormatter(formatter)
+        self.levelsDict = {logging.DEBUG: "DEBUG", logging.INFO: "INFO", logging.WARNING: "WARNING",
+                           logging.ERROR: "ERROR", logging.CRITICAL: "CRITICAL"}
+        self.frameHandler = LoggingHandler(self.addLog)
+        logFileName = os.path.join(path,
+                                   QDateTime.currentDateTime().toString("log_yyMMdd-HHmmss.log"))
+        self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName, when="midnight"
+                                                             ,encoding="utf-8")
```
BECATRUE: Please write it down as follows:
```python
self.fileHandler = handlers.TimedRotatingFileHandler(filename=logFileName, when="midnight",
                                                     encoding="utf-8")
```

BECATRUE (changes requested):

Aijuh:
@BECATRUE 
I understood that `path` in setup.json is for LoggerApp's path.
If I change name error occured. I think I can change to `logFilePath` in both cases.

kangz12345:
> As @kangz12345 suggested, I used namer.
> 
> > Define namer(name) to return name.replace(".log", "") + ".log"
> 
> Also I tested saved file format as below by setting roll over interval as minute.
> 
> Since I used suffix of rotateHandler, when roll over occured, suffixed time is not the current time, but it is previous time. For example when I press log button at `A`, and after two minutes I press log button at `B`, `A` is suffixed at the end of file. I think we can distinguish which file is currently using and which one is the latest roll overed in this case. But There might be some problem that suffixed time is not matched to current time.
> 
> So I am considering not using suffix and add current time using `QdateTime` inside namer. Is it okay to do that?
> 
> image
> 
> Below is the GUI of logger viewer.
> 
> GUI

I don't understand. Did you use `suffix` for the image example or not?

Please check if my expectation is correct:
1. When I started the program at 230810-235950, then a log file "230810-235950.log" is created.
2. At midnight, a new file is created: "230810-235950(230810).log".
3. At the next midnight, i.e. 24 hours later, a new file is created: "230810-235950(230811).log"
4. The first file "230810-235950.log" exists always, but whenever the rollover occurs, it is emptied.

Aijuh:
1. I used suffix at the example

Your expectation is correct.
At roll over, at file name `A`, time is added using suffix. And that time is when the `A` file is created from previous roll over.

I first thought that getting previous roll over time is not desirable, but as it happen only once in a day, it would be okay to use suffix. Just ignore my previous suggestion.

kangz12345:
Thank you for the confirmation!

iquip/apps/logger.py
```diff
@@ -125,52 +137,83 @@ class LoggerApp(qiwis.BaseApp):
...
+        fileHandler: A handler for saving logs to file.
     """
 
-    def __init__(self, name: str, parent: Optional[QObject] = None):
-        """Extended."""
+    def __init__(self, name: str, logFilePath: str, parent: Optional[QObject] = None):
+        """Extended.
+        
+        Args:
+            path: The path of the log record file.
```
kangz12345: Argument name is changed.

iquip/apps/logger.py
```diff
@@ -125,52 +137,83 @@ class LoggerApp(qiwis.BaseApp):
...
+        fileLevelBox = self.loggerFrame.fileLevelBox
+        fileLevelBox.addItems(self.levelsDict.values())
+        fileLevelBox.currentTextChanged.connect(functools.partial(self.setLevel, self.fileHandler))
+        fileLevelBox.setCurrentText(self.levelsDict[self.fileHandler.level])
+
+    def initLogger(self):
+        """Initializes the root logger and handlers.
+        
+        Args:
+            path: The path of the log record file.
```
kangz12345: Argument is removed.

kangz12345 (changes requested):

kangz12345 (approved):
LGTM

kangz12345:
I think you should take both changes (from ours and theirs) for conflict resolution.

Aijuh:
Do you mean that adding scheduler part in setup.json?

kangz12345:
> Do you mean that adding scheduler part in setup.json?

Yes, keep _both_ your changes from this PR and from `develop`.
@BECATRUE for cross check, please confirm my thought.

BECATRUE (approved):
LGTM!

Issue #72, opened 2023-07-19
Construct connection between the scheduler app and the `artiq-proxy` server


giwon2004:
### Feature you want to implement
To display the queued experiments, the information of them has to be uploaded to the server, which is run by `artiq-proxy` now.
I will implement the fetch thread that gets the information from the server and sends it to the `SchedulerFrame`.

PR #89, opened 2023-08-02
Unit test for scheduler

### LINKED ISSUE #71: Implement unit test for the scheduler app
`ExperimentView` inherits `QListView` which needs `QAbstractListModel` to manage data in the list.
Also, `QListView` needs `QAbstractItemDelegate` to display customized `QWidget`'s child class.

Both of them are very abstract and vague as a significant amount of their code is hidden in their parent classes.
Thus to make sure they are working properly without reading codes one by one, I will implement a unit test for these layouts.

giwon2004:
I made a unit test for `scheduler.py`, testing:

- the class `ExperimentModel`
  - Indexing & sorting data
  - Drag & Drop feature
- the class `TestSchedulerApp`
  - Add experiment
  - Run experiment
  - Modify (change/delete) experiment

### Problem
- The drag & drop feature is indirectly tested, by testing the function that is called internally when drag & drop occurs.
- I did not find any necessity of using `mock` as the functions do not call other functions.

### Future Works
- After implementing #72, more unit tests will be waiting for features related to `artiq-proxy`.

### DIFF
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -75,7 +75,7 @@ class SchedulerAppTest(unittest.TestCase):
             for info in data:
                 app.addExperiment(info)
                 mocked_queue.append.assert_called_with(info)
-                assert mocked_queue.sort.called
+            self.assertEqual(mocked_queue.sort.call_count, len(data))
 
     def test_run_experiment(self):
         app = scheduler.SchedulerApp(name="name")
@@ -96,8 +96,8 @@ class SchedulerFunctionalTest(unittest.TestCase):
 
     def test_add_experiment(self):
         app = scheduler.SchedulerApp(name="name", parent=QObject())
-        priorities = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0] # "priority" value for each experiment
-        sorted_indices = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9] # experiment number when sorted
+        priorities = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0]
+        sorted_indices = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9]
         data = [
             ExperimentInfo(str(i), {"rid": i, "priority": priorities[i]}) for i in range(10)
         ]
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -19,16 +19,16 @@ class ExperimentModelTest(unittest.TestCase):
         del self.qapp
 
     def test_row_count(self):
-        data1 = (ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10))
-        data2 = (ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10))
+        data1 = (ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10),)
+        data2 = (ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10),)
         for data in (data1, data2):
             model = scheduler.ExperimentModel()
             model.experimentQueue.extend(data)
             self.assertEqual(model.rowCount(), len(data))
 
     def test_data(self):
-        data1 = (ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10))
-        data2 = (ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10))
+        data1 = (ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10),)
+        data2 = (ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10),)
         for data in (data1, data2):
             model = scheduler.ExperimentModel()
             model.experimentQueue.extend(data)
@@ -71,7 +71,7 @@ class SchedulerAppTest(unittest.TestCase):
     def test_add_experiment(self):
         app = scheduler.SchedulerApp(name="name")
         with mock.patch.object(app.schedulerFrame.model, "experimentQueue") as mocked_queue:
-            data = (ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10))
+            data = (ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10),)
             for info in data:
                 app.addExperiment(info)
                 mocked_queue.append.assert_called_with(info)
@@ -99,7 +99,7 @@ class SchedulerFunctionalTest(unittest.TestCase):
         priorities = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0]
         sorted_indices = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9]
         data = (
-            ExperimentInfo(str(i), {"rid": i, "priority": priorities[i]}) for i in range(10)
+            ExperimentInfo(str(i), {"rid": i, "priority": priorities[i]}) for i in range(10),
         )
         for info in data:
             app.addExperiment(info)
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -19,16 +19,16 @@ class ExperimentModelTest(unittest.TestCase):
         del self.qapp
 
     def test_row_count(self):
-        data1 = (ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10),)
-        data2 = (ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10),)
+        data1 = tuple(ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10))
+        data2 = tuple(ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10))
         for data in (data1, data2):
             model = scheduler.ExperimentModel()
             model.experimentQueue.extend(data)
             self.assertEqual(model.rowCount(), len(data))
 
     def test_data(self):
-        data1 = (ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10),)
-        data2 = (ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10),)
+        data1 = tuple(ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10))
+        data2 = tuple(ExperimentInfo(str(i), {"rid": i, "priority": 0}) for i in range(10))
         for data in (data1, data2):
             model = scheduler.ExperimentModel()
             model.experimentQueue.extend(data)
@@ -50,13 +50,13 @@ class ExperimentModelTest(unittest.TestCase):
         mime2 = QMimeData()
         mime2.setText("2")
         model.dropMimeData(mime0, Qt.MoveAction, 0, 0, model.index(0)) # exp1 above exp1
-        self.assertEqual(model.experimentQueue, data)
+        self.assertEqual(model.experimentQueue, list(data))
         model.dropMimeData(mime0, Qt.MoveAction, 2, 0, model.index(0)) # exp1 above exp3
-        self.assertEqual(model.experimentQueue, data)
+        self.assertEqual(model.experimentQueue, list(data))
         model.dropMimeData(mime1, Qt.MoveAction, 3, 0, model.index(0)) # exp2 below exp3
         self.assertEqual(model.experimentQueue, [data[0], data[2], data[1]])
         model.dropMimeData(mime2, Qt.MoveAction, 1, 0, model.index(0)) # exp2 above exp3
-        self.assertEqual(model.experimentQueue, data)
+        self.assertEqual(model.experimentQueue, list(data))
 
 
 class SchedulerAppTest(unittest.TestCase):
@@ -71,7 +71,7 @@ class SchedulerAppTest(unittest.TestCase):
     def test_add_experiment(self):
         app = scheduler.SchedulerApp(name="name")
         with mock.patch.object(app.schedulerFrame.model, "experimentQueue") as mocked_queue:
-            data = (ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10),)
+            data = tuple(ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10))
             for info in data:
                 app.addExperiment(info)
                 mocked_queue.append.assert_called_with(info)
@@ -98,8 +98,8 @@ class SchedulerFunctionalTest(unittest.TestCase):
         app = scheduler.SchedulerApp(name="name", parent=QObject())
         priorities = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0]
         sorted_indices = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9]
-        data = (
-            ExperimentInfo(str(i), {"rid": i, "priority": priorities[i]}) for i in range(10),
+        data = tuple(
+            ExperimentInfo(str(i), {"rid": i, "priority": priorities[i]}) for i in range(10)
         )
         for info in data:
             app.addExperiment(info)
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -1,4 +1,4 @@
-"""Unit tests for monitor module."""
+"""Unit tests for scheduler module."""
 
 import unittest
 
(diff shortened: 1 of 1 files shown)


BECATRUE:
The GitHub actions somtimes fail. Please re-run the failed actions, manually.

giwon2004:
Did it!

tests/test_scheduler.py
```diff
@@ -0,0 +1,110 @@
+"""Unit tests for monitor module."""
```
BECATRUE: I think you should fix it!

tests/test_scheduler.py
```diff
@@ -0,0 +1,110 @@
...
+
+import unittest
+
+from PyQt5.QtWidgets import QApplication
+from PyQt5.QtCore import Qt, QObject,  QMimeData
+
+from iquip.apps import scheduler
+from iquip.protocols import ExperimentInfo as ExpInfo
```
BECATRUE: Please use just `ExperimentInfo`.
"Exp" looks vague.

tests/test_scheduler.py
```diff
@@ -0,0 +1,110 @@
...
+        self.assertEqual(mdl.experimentQueue, data)
+        mdl.dropMimeData(mime0, Qt.MoveAction, 2, 0, mdl.index(0)) # exp1 above exp3
+        self.assertEqual(mdl.experimentQueue, data)
+        mdl.dropMimeData(mime1, Qt.MoveAction, 3, 0, mdl.index(0)) # exp2 below exp3
+        self.assertEqual(mdl.experimentQueue, [data[0], data[2], data[1]])
+        mdl.dropMimeData(mime2, Qt.MoveAction, 1, 0, mdl.index(0)) # exp2 above exp3
+        self.assertEqual(mdl.experimentQueue, data)
+
+
+class TestSchedulerApp(unittest.TestCase):
```
BECATRUE: For our convention, please rename it to `SchedulerAppTest`.

tests/test_scheduler.py
```diff
@@ -0,0 +1,110 @@
...
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_add_experiment(self):
+        app = scheduler.SchedulerApp(name="name", parent=QObject())
+        data = [ExpInfo("exp" + str(i), {"rid": i, "priority": 10-i}) for i in range(10)]
+        for info in data:
+            app.addExperiment(info=info)
```
BECATRUE: I think it is okay to write just `app.addExperiment(info)`.

tests/test_scheduler.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_add_experiment(self):
+        app = scheduler.SchedulerApp(name="name", parent=QObject())
+        data = [ExpInfo("exp" + str(i), {"rid": i, "priority": 10-i}) for i in range(10)]
```
BECATRUE: Please rename "exp" and add spacings before and after `-` operator, i.e., `10 - i`.

BECATRUE: I have a question.

Doesn't `ExperimentInfo` class contain a RID and a priority?
In `protocols.py`, it has only `name` and `arginfo`.

tests/test_scheduler.py
```diff
@@ -0,0 +1,110 @@
...
+        for info in data:
+            app.addExperiment(info=info)
+        self.assertEqual(app.schedulerFrame.model.experimentQueue,
+                         [data[inv_permutation[i]] for i in range(10)])
+
+    def test_run_experiment(self):
+        app = scheduler.SchedulerApp(name="name", parent=QObject())
+        experiment_run = ExpInfo("exp1", {"rid": 1, "priority": 1})
+        experiment_queue = ExpInfo("exp2", {"rid": 2, "priority": 2})
+        app.schedulerFrame.model.experimentQueue.append(experiment_run)
```
BECATRUE, line 113: I think you should mock the model here.

What you did is functional test rather than unit test.
We should implement functional tests, too, so please separate it, not removing.

BECATRUE (changes requested):

tests/test_scheduler.py
giwon2004: I thought `arginfo` contains RID and priority. Am I wrong?

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+            self.assertEqual(mdl.rowCount(), len(data))
+            for i, exp in enumerate(data):
+                self.assertEqual(mdl.data(mdl.index(i)), exp)
+
+    def test_drag_and_drop(self):
+        mdl = scheduler.ExperimentModel()
+        data = [ExperimentInfo("1", {"rid": 1, "priority": 2}),
+                ExperimentInfo("2", {"rid": 2, "priority": 1}),
+                ExperimentInfo("3", {"rid": 3, "priority": 1})
+               ]
```
kangz12345: More common convention:
```python
data = [
    A,
    B,
    C,
]

File: tests/test_scheduler.py [hunk 8]
```diff
@@ -0,0 +1,136 @@
...
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_model_index(self):
+        data1 = []
+        for i in range(10):
+            data1.append(ExperimentInfo(str(i), {"rid": i, "priority": i}))
```
kangz12345: How about using comprehension?
```python
data1 = [ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10)]

File: tests/test_scheduler.py [hunk 9]
```diff
@@ -0,0 +1,136 @@
...
+        data2 = []
+        for i in range(100):
+            data1.append(ExperimentInfo(str(i), {"rid": i, "priority": 0}))
+        for data in (data1, data2):
+            mdl = scheduler.ExperimentModel()
```
kangz12345: Please use the full word: `model`.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+            mdl.experimentQueue.extend(data)
+            self.assertEqual(mdl.rowCount(), len(data))
+            for i, exp in enumerate(data):
+                self.assertEqual(mdl.data(mdl.index(i)), exp)
```
kangz12345: Since this is a unit test, you should test each method independently.
In this test method, you are testing `rowCount()`, `index()`, and `data()`.
Even if each method is too simple, you have to implement an individual test for each of them.
Note that the simple codes could be updated in the future and the current tests will help remembering what we assumed and intended.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def test_drag_and_drop(self):
```
kangz12345: If you test `dropMimeData()` method, we name it `test_drop_mime_data()`.
This is a unit test, not a functional test.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_add_experiment(self):
+        app = scheduler.SchedulerApp(name="name")
+        with mock.patch.object(app.schedulerFrame.model, "experimentQueue") as mockedQueue:
```
kangz12345: Our naming convention for mocked objects is `mocked_queue`, since `mock` module uses that convention.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+class SchedulerAppTest(unittest.TestCase):
+    """Unit tests for SchedulerApp class."""
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_add_experiment(self):
```
kangz12345, line 71: Please check if the experiments are sorted as well.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+        app = scheduler.SchedulerApp(name="name")
+        with mock.patch.object(app.schedulerFrame.model, "experimentQueue") as mockedQueue:
+            data = [ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10)]
+            for info in data:
+                app.addExperiment(info)
+                mockedQueue.append.assert_called_with(info)
+
+    def test_run_experiment(self):
```
kangz12345, line 80: Please test the case when the added experiment was in the queue, as another test method, e.g., `test_run_experiment_delete_from_queue()`.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+        self.assertEqual(app.schedulerFrame.model.experimentQueue, data)
+        app.schedulerFrame.model.experimentQueue = []
+        permutation = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0]
+        inv_permutation = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9]
+        data = [ExperimentInfo(str(i), {"rid": i, "priority": permutation[i]}
+                       ) for i in range(10)]
+        for info in data:
+            app.addExperiment(info)
+        self.assertEqual(app.schedulerFrame.model.experimentQueue,
+                         [data[inv_permutation[i]] for i in range(10)])
```
kangz12345: If the previous context is not required, please separate this as another test method.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+        app = scheduler.SchedulerApp(name="name", parent=QObject())
+        data = [ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10)]
+        for info in data:
+            app.addExperiment(info)
+        self.assertEqual(app.schedulerFrame.model.experimentQueue, data)
+        app.schedulerFrame.model.experimentQueue = []
+        permutation = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0]
+        inv_permutation = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9]
+        data = [ExperimentInfo(str(i), {"rid": i, "priority": permutation[i]}
+                       ) for i in range(10)]
```
kangz12345:
```python
data = [
    ExperimentInfo(...) for i in range(10)
]
```
would be better.

tests/test_scheduler.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def test_add_experiment(self):
+        app = scheduler.SchedulerApp(name="name", parent=QObject())
+        data = [ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10)]
+        for info in data:
+            app.addExperiment(info)
+        self.assertEqual(app.schedulerFrame.model.experimentQueue, data)
+        app.schedulerFrame.model.experimentQueue = []
+        permutation = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0]
+        inv_permutation = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9]
```
kangz12345: Could you explain more about `permutation` and `inv_permutation`?

kangz12345 (changes requested):

tests/test_scheduler.py
giwon2004: I found out that there is no situation that the added experiment was in the queue, as it receives the entire queue every time. There is no signal emitted when the experiment is run. Do I need to make this test even if it will be deleted afterward?

tests/test_scheduler.py
giwon2004: I could not understand what you mean here.

tests/test_scheduler.py
giwon2004: `permutation` is the priority mix-up, and the `inv_permutation` is the answer for this mix-up.

I will mention this in the code.

tests/test_scheduler.py
kangz12345: Do you mean that you are going to remove the code (in `scheduler.py`) which checks if the experiment is already in the queue?

tests/test_scheduler.py
kangz12345: You emptied the experiment queue and refilled it.
Is it necessary or not? (Is there a shared context between two tests, separated by line 98?) If they are independent tests, please separate it as two methods.

tests/test_scheduler.py
kangz12345: Thank you, now I understood.
How about using more intuitive names?
I understand what you intended, but it is unnecessarily mathematic (which requires more backgrounds for code readers).
For example, `priorities` and `sorted_indices`.

tests/test_scheduler.py
giwon2004: For the first part, it only checks appending without need to sort, and the second part checks the sorting (but also appending is included; inevitable for here.)

I guess the second part tests everything we need so I will just delete the first part.

tests/test_scheduler.py
giwon2004: Yes.

tests/test_scheduler.py
```diff
@@ -96,20 +96,15 @@ def tearDown(self):
...
         app = scheduler.SchedulerApp(name="name", parent=QObject())
-        data = [ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10)]
-        for info in data:
-            app.addExperiment(info)
-        self.assertEqual(app.schedulerFrame.model.experimentQueue, data)
-        app.schedulerFrame.model.experimentQueue = []
-        permutation = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0] # "priority" value for each experiment
-        inv_permutation = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9] # experiment number when sorted
+        priorities = [1, 9, 3, 8, 7, 4, 2, 6, 5, 0] # "priority" value for each experiment
+        sorted_indices = [1, 3, 4, 7, 8, 5, 2, 6, 0, 9] # experiment number when sorted
```
kangz12345: You can remove the comments now because the variable names are descriptive enough.

BTW we put 2 spaces before the comment "#".
See also: [PEP8](https://peps.python.org/pep-0008/#inline-comments)

tests/test_scheduler.py
```diff
@@ -66,18 +70,19 @@ def tearDown(self):
...
     def test_add_experiment(self):
         app = scheduler.SchedulerApp(name="name")
-        with mock.patch.object(app.schedulerFrame.model, "experimentQueue") as mockedQueue:
+        with mock.patch.object(app.schedulerFrame.model, "experimentQueue") as mocked_queue:
             data = [ExperimentInfo(str(i), {"rid": i, "priority": 10 - i}) for i in range(10)]
             for info in data:
                 app.addExperiment(info)
-                mockedQueue.append.assert_called_with(info)
+                mocked_queue.append.assert_called_with(info)
+                assert mocked_queue.sort.called
```
kangz12345: Please use `mocked_queue.sort.assert_called_once()` or something like that, instead of `assert`.

kangz12345 (changes requested):

tests/test_scheduler.py
BECATRUE: Oh, sorry for the late review..

When I defined `ExperimentInfo`, I thought this is a container of information about an experiment file, not a submitted experiment, so it has only the experiment and the build arguments.

But you used this in `scheduler.py` for describing a submitted experiment, so I think we should determine how to use this class.

BECATRUE: In my opinion, it looks better to define `SubmittedExperimentInfo` class separately in `protocols.py` and use this. @giwon2004

tests/test_scheduler.py
```diff
@@ -0,0 +1,137 @@
...
+    """Unit tests for ExperimentModel class."""
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_row_count(self):
+        data1 = [ExperimentInfo(str(i), {"rid": i, "priority": i}) for i in range(10)]
```
BECATRUE: Can you replace a list into a tuple?
For test, an immutable type looks safer.

BECATRUE (changes requested):

tests/test_scheduler.py
giwon2004: Okay, I will define a new type and make a new pull request for it.

kangz12345: @BECATRUE Then should we merge this PR after the new PR (`SubmittedExperimentInfo`), or can we merge this now?

kangz12345 (approved):

BECATRUE:
Is it okay that the coverage rate is 76%, not 100%? @giwon2004

giwon2004:
> Is it okay that the coverage rate is 76%, not 100%? @giwon2004

I guess, because there are functions that are called by the internal system that I cannot access.
Plus, I will fill up the blanks (if any) later by making other unit tests for other PRs for the `scheduler.py`.

BECATRUE:
Then, LGTM!

BECATRUE (approved):

PR #91, opened 2023-08-02
Fetch from the proxy server

### LINKED ISSUE #72: Construct connection between the scheduler app and the `artiq-proxy` server

giwon2004:
## What I made
image
The scheduler now receives a signal from `artiq-proxy` to display the queued experiment in real-time. This GET signal is requested every millisecond.

Due to the fast refreshing, it sometimes give index errors as it goes through the parallel procedure of displaying and deleting items in the list. So some of the parts were modified to handle index errors.

### Problems
 - Occasionally, unknown message blocks show up and close rapidly (I guess this is related to the frequency of fetching) when a new experiment goes to the 'running experiments' section.

The below two are explained in detail in [Disscussion#90](https://github.com/snu-quiqcl/iquip/discussions/90).
 - Drag & Drop feature has to be deleted due to the fast refreshing.
 - The proxy server dies as the number of GET requests piles up.
 
 This closes #72.


BECATRUE:
Before reviewing this PR, let's review and merge snu-quiqcl/artiq-proxy#41.

kangz12345:
Hmm, I think you started implementation before we agree on how to fetch the status.
Anyway, as @BECATRUE said, please reopen this PR after the artiq-proxy is updated.

giwon2004:
This will be done after #104 is merged.

giwon2004:
This is reopened as #104 is merged.

BECATRUE:
First of all, please resolve some conflicts! @giwon2004

giwon2004:
I will resolve conflicts after #113 is merged. If not, the merge commit will give unit test errors and pylint warnings.

BECATRUE:
Why was `SubmittedExperimentInfo` implemeneted again?
Was it added in #104?

giwon2004:
> Why was `SubmittedExperimentInfo` implemeneted again?

It is a manual rebase to use the modified class. It will not change the original `develop` branch for that.

BECATRUE:
> > Why was `SubmittedExperimentInfo` implemeneted again?
> 
> It is a manual rebase to use the modified class. It will not change the original `develop` branch for that.

Then, isn't it right not to say it's modified?

I think you should follow the steps below.
1. Rebase the `develop` branch.
2. Push this branch to the remote one forcely.

giwon2004:
> I will resolve conflicts after #113 is merged. If not, the merge commit will give unit test errors and pylint warnings.

Same reason here.

iquip/apps/scheduler.py
```diff
@@ -219,10 +217,62 @@ def paint(self,
...
+        and emits a list and an ExperimentInfo instance for display.
+        """
+        while True:
+            try:
+                response = requests.get("http://127.0.0.1:8000/experiment/queue/", timeout=10)
+                response.raise_for_status()
+                response = response.json()
+            except requests.exceptions.RequestException as err:
+                print(err)
+                return
```
kangz12345: Since you gave `timeout`, it could raise a timeout error.
Does the proxy server replies only when the queue status is changed?
If so, the timeout error may occur even if it is not an erroneous case.
Then it should `continue` instead of `return`.
Please confirm if I understand the proxy server's behavior correctly.

giwon2004: Oh, you are right. I will change it that way.

iquip/apps/scheduler.py
```diff
@@ -258,6 +258,8 @@ def run(self):
                 response.raise_for_status()
                 response = response.json()
             except requests.exceptions.RequestException as err:
+                if isinstance(err, requests.exceptions.Timeout):
+                    continue
```
kangz12345: How about adding `except requests.exceptions.Timeout` instead?
For example,
```python
try:
    ...
except requests.exceptions.Timeout:
    continue
except requests.exceptions.RequestException as err:
    ...
```

kangz12345 (changes requested):

iquip/apps/scheduler.py
```diff
@@ -63,24 +67,20 @@ class RunningExperimentView(QWidget):
...
     def __init__(self, parent: Optional[QWidget] = None):
         """Extended."""
         super().__init__(parent=parent)
         self.experimentInfo = None
-        # widgets
-        self.nameLabel = QLabel("None", self)
         # layout
         layout = QVBoxLayout(self)
         self.argsLayout = QHBoxLayout()
-        layout.addWidget(self.nameLabel)
```
BECATRUE: Now, you don't use `nameLabel`, so there is only on element, `self.argsLayout` in `layout`.
Then, is it able to merge two layouts, as follows?

```python
self.layout = QHBoxLayout(self)
# replace argLayout into layout 
```

iquip/apps/scheduler.py
```diff
@@ -234,9 +286,31 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         """Extended."""
         super().__init__(name, parent=parent)
         self.schedulerFrame = SchedulerFrame()
+        self.thread = _ExperimentQueueFetcherThread(
```
BECATRUE, line 365: Please write the docstring about `self.thread`.

iquip/apps/scheduler.py
```diff
@@ -209,6 +205,8 @@ def paint(self,
         index: QModelIndex):
         """Overridden."""
         data = index.data(Qt.DisplayRole)
+        if data is None:
```
BECATRUE, line 249: When is the `data` None? Just for safety?

iquip/apps/scheduler.py
```diff
@@ -219,10 +217,64 @@ def paint(self,
...
+        """
+        while True:
+            try:
+                response = requests.get("http://127.0.0.1:8000/experiment/queue/", timeout=10)
+                response.raise_for_status()
+                response = response.json()
+            except requests.exceptions.Timeout:
+                continue
+            except requests.exceptions.RequestException as err:
+                print(err)
```
BECATRUE: Now the `LoggerApp` is avaiable, please print through it.

iquip/apps/scheduler.py
```diff
@@ -219,10 +217,64 @@ def paint(self,
...
+                return
+            runningExperiment = None
+            experimentList = []
+            for key, value in response.items():
+                experimentInfo = SubmittedExperimentInfo(rid=int(key))
+                for item in tuple(item for item, _ in experimentInfo.items()):
+                    setattr(experimentInfo, item, value[item])
+                if value["status"] in ["running", "run_done", "analyzing", "deleting"]:
```
BECATRUE: How can you find these four statuses? Is there any reference?

BECATRUE: Also, I think one experiment may be being analyzed while the other is running.

iquip/apps/scheduler.py
```diff
@@ -219,10 +217,64 @@ def paint(self,
...
+            except requests.exceptions.Timeout:
+                continue
+            except requests.exceptions.RequestException as err:
+                print(err)
+                return
+            runningExperiment = None
+            experimentList = []
+            for key, value in response.items():
+                experimentInfo = SubmittedExperimentInfo(rid=int(key))
+                for item in tuple(item for item, _ in experimentInfo.items()):
```
BECATRUE: It looks little awkward. How about defining a method like `keys()` in `SubmittedExperimentInfo`?

iquip/apps/scheduler.py
```diff
@@ -234,9 +286,31 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+
+        Args:
+            experimentList: The queue of pending experiments.
+            runningExperiment: The experiment running now. None if there is no experiements running.
+        """
+        self.schedulerFrame.model.experimentQueue = experimentList
+        self.schedulerFrame.model.dataChanged.emit(self.schedulerFrame.model.index(0), # refresh
+                                                   self.schedulerFrame.model.index(0),
+                                                   [Qt.EditRole])
+        self.schedulerFrame.model.experimentQueue.sort(key=lambda x: x.arginfo["priority"],
```
BECATRUE: The `sort()` seems to be called frequently with the same option. Is there a way to simplify it?

iquip/apps/scheduler.py
```diff
@@ -234,9 +286,31 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+                             runningExperiment: Optional[SubmittedExperimentInfo] = None,
+        ):
+        """Displays the experiments fetched from the uploaded queue in the proxy server.
+
+        Args:
+            experimentList: The queue of pending experiments.
+            runningExperiment: The experiment running now. None if there is no experiements running.
+        """
+        self.schedulerFrame.model.experimentQueue = experimentList
+        self.schedulerFrame.model.dataChanged.emit(self.schedulerFrame.model.index(0), # refresh
```
BECATRUE: Please add one more space before the comment.

BECATRUE (changes requested):

iquip/apps/scheduler.py
giwon2004: It is already done on `develop` (when `SubmittedExperimentInfo` is made), but this branch is too old that it does not have that applied. I will resolve this _conflict_ after PR #113 is done.

iquip/apps/scheduler.py
giwon2004: you can find the statuses here: [`artiq/artiq/master/scheduler.py`](https://github.com/m-labs/artiq/blob/01d777c977083c9166570eb26bc43112cecfde0a/artiq/master/scheduler.py#L18)

giwon2004: Then I will change it to detect only two states, `running`, `run_done`. I can add them later in another list view mentioned in #106.

iquip/apps/scheduler.py
giwon2004: I added this because when `_syncExperimentQueue` is called and the input value is `experimentList = []`, the `self.schedulerFrame.model`'s `dataChanged` call gives an error, as there is nothing to refresh. Here, adding the case when `data` is `None` enables it to think there is something (which is nothing as you can see: `QSize(0, 0)`) to refresh.

iquip/apps/scheduler.py
giwon2004: I will make a new issue & PR for this later.

iquip/apps/scheduler.py
BECATRUE: @giwon2004 Please check it!

iquip/apps/scheduler.py
BECATRUE: Okay! Then, please refer to this comment when creating the issue.

BECATRUE (changes requested):
Maybe final!

iquip/apps/scheduler.py
giwon2004: Already made it in #116.

iquip/protocols.py
```diff
@@ -15,3 +15,28 @@ class ExperimentInfo:
...
+        pipeline: The pipeline of the experiment.
+        expid: The overall information of the experiment, 
+          which is a dictionary that may include arguments, file, etc.
+        due_date: The due date of the experiment.
+    """
+    rid: int
+    priority: int = 0
+    status: str = ""
+    pipeline: str = ""
+    expid: Dict[str, Any] = dataclasses.field(default_factory = dict)
```
BECATRUE: I'm sorry I checked this late.

I think the spaces in `dataclasses.field()` should be removed.

BECATRUE (changes requested):
Really final😅

iquip/protocols.py
giwon2004: This will be done in #118.

giwon2004:
I do not understand, but merge commit makes a unit test error.

BECATRUE:
> I do not understand, but merge commit makes a unit test error.

It is because `app.thread` in the unittest is destroyed even if the thread is running yet.
To resolve it, you should pause the thread in each unittest.

I tried several methods using `thread.quit()` or `thread.terminate()`, but it doesn't work well because `run()` is an infinite loop.

So I suggest to make a flag, e.g., `is_running` and before pausing the thread, set this flag to `False`.
Please refer to [this](https://stackoverflow.com/questions/68163578/stopping-an-infinite-loop-in-a-worker-thread-in-pyqt5-the-simplest-way).

giwon2004:
> So I suggest to make a flag, e.g., `is_running` and before pausing the thread, set this flag to `False`. Please refer to [this](https://stackoverflow.com/questions/68163578/stopping-an-infinite-loop-in-a-worker-thread-in-pyqt5-the-simplest-way).

Can you make it clear? Do I have to make a new flag `is_running` outside the function just for debugging?
And what solution do I have to check?

giwon2004:
I guess `timeout` of the `request` is making the error here because it lasts for at most 10 seconds more after receiving `thread.ctrl["break"] = True`.

iquip/apps/scheduler.py
```diff
@@ -244,10 +258,68 @@ def paint(self,
...
+        parent: Optional[QObject] = None
+    ):
+        """Extended.
+
+        Args:
+            callback: The callback method called after this thread is finished.
+        """
+        super().__init__(parent=parent)
+        self.fetched.connect(callback, type=Qt.QueuedConnection)
+        self.ctrl = {"break": False}
```
BECATRUE: It is just a question!
How come you define `ctrl` as a dictionary, not a boolean variable?
Is it for scalability?

BECATRUE: Also, the unittest still failed, right?

giwon2004: > How come you define `ctrl` as a dictionary, not a boolean variable?

It is from the site you linked, and I guess it has to be mutable to make the value affect each other.

giwon2004: > Also, the unittest still failed, right?

Yes...

BECATRUE: > It is from the site you linked, and I guess it has to be mutable to make the value affect each other.

Umm.. I think setting `ctrl` as a boolean value is okay and could you check this?

BECATRUE: > Yes...

How about waiting the thread at the end of each unittest, as follows?
```python
app.thread.ctrl["break"] = True  # app.thread.isRunning = True
app.thread.wait()
```

giwon2004:
Hmm, unit tests still fail.

BECATRUE:
The unit tests still fail😥 If it looks difficult to resolve it, you can create another issue with this problem, and delegate it.
Which one is better?

giwon2004:
> Which one is better?

I will delete the `ctrl` part and merge this one, and then we can resolve the unit test error in different issue / PR.

BECATRUE (approved):
Good!

Then, please create an issue about the unittest bug.

BECATRUE:
Please check this PR! @kangz12345

kangz12345 (approved):

PR #92, opened 2023-08-03
Unit test for logger App

### LINKED ISSUE #83: Implement unit test for logger app
Implement unit test for the logger.py.

### Plan for implementing feature.
1. Make test functions for class methods in LoggerApp. (If it don't change any variable, I will check whether specific function is called.) 

2. Make sure all the signals are connected to the proper function. (I am planning to call signal.emit and whether connected class method is called.)

### Additional context
If I missed something please let me know.

Aijuh:
This closes #83 

### Implemented basic unittests for LoggerApp.
I tested all the method in LoggerApp in module logger.py.
Since most of them are connected to specific signal, I tested whether LoggerApp method responds to connected signal well.
I followed previous codes and reference in our Slack chat.

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -40,12 +40,12 @@ class LoggingHandlerTest(unittest.TestCase):
 
     def test_emit(self):
         with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
-                app = logger.LoggerApp(name="name", parent=QObject())
-                temp_logger= logging.getLogger()
-                mocked_signal.reset_mock()
-                temp_logger.error("halo")
-                mocked_signal.emit.assert_called()
-                
+            app = logger.LoggerApp(name="name", parent=QObject())
+            temp_logger= logging.getLogger()
+            mocked_signal.reset_mock()
+            temp_logger.error("hello")
+            mocked_signal.emit.assert_called()
+    
 
 class ConfirmClearingFrameTest(unittest.TestCase):
     """Unit tests for the LoggerFrame class."""
@@ -77,7 +77,7 @@ class LoggerAppTest(unittest.TestCase):
 
     def tearDown(self):
         del self.qapp
-        
+
     def test_set_level(self):
         levels = {
             "DEBUG": logging.DEBUG,
@@ -122,11 +122,11 @@ class LoggerAppTest(unittest.TestCase):
 
     @mock.patch("iquip.apps.logger.QTextEdit.clear")
     def test_clear_log(self, mocked_clear):
-            with mock.patch("iquip.apps.logger.logger") as mocked_logger:
-                app = logger.LoggerApp(name="name", parent=QObject())
-                app.clearLog()
-                mocked_clear.assert_called_once()
-                mocked_logger.info.called_once_with("Tried to clear logs by clicking clear button")
+        with mock.patch("iquip.apps.logger.logger") as mocked_logger:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            app.clearLog()
+            mocked_clear.assert_called_once()
+            mocked_logger.info.called_once_with("Tried to clear logs by clicking clear button")
 
     def test_add_log(self):
         with mock.patch("iquip.apps.logger.QTextEdit.insertPlainText") as mocked_insert:
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -1,0 +1,68 @@
+"""Unit tests for logger module."""
+
+import unittest
+from unittest import mock
+import logging
+
+from PyQt5.QtCore import QObject
+from PyQt5.QtWidgets import QApplication
+
+from iquip.apps import logger
+
+class TestloggerApp(unittest.TestCase):
+    """Unit tests for loggerApp class."""
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_handler_connection(self):
+        """Tests that the handler's signal is successfully connected to LoggerApp method addLog."""
+        with mock.patch('iquip.apps.logger._Signaller.signal') as mocked_signal:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            mocked_signal.connect.assert_called_once_with(app.addLog)
+
+    def test_set_level(self):
+        levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
+        app = logger.LoggerApp('name')
+        for level in levels_dict.values():
+            app.setLevel(levelText = level)
+            self.assertEqual(levels_dict[app.handler.level], level)
+
+    def test_frames(self):
+        app = logger.LoggerApp(name="name", parent=QObject())
+        self.assertEqual(app.frames(), (app.loggerFrame,))
+
+    def test_call_check_to_clear(self):
+        with mock.patch('iquip.apps.logger.LoggerApp.checkToClear') as mocked_method:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            app.loggerFrame.clearButton.clicked.emit()
+            mocked_method.assert_called_once()
+
+    def test_call_clear_log(self):
+        with mock.patch('iquip.apps.logger.LoggerApp.clearLog') as mocked_method:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            app.confirmFrame.buttonBox.accepted.emit()
+            mocked_method.assert_called_once()
+
+    def test_call_add_log(self):
+        with mock.patch('iquip.apps.logger.LoggerApp.addLog') as mocked_method:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            log = logger.logger
+            app.frames()
+            messages = {"DEBUG": "Hello", "INFO": "this", "WARNING": "code",
+                        "ERROR": "is", "CRITICAL": "unittest"}
+            for level, msg in messages.items():
+                app.setLevel(level)
+                log.warning(msg)
+                if logging.WARNING >= getattr(logging, level):
+                    mocked_method.assert_called_once()
+                    mocked_method.reset_mock()
+                else:
+                    mocked_method.assert_not_called()
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -22,9 +22,8 @@ class LoggingHandlerTest(unittest.TestCase):
         with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
             app = logger.LoggerApp(name="name", parent=QObject())
             app.setLevel("DEBUG")
-            temp_logger= logging.getLogger(__name__)
-            mocked_signal.reset_mock()
-            temp_logger.info("hello")
+            test_log = logging.makeLogRecord({"name": "hello"})
+            app.handler.emit(test_log)
             mocked_signal.emit.assert_called_once()
 
 
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -9,21 +9,6 @@ from PyQt5.QtWidgets import QApplication
 
 from iquip.apps import logger
 
-class SignallerTest(unittest.TestCase):
-    """Unit tests for _Signaller class."""
-
-    def setUp(self):
-        self.qapp = QApplication([])
-
-    def tearDown(self):
-        del self.qapp
-
-    def test_signal_connection(self):
-        with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
-            app = logger.LoggerApp(name="name", parent=QObject())
-            mocked_signal.connect.assert_called_once_with(app.addLog)
-
-
 class LoggingHandlerTest(unittest.TestCase):
     """Unit tests for LoggingHandler class."""
 
@@ -32,11 +17,6 @@ class LoggingHandlerTest(unittest.TestCase):
 
     def tearDown(self):
         del self.qapp
-
-    def test_handler_connection(self):
-        with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
-            app = logger.LoggerApp(name="name", parent=QObject())
-            mocked_signal.connect.assert_called_once_with(app.addLog)
 
     def test_emit(self):
         with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
@@ -48,7 +28,7 @@ class LoggingHandlerTest(unittest.TestCase):
     
 
 class ConfirmClearingFrameTest(unittest.TestCase):
-    """Unit tests for the LoggerFrame class."""
+    """Unit tests for LoggerFrame class."""
 
     def setUp(self):
         self.qapp = QApplication([])
@@ -70,7 +50,7 @@ class ConfirmClearingFrameTest(unittest.TestCase):
 
 
 class LoggerAppTest(unittest.TestCase):
-    """Unit tests for the LoggerApp class."""
+    """Unit tests for LoggerApp class."""
 
     def setUp(self):
         self.qapp = QApplication([])
@@ -103,7 +83,7 @@ class LoggerAppTest(unittest.TestCase):
         self.assertEqual(app.frames(), (app.loggerFrame,))
 
     def test_call_check_to_clear(self):
-        with mock.patch('iquip.apps.logger.LoggerApp.checkToClear') as mocked_method:
+        with mock.patch("iquip.apps.logger.LoggerApp.checkToClear") as mocked_method:
             app = logger.LoggerApp(name="name", parent=QObject())
             app.loggerFrame.clearButton.clicked.emit()
             mocked_method.assert_called_once()
@@ -115,7 +95,7 @@ class LoggerAppTest(unittest.TestCase):
             mocked_show.assert_called_once()
 
     def test_call_clear_log(self):
-        with mock.patch('iquip.apps.logger.LoggerApp.clearLog') as mocked_method:
+        with mock.patch("iquip.apps.logger.LoggerApp.clearLog") as mocked_method:
             app = logger.LoggerApp(name="name", parent=QObject())
             app.confirmFrame.buttonBox.accepted.emit()
             mocked_method.assert_called_once()
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -21,11 +21,12 @@ class LoggingHandlerTest(unittest.TestCase):
     def test_emit(self):
         with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
             app = logger.LoggerApp(name="name", parent=QObject())
-            temp_logger= logging.getLogger()
+            app.setLevel("DEBUG")
+            temp_logger= logging.getLogger(__name__)
             mocked_signal.reset_mock()
-            temp_logger.error("hello")
-            mocked_signal.emit.assert_called()
-    
+            temp_logger.info("hello")
+            mocked_signal.emit.assert_called_once()
+
 
 class ConfirmClearingFrameTest(unittest.TestCase):
     """Unit tests for LoggerFrame class."""
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -1,16 +1,31 @@
 """Unit tests for logger module."""
 
+import logging
 import unittest
 from unittest import mock
-import logging
 
 from PyQt5.QtCore import QObject
 from PyQt5.QtWidgets import QApplication
 
 from iquip.apps import logger
 
-class TestloggerApp(unittest.TestCase):
-    """Unit tests for loggerApp class."""
+class _SignallerTest(unittest.TestCase):
+    """Unit tests for the _Signaller class."""
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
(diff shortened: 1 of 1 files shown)


Aijuh:
Now, file saving feature of log viewer is in PR and going to be merged later. But this code does not regard this new branch. 
After merging file saving feature of log viewer, I will upgrade unittest in next PR later.

BECATRUE:
I think you implemented unit tests only for `LoggerApp`, not all in `looger.py`.
Is it right?

tests/test_logger.py
```diff
@@ -0,0 +1,94 @@
+"""Unit tests for logger module."""
+
+import unittest
+from unittest import mock
+import logging
```
BECATRUE: Please write this down before `from ... import ...` statements.

tests/test_logger.py
```diff
@@ -0,0 +1,94 @@
...
+
+from PyQt5.QtCore import QObject
+from PyQt5.QtWidgets import QApplication
+
+from iquip.apps import logger
+
+class TestloggerApp(unittest.TestCase):
```
BECATRUE: Please rename this to `LoggerAppTest` like `test_explorer.py`.

tests/test_logger.py
```diff
@@ -0,0 +1,94 @@
...
+    """Unit tests for loggerApp class."""
```
BECATRUE: LoggerApp!

tests/test_logger.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_handler_connection(self):
```
BECATRUE: This test is a unit test for `LoggingHandler`, not `LoggerApp`.
Please move this into another test class.

tests/test_logger.py
```diff
@@ -0,0 +1,94 @@
...
+        """Tests for the LoggerApp's method setLevel.
+        
+        For vaild input, tests whether level of rootLogger and handler changes succesfully.
+        For invaild input, tests whether level of rootLogger and handler remains same.
+        """
+        levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
+        app = logger.LoggerApp('name')
+        root_logger = logging.getLogger()
+        for level in levels_dict.values():
+            app.setLevel(levelText = level)
```
BECATRUE: Please write only `app.setLevel(level)`.

tests/test_logger.py
```diff
@@ -0,0 +1,94 @@
...
+    def test_set_level(self):
+        """Tests for the LoggerApp's method setLevel.
+        
+        For vaild input, tests whether level of rootLogger and handler changes succesfully.
+        For invaild input, tests whether level of rootLogger and handler remains same.
+        """
+        levels_dict = {10: "DEBUG", 20: "INFO", 30: "WARNING", 40: "ERROR", 50: "CRITICAL"}
+        app = logger.LoggerApp('name')
+        root_logger = logging.getLogger()
+        for level in levels_dict.values():
```
BECATRUE: I think using `levels` in `setLevel()` is more legible.

```python
levels = {
    "DEBUG": logging.DEBUG,
    "INFO": logging.INFO,
    "WARNING": logging.WARNING,
    "ERROR": logging.ERROR,
    "CRITICAL": logging.CRITICAL
}
...
for level_text in levels:
  app.setLevel(level_text)
```

tests/test_logger.py
```diff
@@ -0,0 +1,94 @@
...
+            app.setLevel(levelText = level)
+            self.assertEqual(levels_dict[app.handler.level], level)
+            self.assertEqual(levels_dict[root_logger.level], level)
+        prev_level = app.handler.level
+        non_levels = ["debug", "CODE", "ERR"]
```
BECATRUE: It looks enough to check only one non-level string😅

BECATRUE (changes requested):
In our unit tests, if the docstring looks unnecessary, you don't need to wrtie it.

Also, you should make the coverage rate 100%.

tests/test_logger.py
Aijuh: I will check details

Aijuh:
Yes what I currently implemented are tests for LoggerApp's method. I thought that other parts are quit simple and they don't need any test. But as you tell me to cover 100%, what I am going to implement is as follows.

1. I will make tests for confirmClearingFrame whether methods are connected well.
2. As for LoggerFrame, I can't figure out what should be tested, because it only has constructor.
3. Tests whether LoggingHandler's emit function works well when logging is used.

Aijuh:
@BECATRUE 
I update codes with respect to your feedback. I added all the tests for classes.

In my local, pylint error occured if I missed docstring. But In this Github server, pylint error did not occur even if I didn't write docstring. Thank you for the information.

If I should implement more, please inform me.

BECATRUE:
> In my local, pylint error occured if I missed docstring. But In this Github server, pylint error did not occur even if I didn't write docstring.

I'm having a similar problem. It works fine on the server right now, so let's ignore it.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+import logging
+import unittest
+from unittest import mock
+
+from PyQt5.QtCore import QObject
+from PyQt5.QtWidgets import QApplication
+
+from iquip.apps import logger
+
+class _SignallerTest(unittest.TestCase):
```
BECATRUE: It is a really minor comment, but please rename it to `SignallerTest`.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+    def test_button_cancel_clicked(self):
+        method_dir = 'iquip.apps.logger.ConfirmClearingFrame.buttonCancelClicked'
+        with mock.patch(method_dir) as mocked_method:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            app.confirmFrame.buttonBox.rejected.emit()
+            mocked_method.assert_called_once()
+
+
+class LoggerAppTest(unittest.TestCase):
+    """Unit tests for the LoggerApp class."""
```
BECATRUE: Remove "the".

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_set_level(self):
+        """Tests for the LoggerApp's method setLevel.
```
BECATRUE: I think it is okay to remove all docstring in here.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+        
+        For vaild input, tests whether level of rootLogger and handler changes succesfully.
+        For invaild input, tests whether level of rootLogger and handler remains same.
+        """
+        levels = {"DEBUG": logging.DEBUG, "INFO": logging.INFO, "WARNING": logging.WARNING,
```
BECATRUE: Please write each pair in only one line like `logger.py`.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+                  "ERROR": logging.ERROR, "CRITICAL": logging.CRITICAL}
+        app = logger.LoggerApp('name')
```
BECATRUE: Please use double quotes.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+            self.assertEqual(app.handler.level, value)
+            self.assertEqual(root_logger.level, value)
+        prev_level = app.handler.level
+        non_level = "non_level"
+        app.setLevel(non_level)
+        self.assertEqual(prev_level, app.handler.level)
+        self.assertEqual(prev_level, root_logger.level)
+
+    def test_frames(self):
+        """Tests for the LoggerApp's method frames."""
```
BECATRUE: Remove it!

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+        For vaild input, tests whether level of rootLogger and handler changes succesfully.
+        For invaild input, tests whether level of rootLogger and handler remains same.
+        """
+        levels = {"DEBUG": logging.DEBUG, "INFO": logging.INFO, "WARNING": logging.WARNING,
+                  "ERROR": logging.ERROR, "CRITICAL": logging.CRITICAL}
+        app = logger.LoggerApp('name')
+        root_logger = logging.getLogger()
+        for level, value in levels.items():
+            app.setLevel(level)
+            self.assertEqual(app.handler.level, value)
```
BECATRUE: I think you should mock `self.handler` and modify this test as follows:
```python
mocked_handler.level.assert_called_with(level)
```

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+        """Tests for the LoggerApp's method setLevel.
+        
+        For vaild input, tests whether level of rootLogger and handler changes succesfully.
+        For invaild input, tests whether level of rootLogger and handler remains same.
+        """
+        levels = {"DEBUG": logging.DEBUG, "INFO": logging.INFO, "WARNING": logging.WARNING,
+                  "ERROR": logging.ERROR, "CRITICAL": logging.CRITICAL}
+        app = logger.LoggerApp('name')
+        root_logger = logging.getLogger()
+        for level, value in levels.items():
```
BECATRUE: Please rename them to `levelText` and `level`, the same as those in `logger.py`.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+        """Tests for the LoggerApp's method clearLog.
+
+        Tests that LoggerApp's method clearLog is connected to confirmFrame's OK button.
+        """
+        with mock.patch('iquip.apps.logger.LoggerApp.clearLog') as mocked_method:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            app.confirmFrame.buttonBox.accepted.emit()
+            mocked_method.assert_called_once()
+
+    def test_call_add_log(self):
```
BECATRUE: This is also not a unit test, because this works well only if `LoggingHandler` is implemented correctly.

Generally, tests you implemented are called functional tests and we need these, too.
Referred to `test_explorer.py`, please separate unit tests and functional tests.

BECATRUE (changes requested):

BECATRUE:
Also, please check the coverage rate!

<image width="40%" src="https://github.com/snu-quiqcl/iquip/assets/76851886/92d63ef6-8647-4b3f-a1cc-ea593197f641">

You should reach 100% for `logger.py` without good reason.

Aijuh:
@BECATRUE 
I don't know how to check coverage rate of unittest. I will search for it.
If you can describe how to do it quickly, it would be great help, otherwise I will search it.

BECATRUE:
> @BECATRUE I don't know how to check coverage rate of unittest. I will search for it. If you can describe how to do it quickly, it would be great help, otherwise I will search it.

1. click the unit test result below. (Details button)
2. Click 'Run the unit tests and check coverage'.
3. You can check the coverage rate!

In addition, you can check it in you local as follows:
```shell
coverage report -m
```

kangz12345:
> > In my local, pylint error occured if I missed docstring. But In this Github server, pylint error did not occur even if I didn't write docstring.
> 
> I'm having a similar problem. It works fine on the server right now, so let's ignore it.

What is happening? I think we should find out why they behave differently.
Did you set the pylintrc file properly in your local?

Aijuh:
@BECATRUE 
I will check whether coverage rate is 100% before re-request.

@kangz12345 
Yes in my environment, there is pylint of version 2.17.4. Previously pylint result of my local is same as pylint result of the server. I don't know why there is difference of pylint at docstring in unittest.

BECATRUE:
I'm also haveing the similar problem to @Aijuh. I will find out causes about this.

tests/test_logger.py
Aijuh: I am sorry but I fail to mock handler.

1. I tried to mock as `mock.patch('iquip.apps.logger.LoggerApp.handler')`, but is says there is no attribute handler.
2. If I mock as `mock.patch('iquip.apps.logger.LoggingHandler')`, and call `mocked_handler.level.assert_called_with(level)` it says level is not called.
3. If I call `self.assertEqual(mocked_handler.level, level)`, mocked_handler.level is not a int value, it is magic mock type and test failed.

Should I move it to functional test?

kangz12345:
@BECATRUE Did you find out the cause?

BECATRUE:
> @BECATRUE Did you find out the cause?

Not, yet😅 I will find out it right now.

BECATRUE:
> @BECATRUE Did you find out the cause?

Oh, most of us use a conda environment, so we should change the Pylint command in our local as follows:
```shell
python -m pylint ...
```
Then, it will run the same as GitHub action.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+import unittest
+from unittest import mock
+
+from PyQt5.QtCore import QObject
+from PyQt5.QtWidgets import QApplication
+
+from iquip.apps import logger
+
+class _SignallerTest(unittest.TestCase):
+    """Unit tests for the _Signaller class."""
```
BECATRUE: Please remove "the".

I think this is just our convention.😅

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_signal_connection(self):
```
BECATRUE: I'm really sorry to move this test into `_Signaller` test..

I think this is so obvious that it is unnecessary.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+        with mock.patch('iquip.apps.logger._Signaller.signal') as mocked_signal:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            mocked_signal.connect.assert_called_once_with(app.addLog)
+
+
+class LoggingHandlerTest(unittest.TestCase):
+    """Unit tests for the LoggingHandler class."""
```
BECATRUE: "the"

tests/test_logger.py
BECATRUE: Please remove this test!

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+    def test_handler_connection(self):
+        with mock.patch('iquip.apps.logger._Signaller.signal') as mocked_signal:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            mocked_signal.connect.assert_called_once_with(app.addLog)
+
+    def test_emit(self):
+        with mock.patch('iquip.apps.logger.LoggingHandler.emit') as mocked_method:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            test_logger = logger.logger
+            app.setLevel("INFO")
```
BECATRUE: You CANNOT use `app.setLevel()` here because it is a unit test!
Thus, you should use `logging.getLogger().setLevel(...)` or `test_logger.warning(...)`. (higher that the default level)

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+    def tearDown(self):
+        del self.qapp
+
+    def test_handler_connection(self):
+        with mock.patch('iquip.apps.logger._Signaller.signal') as mocked_signal:
+            app = logger.LoggerApp(name="name", parent=QObject())
+            mocked_signal.connect.assert_called_once_with(app.addLog)
+
+    def test_emit(self):
+        with mock.patch('iquip.apps.logger.LoggingHandler.emit') as mocked_method:
```
BECATRUE: `mocked_method` is too vague.
Please rename it to `mocked_emit`.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+            app = logger.LoggerApp(name="name", parent=QObject())
+            test_logger = logger.logger
+            app.setLevel("INFO")
+            test_logger.info("hello")
+            mocked_method.assert_called_once()
```
BECATRUE: I think you should test the called parameters, too.

tests/test_logger.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class LoggerFrameTest(unittest.TestCase):
```
BECATRUE: So far, we didn't test the frame class because there is no function to test, and it is just a container of widgets.
Also the following two tests are not unit test, but functional tests for `ConfirmClearingFrame`.

I think you can remove them.

BECATRUE (changes requested):

tests/test_logger.py
Aijuh: I want to do that but it is little bit difficult because the input to the emit is log record, not a simple text.
It contains several information like time, logger generated by `logging`, I have no idea how to check parameter with `called_with`.

Aijuh:
I updated code for 100% coverage of Unittest. 

But In `LoggingHandlerTest`'s `test_emit` function, implementation is not completed. I tried to figure out but cannot find way at this point.
1. If I define LoggerApp as mocked object, signal.emit does not called.
2. If I define LoggerApp as real object, signal.emit is already called several times and `assert_called_once` fails.
3. I want to directly call LoggingHandler.emit but it takes log record as parameter, and I cannot find way to generate log record.

If we change code in `logger.py`'s `LoggingHandler` as take parameter of two types (`str`, `logging.LogRecord` ), maybe I can solve problem. But I didn't touch logger.py just in case.
If I missed something please let me know.

BECATRUE:
Please check the above remaining conversations!

Aijuh:
Can I understand your comment as fix unsolved comments first?

kangz12345 (approved):
I trust you @BECATRUE 😄

Aijuh:
I am still stuck with `emit` in LoggingHandlerTest. 

Unittest should check whether all codes worked well. So I should check whether `self.signaller.signal.emit(logMsg)` worked well in logger.py.
I want to check whether `signaller.siganl.emit` is called with given logrecord. But there is following problem.

1. If I directly define Logging Handler and call emit, I should give logging.LogRecord as input. But I don't have any idea of how to generate it.
2. If I defined LoggerApp and defined new logger and sent logs to root logger, `signaller.signal.emit` was not called. I tested by assuming `signal.emit` is in pyqtSignal or logging.Handler but in both case, emit is not called.

Of course I will test more, but if you have any insight about it, lease tell me.

kangz12345:
> I am still stuck with `emit` in LoggingHandlerTest.
> 
> Unittest should check whether all codes worked well. So I should check whether `self.signaller.signal.emit(logMsg)` worked well in logger.py. I want to check whether `signaller.siganl.emit` is called with given logrecord. But there is following problem.
> 
> 1. If I directly define Logging Handler and call emit, I should give logging.LogRecord as input. But I don't have any idea of how to generate it.
> 2. If I defined LoggerApp and defined new logger and sent logs to root logger, `signaller.signal.emit` was not called. I tested by assuming `signal.emit` is in pyqtSignal or logging.Handler but in both case, emit is not called.
> 
> Of course I will test more, but if you have any insight about it, lease tell me.

Could you include the codes? The lingual description is too ambiguous.

Aijuh:
```
def test_emit(self):
        with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
            app = logger.LoggerApp(name="name", parent=QObject())
            temp_logger= logging.getLogger(__name__)
            mocked_signal.reset_mock()
            temp_logger.error("hello")
            mocked_signal.emit.assert_called_once()
```
I find that signal.emit works in this code.

But in this code, when sending log, signal.emit works well. But since root logger has 9 handler in this entire test, signal.emit called 9 times. If I use mocked app or mocked signal to avoid it, didn't work well.

Aijuh:
```
       with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
            app = logger.LoggerApp(name="name", parent=QObject())
            app.setLevel("DEBUG")
            temp_logger= logging.getLogger(__name__)
            mocked_signal.reset_mock()
            temp_logger.info("hello")
            mocked_signal.emit.assert_called_once()
```
I changed `test_emit` as above. Since default level of handler is warning, if I send signal as `info`, only currently defined handler works. 

But I don't think it is not a good implementation. I try to solve problem using mocking, but I cannot find way at this point.
When I mock specific method, its internal code of mocked method does not executed and fail to send signal to root logger, and test_emit fails.

tests/test_logger.py
BECATRUE: You can use [`call_args`](https://docs.python.org/ko/3/library/unittest.mock.html#unittest.mock.Mock.call_args), or similar one. @Aijuh

BECATRUE: @Aijuh

BECATRUE:
> If I directly define Logging Handler and call emit, I should give logging.LogRecord as input. But I don't have any idea of how to generate it.

What point makes this difficult? I think you can generate `LogRecord` instance referred to [LogRecord](https://docs.python.org/3/library/logging.html#logrecord-objects).

BECATRUE:
How about the following code? I didn't test this so it may not work well..

```python
app = logger.LoggerApp(...)
with mock.patch.object(app.frameHandler.signaller.signal, "emit") as mocked_emit:
  ...
  temp_logger.info("hello")
  mocked_emit.assert_called_with(...)
```

Aijuh:
Thank you for giving specific reference for creating `logging.logrecord` class.
I missed the `logging.makeLogRecord` method is in the reference.
When I used it, the unittest works well.

BECATRUE (approved):
Perfect👍

PR #99, opened 2023-08-08
Handle an empty `choice`s of enumeration value

### LINKED ISSUE #55: Handle an empty `choices` of enumeration value
As we discussed in #47, an empty `choices` of enumeration value doesn't cause an error before submission, but it CANNOT be successful due to an empty value.
So, we should handle that case.

_Originally posted by @kangz12345 in https://github.com/snu-quiqcl/iquip/pull/47#discussion_r1252489188_

BECATRUE:
This PR will close #55.

To handle an empty `choice` of enumeration value, I implemented the following:
1. When an `_EnumerationEntry` has an empty `choice` and its `value()` is called, it raises a `ValueError`.
2. Once the submission button is cliced, the app obtains each value of all entries and if a `ValueError` occurs, the submission is rejected.

### Result
image

In this case, the `color` argument has an empty `choice` and if the submission button is clicked, several logs are printed.

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -141,7 +141,11 @@ class _EnumerationEntry(_BaseEntry):
         
         Returns the value of the comboBox.
         """
-        return self.comboBox.currentText()
+        if self.argInfo["choices"]:
+            return self.comboBox.currentText()
+        else:
+            
+            raise ValueError
 
 
 class _NumberEntry(_BaseEntry):
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -541,8 +541,12 @@ class BuilderApp(qiwis.BaseApp):
         
         Once the submitButton is clicked, this is called.
         """
-        experimentArgs = self.argumentsFromListWidget(self.builderFrame.argsListWidget)
-        schedOpts = self.argumentsFromListWidget(self.builderFrame.schedOptsListWidget)
+        try:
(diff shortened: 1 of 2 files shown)


giwon2004 (approved):
LGTM!

BECATRUE:
As @kangz12345 suggested offline, I updated the logging mechanism a little.

iquip/apps/builder.py
```diff
@@ -543,8 +541,8 @@ def submit(self):
         try:
             experimentArgs = self.argumentsFromListWidget(self.builderFrame.argsListWidget)
             schedOpts = self.argumentsFromListWidget(self.builderFrame.schedOptsListWidget)
-        except ValueError:
-            logger.error("The submission is rejected because of invalid arguments.")
+        except ValueError as e:
+            logger.exception("The submission is rejected because of invalid arguments: %s", e)
```
kangz12345: [`logging.exception()`](https://docs.python.org/3/library/logging.html#logging.exception) logs the exception information as well, so you don't have to explicitly write it down.
For example,
```python
except ValueError:
    logger.exception("...")
```

kangz12345 (changes requested):

iquip/apps/builder.py
BECATRUE: Oh, I see. I will fix it.

kangz12345 (approved):
LGTM!

PR #102, opened 2023-08-10
Add menu for queued experiments

### LINKED ISSUE #96: Displaying a menu widget when an experiment on the scheduler is right-clicked
I want to implement a menu widget that pops up from each experiment when the item is right-clicked.

From the displayed part of the `SchedulerFrame.queueView`, it will detect a click event to display `QMenu` widget that belongs to the selected `ExperimentView` or `RunningExperimentView` instance.

This `QMenu` widget will contain menus, labeled:
- Edit > Start an app to edit the "scannable"s (will be implemented in future)
- Delete > Kills the experiment
- Request termination > Requests graceful termination of the run

According to the functions supported by the `Scheduler` class in [`artiq/artiq/master/scheduler.py`](https://github.com/m-labs/artiq/blob/master/artiq/master/scheduler.py#L400), which will call functions (mentioned in [Issue#44](https://github.com/snu-quiqcl/artiq-proxy/issues/44) of `artiq-proxy`) in [`artiq_proxy/main.py`](https://github.com/snu-quiqcl/artiq-proxy/blob/develop/main.py) to request their work to scheduler in the proxy server.

giwon2004:
I made a `QMenu` object that displays a menu when right-clicked:
image
(The space in front of the text will be deleted in the future, using `QMenu.setStyleSheet`.)

Here, Edit will be future implemented, and the other two are connected to POST request mentioned in [artiq-proxy#44](https://github.com/snu-quiqcl/artiq-proxy/issues/44) and #96.

However, it is not fully tested as the experiments are refreshed so fast that it needs a clicking skill to terminate one.
(Maybe there is a simple experiment that holds for a long time to test this?)

Plus, it gave an error that `destroyed a QThread object while running` when I deleted an experiment. 
(I do not know whether this is intended as I did a deletion rather than a termination request.)

This closes #96.


iquip/apps/scheduler.py
```diff
@@ -234,6 +302,35 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+            index = self.schedulerFrame.model.index(i)
+            if self.schedulerFrame.queueView.rectForIndex(index).contains(event.pos()):
+                rid = self.schedulerFrame.model.data(index).arginfo["rid"]
+                menu = QMenu(self.schedulerFrame)
+                edit = menu.addAction("Edit")
+                delete = menu.addAction("Delete")
+                request_termination = menu.addAction("Request termination")
+                # TODO(giwon2004) Remove icon space from the menu list (use menu.setStyleSheet)
+
+                action = menu.exec_(self.schedulerFrame.mapToGlobal(QPoint(0,25)) + event.pos())
```
kangz12345: Could you explain more about calculating the position here for me?

iquip/apps/scheduler.py
```diff
@@ -223,6 +258,39 @@ def sizeHint(self, option: QStyleOptionViewItem, index: QModelIndex) -> QSize:
...
+        self.mode = mode
+        self.rid = rid
+
+    def run(self):
+        """Overridden."""
+        base_path = "http://127.0.0.1:8000/experiment/"
+        if self.mode == "delete":
+            requests.post(base_path + "delete", params={"rid": self.rid}, timeout = 10)
+        elif self.mode == "request_termination":
+            requests.post(base_path + "terminate", params={"rid": self.rid}, timeout = 10)
```
kangz12345: Is there a reason you separated `mode` and the path? For example, "request_termination" and "terminate".

iquip/apps/scheduler.py
```diff
@@ -30,6 +34,40 @@ def _dismiss_items(layout: Optional[QLayout] = None):
...
+    thread.start()
+
+
+class ExperimentListView(QListView):
+    """Customized QListView class to detect right-click input.
+    
+    Signals:
+        fetched_event(QMouseEvent): The information of the click input is sent.
+    """
+    fetched_event = pyqtSignal(QMouseEvent)
```
kangz12345: Please use the camelcase: `fetchedEvent`.
BTW what is the meaning of its name? For me something like `rightButtonPressed` would be more natural.

iquip/apps/scheduler.py
```diff
@@ -223,6 +258,39 @@ def sizeHint(self, option: QStyleOptionViewItem, index: QModelIndex) -> QSize:
...
+            mode: The type of command that is requested to the server.
+            rid: The run identifier value of the target experiment.
+        """
+        super().__init__()
+        self.mode = mode
+        self.rid = rid
+
+    def run(self):
+        """Overridden."""
+        base_path = "http://127.0.0.1:8000/experiment/"
```
kangz12345: Well, it is quite confusing to use `PyQt` since they use camel case, but we decided to use camel case near the Qt-context.

See also: snu-quiqcl/qiwis#2

iquip/apps/scheduler.py
```diff
@@ -30,6 +34,40 @@ def _dismiss_items(layout: Optional[QLayout] = None):
                 _dismiss_items(item.layout())
 
 
+def _run_thread_with_worker(worker: QObject):
```
kangz12345: Nice idea to make it as a separate function.
However, please check that it is okay to
1. create `QThread` without `parent` (maybe that determines the thread where the new thread _lives_)
2. the `thread` object becomes not accessible after the function returns, and won't it make any issues (e.g., garbage collected)

In addition, I think this should also use camel case.

kangz12345 (changes requested):

iquip/apps/scheduler.py
giwon2004: For 1, I will add a `parent` argument to this function.

For 2, the lines
```
worker.done.connect(thread.quit)
worker.done.connect(worker.deleteLater)
thread.finished.connect(thread.deleteLater)
```
does the work.

iquip/apps/scheduler.py
giwon2004: So the full name of the function is `request termination` in artiq server, so I wanted to get that name preserved, but at the same time `terminate` is the URL here.

I will use `request termination` only on the menu display.

iquip/apps/scheduler.py
giwon2004: If I calculate the position in the `SchedulerFrame`, then it will be just right at the position `event.pos()`.

However when it is in the `SchedulerApp`, the header (image below) takes 25 of space:
image
and this `QPoint(0, 25)` is the adjustment of it.

giwon2004: However, the value of the margin changes when this is opened by a separate window, so I need to find a way to solve this.

iquip/apps/scheduler.py
giwon2004: For the camel case, this gives a pylint warning

```scheduler.py:37:0: C0103: Function name "_runThreadWithWorker" doesn't conform to snake_case naming style (invalid-name)```

File: iquip/apps/scheduler.py (hunk 1 above)
Comment (kangz12345) [reply iquip/apps/scheduler.py]: Okay, thank you for the explanation.
Then how about using `event.globalPos()` as shown in the _common usage_ example of [`QMenu.exec()`](https://doc.qt.io/qt-5/qmenu.html#exec-1)?

File: iquip/apps/scheduler.py (hunk 2 above)
Comment (kangz12345) [reply iquip/apps/scheduler.py]: However, the "API" is composed of the URL paths, not the artiq proxy server functions.
I think this design inserts an additional unnecessary layer (2):
```
0. user
1. -> GUI menu "Request termination"
2. -> mode "request_termination"
3. -> URL path "terminate"
4. -> server function "request_termination"
5. -> artiq scheduler
```
Isn't it more natural to match the `mode` to layer 3 ("terminate") instead of 4 ("request_terminate")?

File: iquip/apps/scheduler.py (hunk 5 above)
Comment (kangz12345) [reply iquip/apps/scheduler.py]: > For 1, I will add a `parent` argument to this function.

👍 

> For 2, the lines
> 
> ```
> worker.done.connect(thread.quit)
> worker.done.connect(worker.deleteLater)
> thread.finished.connect(thread.deleteLater)
> ```
> 
> does the work.

I was worrying about _unwanted_ garbage collection. 
As you experienced before, objects are garbage-collected when its reference is removed in Python.
For example, when we didn't assign an object to a variable, or the scope of a variable ended, the object is deleted by Python.
However in this case, as you said, `worker.done.connect(thread.quit)` would make a reference to `thread` from `worker` and `worker` is returned from the function, so it would be okay... I guess.

Comment (kangz12345) [reply iquip/apps/scheduler.py]: > For the camel case, this gives a pylint warning
> 
> `scheduler.py:37:0: C0103: Function name "_runThreadWithWorker" doesn't conform to snake_case naming style (invalid-name)`

Oh, where did you run pylint? You should run it in the repository root directory, since the `.pylintrc` file is located there. Or you have to set the rc file manually when you run the command.

Comment (giwon2004) [reply iquip/apps/scheduler.py]: > Oh, where did you run pylint? You should run it in the repository root directory, since the `.pylintrc` file is located there. Or you have to set the rc file manually when you run the command.

Can you tell me how I do that?

File: iquip/apps/scheduler.py (hunk 1 above)
Comment (giwon2004) [reply iquip/apps/scheduler.py]: Oh wow it worked, I did not know that existed. Thank you!

File: iquip/apps/scheduler.py (hunk 2 above)
Comment (giwon2004) [reply iquip/apps/scheduler.py]: Is it resolved now?

File: iquip/apps/scheduler.py (hunk 5 above)
Comment (kangz12345) [reply iquip/apps/scheduler.py]: Just go to the root directory `iquip` and run `pylint iquip`.
To designate an rcfile, run `pylint iquip --rcfile=.pylintrc`.
However, as the [documentation](https://pylint.pycqa.org/en/latest/user_guide/usage/run.html#command-line-options) says, it will automatically find `.pylintrc` file.

File: iquip/apps/scheduler.py [hunk 6]
```diff
@@ -302,7 +299,8 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         """Extended."""
         super().__init__(name, parent=parent)
         self.schedulerFrame = SchedulerFrame()
-        self.schedulerFrame.queueView.fetched_event.connect(self.displayMenu)
+        self.schedulerFrame.queueView.rightButtonPressed.connect(self.displayMenu)
+        self.addExperiment(ExperimentInfo("HI", {"rid": 1, "priority": 1}))
```
kangz12345: Is this for testing?

iquip/apps/scheduler.py
```diff
@@ -34,15 +34,15 @@ def _dismiss_items(layout: Optional[QLayout] = None):
...
+def _runThreadWithWorker(worker: QObject, parent: Optional[QObject] = None):
     """Runs another thread with given worker.
 
     Args:
         worker: The worker that must be run through another thread. It must have:
           - worker.run: the main function that has to be run.
           - worker.done: the signal that is emitted when the work is done.
     """
-    thread = QThread()
+    thread = QThread(parent = parent)
```
kangz12345: No spaces around `=` for keyword arguments!

kangz12345 (changes requested):

iquip/apps/scheduler.py
kangz12345: Yes, that's exactly what I thought.

iquip/apps/scheduler.py
kangz12345: Oh wait, I was confused. That was a function, not a method!
I am sorry for the confusion. I'm not sure which case (snake or camel) to use here...
Which do you prefer?

iquip/apps/scheduler.py
giwon2004: Oh, I forgot to delete it after testing the code 😅

iquip/apps/scheduler.py
giwon2004: I guess it is better to use the snake case, to have consistency with the function above!

kangz12345 (approved):

kangz12345:
@BECATRUE You can approve this (of course you can review this if you want).

BECATRUE:
> However, it is not fully tested as the experiments are refreshed so fast that it needs a clicking skill to terminate one.

If you want to make an experiment with a long running time, you can add just `time.sleep(...)` in `run()`.

iquip/apps/scheduler.py
```diff
@@ -234,6 +299,35 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         """Extended."""
         super().__init__(name, parent=parent)
         self.schedulerFrame = SchedulerFrame()
+        self.schedulerFrame.queueView.rightButtonPressed.connect(self.displayMenu)
+
```
BECATRUE, line 292: Please remove an empty line.
There are two empty ones!

iquip/apps/scheduler.py
```diff
@@ -198,10 +236,7 @@ def flags(self, index: QModelIndex) -> int:
 
 
 class ExperimentDelegate(QAbstractItemDelegate):
-    """Delegate for displaying the layout of each data in the experiment list.
-
-    TODO(giwon2004): Enabling buttons when displayed using QAbstractDelegate.
-    """
+    """Delegate for displaying the layout of each data in the experiment list."""
```
BECATRUE: "Delegates" looks more right!

giwon2004:
> If you want to make an experiment with a long running time, you can add just `time.sleep(...)` in `run()`.

Okay, I will try this later.

iquip/apps/scheduler.py
```diff
@@ -30,6 +34,40 @@ def _dismiss_items(layout: Optional[QLayout] = None):
...
 
+def _run_thread_with_worker(worker: QObject, parent: Optional[QObject] = None):
+    """Runs another thread with given worker.
+
+    Args:
+        worker: The worker that must be run through another thread. It must have:
+          - worker.run: the main function that has to be run.
+          - worker.done: the signal that is emitted when the work is done.
+    """
+    thread = QThread(parent=parent)
```
BECATRUE, line 45: Umm.. I think the `thread` will be removed just after finishing this function, then it may cause the error `destroyed a QThread object while running` you referred.

Is it possible to implement this function like [`explorer.py`](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/explorer.py#L118).

BECATRUE (changes requested):
Also, could you arrange the classes more legibily?

For example, I think it would be better if `ExperimentListView` and `RunningExperimentView` were together in front of or behind `SchedulerFrame`.

giwon2004:
Uh oh, this merge commit made something wrong.

giwon2004:
This PR will ruin many codes as it now has a history of changes in other unrelated codes.

I will make a new branch that has these things up to date with a clear history.

PR #105, opened 2023-08-11
Implement the functional tests for the builder app

### LINKED ISSUE #87: Implement the functional tests for the builder app
I will implement the functional tests for the builder app.

BECATRUE:
This PR will close #87.

I implemented the left unit tests and functional tests.
Now, the coverage rate has reached 100%.

### DIFF
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -112,6 +112,47 @@ class EnumerationEntryFunctionalTest(unittest.TestCase):
         self.assertEqual(str(cm.exception), msg)
 
 
+class NumberEntryFunctionalTest(unittest.TestCase):
+    """Functional tests for _NumberEntry class."""
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_init(self):
+        argInfo1 = {"unit": "s", "scale": 1, "step": 0.1, "min": None, "max": 100,
+                    "ndecimals": 3, "type": "float", "default": 20}
+        argInfo2 = {"unit": "s", "scale": 1, "step": 0.1, "min": 10, "max": None,
+                    "ndecimals": 3, "type": "float", "default": 20}
+        argInfo3 = {"unit": "s", "scale": 1, "step": 0.1, "min": 100, "max": 10,
+                    "ndecimals": 3, "type": "float", "default": 20}
+        for argName, argInfo, minValue, maxValue in (
+            ("name1", argInfo1, 0.0, 100),
+            ("name2", argInfo2, 10, 99.99),
+            ("name2", argInfo3, 10, 100)
+        ):
+            entry = builder._NumberEntry(argName, argInfo)
+            self.assertEqual(entry.spinBox.minimum(), minValue)
+            self.assertEqual(entry.spinBox.maximum(), maxValue)
+
+    def test_value(self):
+        argInfo1 = {"unit": "us", "scale": 1e-6, "step": 1e-7, "min": 10 * 1e-6, "max": 100 * 1e-6,
+                    "ndecimals": 3, "type": "float", "default": 20 * 1e-6}
+        argInfo2 = {"unit": "us", "scale": 1e-6, "step": 1e-7, "min": 10 * 1e-6, "max": 100 * 1e-6,
+                    "ndecimals": 3, "type": "float"}
+        argInfo3 = {"unit": "s", "scale": 1, "step": 1, "min": 10, "max": 100,
+                    "ndecimals": 0, "type": "int", "default": 20}
+        for argName, argInfo, value in (
+            ("name1", argInfo1, 20 * 1e-6),
+            ("name2", argInfo2, 10 * 1e-6),
+            ("name2", argInfo3, 20)
+        ):
+            entry = builder._NumberEntry(argName, argInfo)
+            self.assertEqual(entry.value(), value)
+
+
 class DateTimeEntryFunctionalTest(unittest.TestCase):
     """Functional tests for _DateTimeEntry class."""
 
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -152,6 +152,22 @@ class NumberEntryFunctionalTest(unittest.TestCase):
             entry = builder._NumberEntry(argName, argInfo)
             self.assertEqual(entry.value(), value)
 
+    def test_typical_scale(self):
+        """Tests when the scale for the unit is not typical."""
+        argInfo1 = {"unit": "us", "scale": 1e-6, "step": 1e-7, "min": 10 * 1e-6, "max": 100 * 1e-6,
+                    "ndecimals": 3, "type": "float", "default": 20 * 1e-6}
+        argInfo2 = {"unit": "us", "scale": 1e-7, "step": 1e-7, "min": 10 * 1e-6, "max": 100 * 1e-6,
+                    "ndecimals": 3, "type": "float", "default": 20 * 1e-6}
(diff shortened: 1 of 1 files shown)


tests/test_builder.py
```diff
@@ -27,6 +29,16 @@
     }
 }
 
+EXPERIMENT_INFO_2 = {
+    "name": "name2",
+    "arginfo": {
+        "arg4": [{"ty": "BooleanValue", "default": "default0"}, None, None],
+        "arg5": [{"ty": "StringValue", "default": "default1"}, None, None],
+        "arg6": [{"ty": "EnumerationValue", "default": "default2"}, None, None],
+        "arg7": [{"ty": "NumberValue", "default": "default3"}, None, None]
```
kangz12345: Not a big deal but why don't you change the `default` indices as well? (`"default4"`, etc.)

tests/test_builder.py
```diff
@@ -55,6 +67,149 @@ def test_value(self):
...
+            self.assertEqual(entry.value(), value)
+
+    def test_value_exception(self):
+        """Tests when the choices argument is empty."""
+        argName, argInfo = "name", {"choices": []}
+        msg = f"_EnumerationEntry {argName} with the empty choice"
+        with self.assertRaises(ValueError) as cm:
+            entry = builder._EnumerationEntry(argName, argInfo)
+            entry.value()
+        self.assertEqual(str(cm.exception), msg)
```
kangz12345: I think checking the error message is too "tight" and not scalable.
If we want to check specific errors, I think we should define error codes.
However, for this, just checking `ValueError` would be enough.

tests/test_builder.py
```diff
@@ -55,6 +67,149 @@ def test_value(self):
...
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_value(self):
+        for argName, argInfo, value in (
+            ("name1", {"choices": ["value1", "value2", "value3"], "default": "value2"}, "value2"),
```
kangz12345, line 98: I don't remember: what happens if `"default"` is not in `"choices"`?

tests/test_builder.py
```diff
@@ -55,6 +67,149 @@ def test_value(self):
...
+class NumberEntryFunctionalTest(unittest.TestCase):
+    """Functional tests for _NumberEntry class."""
+
+    def setUp(self):
+        self.qapp = QApplication([])
+
+    def tearDown(self):
+        del self.qapp
+
+    def test_init(self):
```
kangz12345: This is testing just the spinbox min/max values, so the name `test_init()` seems too broad.
How about change it to, e.g. `test_spinbox()`?
You can test `decimals()` as well.

tests/test_builder.py
```diff
@@ -55,6 +67,149 @@ def test_value(self):
...
+            ("name1", argInfo1, 0.0, 100),
+            ("name2", argInfo2, 10, 99.99),
+            ("name2", argInfo3, 10, 100)
+        ):
+            entry = builder._NumberEntry(argName, argInfo)
+            self.assertEqual(entry.spinBox.minimum(), minValue)
+            self.assertEqual(entry.spinBox.maximum(), maxValue)
+
+    def test_value(self):
+        argInfo1 = {"unit": "us", "scale": 1e-6, "step": 1e-7, "min": 10 * 1e-6, "max": 100 * 1e-6,
```
kangz12345: I think `10e-6` is more readable than `10 * 1e-6`.

tests/test_builder.py
```diff
@@ -55,6 +67,149 @@ def test_value(self):
...
+            entry = builder._NumberEntry(argName, argInfo)
+            self.assertEqual(entry.value(), value)
+
+    def test_typical_scale(self):
+        """Tests when the scale for the unit is not typical."""
+        argInfo1 = {"unit": "us", "scale": 1e-6, "step": 1e-7, "min": 10 * 1e-6, "max": 100 * 1e-6,
+                    "ndecimals": 3, "type": "float", "default": 20 * 1e-6}
+        argInfo2 = {"unit": "us", "scale": 1e-7, "step": 1e-7, "min": 10 * 1e-6, "max": 100 * 1e-6,
+                    "ndecimals": 3, "type": "float", "default": 20 * 1e-6}
+        with mock.patch("iquip.apps.builder.compute_scale") as mocked_compute_scale:
```
kangz12345: How about not mocking it since this is a functional test?

tests/test_builder.py
```diff
@@ -55,6 +67,149 @@ def test_value(self):
...
+            mocked_compute_scale.return_value = 1e-6
+            for argName, argInfo, is_typical in (
+                ("name1", argInfo1, True),
+                ("name2", argInfo2, False)
+            ):
+                entry = builder._NumberEntry(argName, argInfo)
+                warningLabelText = "" if is_typical else "Not a typical scale for the unit."
+                self.assertEqual(entry.warningLabel.text(), warningLabelText)
```
kangz12345: How about checking if the label text is empty or not, instead of matching the full text?

tests/test_builder.py
```diff
@@ -257,6 +442,23 @@ def test_submit(self, mocked_experiment_submit_thread_cls):
...
+        """Tests when argumentsFromListWidget() causes a ValueError."""
+        app = builder.BuilderApp(
+            name="name",
+            experimentPath="experimentPath",
+            experimentClsName="experimentClsName",
+            experimentInfo=copy.deepcopy(EMPTY_EXPERIMENT_INFO)
+        )
+        with mock.patch.object(app, "argumentsFromListWidget") as mocked_arguments_from_list_widget:
+            mocked_arguments_from_list_widget.side_effect = ValueError
+            logger = logging.getLogger()
```
kangz12345: How about designating the source logger directly: `"iquip.apps.builder"`?

kangz12345 (changes requested):

tests/test_builder.py
BECATRUE: I didn't even think of that😅 I will update it.

tests/test_builder.py
BECATRUE: Okay, then I will remove the part checking the detailed message.

tests/test_builder.py
BECATRUE: When fetching the experiment arguments information, the artiq-master makes an error as follows:
image

Thus, the proxy server returns a 500 Internal Server Error.

tests/test_builder.py
BECATRUE: I agree with you! Also, I will add tests about `decimals()`.

tests/test_builder.py
BECATRUE: It looks more appropriate!

tests/test_builder.py
BECATRUE: Similar to the review above, I will only test whether or not.

tests/test_builder.py
kangz12345: Thank you for the explanation. Then does the builder app handle such errors?

BECATRUE: The error occurs when fetching the experiment info, thus it is handled in `explorer.py`.

If an Internal Server Error is received, this [`try`](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/thread.py#L60) statement catches the exception.

kangz12345: Oh I see. Thank you!

kangz12345 (approved):
LGTM

BECATRUE:
Could you check this PR? @Aijuh

Aijuh (approved):
LGTM

PR #101, opened 2023-08-09
Add `logger` back

giwon2004:
This reverts a wrong commit made by me while reverting "merge commit to the branch" back mentioned in [`85a6eb5`#commitcomment-123892867](https://github.com/snu-quiqcl/iquip/commit/85a6eb5821d67d0f7ed491eee43b5123c933e072#commitcomment-123892867)

This closes #100.

PR #108, opened 2023-08-14
fix setup.json

### LINKED ISSUE #107: Remove duplicated codes in setup.json
### Remove duplicated codes in setup.json

There are duplicated codes in the recently merged branch.
This happened while managing conflict resolve before merging.
Plan to handle this issue by removing duplicated part.

image

Aijuh:
Fixed duplicated codes in setup.json.

This error occured because of repeated updates in setup.json.
I first added `scheduler` info in closed PR #88.
And @giwon2004 also added `logger` info in closed PR #101.

This happened because I couldn't see `scheduler` in setup.json and @giwon2004  couldn't see `logger` in setup.json.
So we added both app again and that difference is added to the develop branch, which caused this error.

This conflict would be resolved by this PR.
This closes #107.

### DIFF
--- a/setup.json
+++ b/setup.json
@@ -20,18 +20,6 @@
             "cls": "SchedulerApp",
             "show": true,
             "pos": "bottom"
-        },
-        "scheduler": {
-            "module": "iquip.apps.scheduler",
-            "cls": "SchedulerApp",
-            "show": true,
-            "pos": "bottom"
-        },
-        "logger": {
-            "module": "iquip.apps.logger",
-            "cls": "LoggerApp",
-            "show": true,
-            "pos": "bottom"
         }
     }
 }


giwon2004 (approved):

setup.json
```diff
@@ -20,18 +20,6 @@
             "cls": "SchedulerApp",
             "show": true,
             "pos": "bottom"
-        },
```
giwon2004, line 23: How about making the `logger` located at the bottom? I think the logger must wrap all the others up.

Aijuh: Good point. 
Since it determines overall design, I want to know how @kangz12345 and @BECATRUE think about this.
I will update it after their review.

Aijuh:
By the review of @giwon2004, I think it is good to determine initial position arrangement of apps.
By setting position of apps as follows,

explore - left,
scheduler - left,
logger - bottom,

following GUI format is made.

If I set scheduler's position as right, when I open experiment, builder is created right above the scheduler. 

I think this arrangement seems clear.
If you have other idea or want to change something, please tell me.

image

setup.json
kangz12345: Well, I think changing the item arrangement in the json file does not matter. It seems okay to locate the "logger" part at the bottom of the json file.

kangz12345:
I think @giwon2004 didn't mean to change the position of "scheduler" to the "left". Isn't it too narrow for the scheduler?

Aijuh:
This is the design of the previous one.
Do you prefer this one?

image

BECATRUE:
Soon there will be a big change in the widget arrangement in qiwis, and accordingly there will also be a change in the arrangement method in iquip.
You won't have to worry too much about the location right now.

BECATRUE (approved):

PR #111, opened 2023-08-16
Add menu for queued experiments

### LINKED ISSUE #96: Displaying a menu widget when an experiment on the scheduler is right-clicked

giwon2004:
### This is a new PR made from the contaminated pull request #102.

I made a `QMenu` object that displays a menu when right-clicked:
image
(The space in front of the text will be deleted in the future, using `QMenu.setStyleSheet`.)

Here, Edit will be future implemented, and the other two are connected to POST request mentioned in [artiq-proxy#44](https://github.com/snu-quiqcl/artiq-proxy/issues/44) and #96.

This closes #96.


giwon2004:
The pylint & unit tests' errors are not caused by this commit.

iquip/apps/scheduler.py
```diff
@@ -30,6 +34,24 @@ def _dismiss_items(layout: Optional[QLayout] = None):
...
+def _run_thread_with_worker(worker: QObject, parent: Optional[QObject] = None):
+    """Runs another thread with given worker.
+
+    Args:
+        worker: The worker that must be run through another thread. It must have:
+          - worker.run: the main function that has to be run.
+          - worker.done: the signal that is emitted when the work is done.
+    """
+    thread = QThread(parent=parent)
+    worker = SchedulerPostWorker("delete")
```
kangz12345: Why is `worker` overwritten here?

iquip/apps/scheduler.py
```diff
@@ -59,6 +81,22 @@ def __init__(self, parent: Optional[QWidget] = None):
         layout.addWidget(self.queueView)
 
 
+class ExperimentListView(QListView):
+    """Customized QListView class to detect right-click input.
+    
+    Signals:
+        rightButtonPressed(QMouseEvent): The information of the click input is sent.
```
kangz12345: In the docstring for signals, we use intuitive argument names instead of their types, since the types are already in the codes right below.
For example, `rightButtonPressed(mouseEvent)`.

iquip/apps/scheduler.py
```diff
@@ -59,6 +81,22 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+    """
+    rightButtonPressed = pyqtSignal(QMouseEvent)
+
+    def mousePressEvent(self, event):
+        """Overridden."""
+        if event.type() == QEvent.MouseButtonPress:
+            if event.button() == Qt.RightButton:
+                self.rightButtonPressed.emit(event)
+        super().mousePressEvent(event) # hand the signal to drag & drop
```
kangz12345: Please make 2 spaces before comments (`__#` instead of `_#`).

iquip/apps/scheduler.py
```diff
@@ -59,6 +81,22 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+    
+    Signals:
+        rightButtonPressed(QMouseEvent): The information of the click input is sent.
+    """
+    rightButtonPressed = pyqtSignal(QMouseEvent)
+
+    def mousePressEvent(self, event):
+        """Overridden."""
+        if event.type() == QEvent.MouseButtonPress:
+            if event.button() == Qt.RightButton:
```
kangz12345: Why don't you use `and` instead of nesting 2 `if` statements? Is it for future use?

iquip/apps/scheduler.py
```diff
@@ -223,6 +288,33 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+            event: Information holder containing the clicked position.
+        """
+        for i in range(self.schedulerFrame.model.rowCount()):
+            index = self.schedulerFrame.model.index(i)
+            if self.schedulerFrame.queueView.rectForIndex(index).contains(event.pos()):
+                rid = self.schedulerFrame.model.data(index).rid
+                menu = QMenu(self.schedulerFrame)
+                edit = menu.addAction("Edit")
+                delete = menu.addAction("Delete")
+                request_termination = menu.addAction("Request termination")
```
kangz12345: Is it common to create menu instances every time? Is it hard to reuse a single `QMenu` instance?

kangz12345 (changes requested):

kangz12345:
Please resolve the pylint warning.

iquip/apps/scheduler.py
```diff
@@ -282,13 +280,22 @@ class SchedulerApp(qiwis.BaseApp):
...
     """
 
     def __init__(self, name: str, parent: Optional[QObject] = None):
         """Extended."""
         super().__init__(name, parent=parent)
         self.schedulerFrame = SchedulerFrame()
         self.schedulerFrame.queueView.rightButtonPressed.connect(self.displayMenu)
+        self.menu = QMenu(self.schedulerFrame)
+        # TODO(giwon2004) Remove icon space from the menu list (use menu.setStyleSheet)
+        self.signals = {
```
kangz12345: Why is the name "signals", not "actions"?

iquip/apps/scheduler.py
```diff
@@ -30,6 +34,23 @@ def _dismiss_items(layout: Optional[QLayout] = None):
                 _dismiss_items(item.layout())
 
 
+def _run_thread_with_worker(worker: QObject, parent: Optional[QObject] = None):
+    """Runs another thread with given worker.
```
kangz12345: The name seems awkward since it does not actually run the thread, but create and prepare the threaded worker.

kangz12345 (changes requested):

iquip/apps/scheduler.py
```diff
@@ -34,7 +34,7 @@ def _dismiss_items(layout: Optional[QLayout] = None):
                 _dismiss_items(item.layout())
 
 
-def _run_thread_with_worker(worker: QObject, parent: Optional[QObject] = None):
+def _thread_with_worker(worker: QObject, parent: Optional[QObject] = None):
     """Runs another thread with given worker.
```
kangz12345: Good, now please annotate the return type: `def _thread_with_worker(...) -> QThread:`.
Also, please update the docstring summary as well: `"""Returns a new thread with given worker.`

kangz12345 (changes requested):
Do you know why the pylint and unit test actions fail?

giwon2004:
> Do you know why the pylint and unit test actions fail?

We discussed about this in the previous meeting, it is about #112. I made this branch from `develop` after the commit that made the errors.

kangz12345 (approved):

iquip/apps/scheduler.py
```diff
@@ -30,6 +34,23 @@ def _dismiss_items(layout: Optional[QLayout] = None):
                 _dismiss_items(item.layout())
 
 
+def _thread_with_worker(worker: QObject, parent: Optional[QObject] = None) -> QThread:
+    """Returns a new thread with given worker.
+
+    Args:
+        worker: The worker that must be run through another thread. It must have:
```
BECATRUE, line 41: Please add a description about `parent`.

iquip/apps/scheduler.py
```diff
@@ -59,6 +80,21 @@ def __init__(self, parent: Optional[QWidget] = None):
         layout.addWidget(self.queueView)
 
 
+class ExperimentListView(QListView):
+    """Customized QListView class to detect right-click input.
```
BECATRUE, line 85: I think this description is too minor.

How about writing that this list view contains the running experiments.

iquip/apps/scheduler.py
```diff
@@ -59,6 +80,21 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+    
+    Signals:
+        rightButtonPressed(mouseEvent): The information of the click input is sent.
+    """
+    rightButtonPressed = pyqtSignal(QMouseEvent)
+
+    def mousePressEvent(self, event):
```
BECATRUE: Can you add the type hinting about `event`?

tests/test_scheduler.py
```diff
@@ -123,8 +119,8 @@ def test_modify_experiment(self):
         experiment_change = SubmittedExperimentInfo(rid=1, priority=1)
         experiment_new_info = SubmittedExperimentInfo(rid=1, priority=2)
         experiment_delete = SubmittedExperimentInfo(rid=2, priority=1)
-        app.schedulerFrame.model.experimentQueue.append(experiment_delete)
-        app.schedulerFrame.model.experimentQueue.append(experiment_change)
+        app.addExperiment(experiment_delete)
```
BECATRUE, line 122: Please modify the test function name!
I think it should be updated to `test_add_experiment()`.

iquip/apps/scheduler.py
```diff
@@ -212,17 +245,77 @@ def sizeHint(self, option: QStyleOptionViewItem, index: QModelIndex) -> QSize:
...
         super().__init__(name, parent=parent)
         self.schedulerFrame = SchedulerFrame()
+        self.schedulerFrame.queueView.rightButtonPressed.connect(self.displayMenu)
+        self.menu = QMenu(self.schedulerFrame)
+        # TODO(giwon2004) Remove icon space from the menu list (use menu.setStyleSheet)
+        self.actions = {
+            "edit": self.menu.addAction("Edit"),
+            "delete": self.menu.addAction("Delete"),
+            "terminate": self.menu.addAction("Request termination")
+        }
```
BECATRUE, line 303: 👍

iquip/apps/scheduler.py
```diff
@@ -212,17 +245,77 @@ def sizeHint(self, option: QStyleOptionViewItem, index: QModelIndex) -> QSize:
...
+    @pyqtSlot(QMouseEvent)
+    def displayMenu(self, event: QMouseEvent):
+        """Displays the menu pop-up.
+
+        Args:
+            event: Information holder containing the clicked position.
+        """
+        for i in range(self.schedulerFrame.model.rowCount()):
+            index = self.schedulerFrame.model.index(i)
+            if self.schedulerFrame.queueView.rectForIndex(index).contains(event.pos()):
```
BECATRUE, line 315: Oh.. Is it the best solution?

Sure, this code seems to work well so you don't have to modify it now, but it looks little convoluted.

iquip/apps/scheduler.py
```diff
@@ -212,17 +245,77 @@ def sizeHint(self, option: QStyleOptionViewItem, index: QModelIndex) -> QSize:
...
+                rid = self.schedulerFrame.model.data(index).rid
+                action = self.menu.exec_(event.globalPos())
+                if action == self.actions["edit"]:
+                # TODO(giwon2004) Create an app for editing scannables.
+                    pass
+                elif action == self.actions["delete"]:
+                    _thread_with_worker(SchedulerPostWorker("delete", rid), self).start()
```
BECATRUE: Unless the `QThread` object isn't stored as an attribute of this class, does it work well?

I'm worried becuase the thread object may be removed before it finishes because the thread is defined only in `displayMenu()`.

iquip/apps/scheduler.py
```diff
@@ -212,17 +245,77 @@ def sizeHint(self, option: QStyleOptionViewItem, index: QModelIndex) -> QSize:
...
+            rid: The run identifier value of the target experiment.
+        """
+        super().__init__()
+        self.mode = mode
+        self.rid = rid
+
+    def run(self):
+        """Overridden."""
+        basePath = "http://127.0.0.1:8000/experiment/"
+        requests.post(basePath + self.mode, params={"rid": self.rid}, timeout=10)
```
BECATRUE: How simple it is!

BECATRUE (changes requested):
Did you check if the thread works well and doesn't block the main event loop?

iquip/apps/scheduler.py
giwon2004: True, but I could not find the function that directly gives the index :(

tests/test_scheduler.py
giwon2004: I do not get it, this part tests `app.changeExperiment` too, as a functional test.

iquip/apps/scheduler.py
giwon2004: I have tested the thread object, and the termination of `displayMenu` did not delete it.
I guess it works well as the thread has a parent `self`.

However, I found a new problem: it does not run `worker.run` here! What could be the problem?

iquip/apps/scheduler.py
BECATRUE: Please check it! @giwon2004

iquip/apps/scheduler.py
BECATRUE: > I have tested the thread object, and the termination of displayMenu did not delete it.
I guess it works well as the thread has a parent self.

Good!

BECATRUE: > However, I found a new problem: it does not run worker.run here! What could be the problem?

What do you mean?
Even if the thread starts, isn't `worker.run()` called?

BECATRUE (changes requested):

iquip/apps/scheduler.py
giwon2004: > Even if the thread starts, isn't worker.run() called?

No, it does not call the function. :(

iquip/apps/scheduler.py
giwon2004: I did this under this line! Github only shows the modification on that exact line :(

BECATRUE: Oh, you added a description under this line!
I'm sorry I didn't check well.

iquip/apps/scheduler.py
BECATRUE: Umm.. I think the problem should be resolved in this PR..
I also try to find out the reasons, so please hold for a moment.

BECATRUE (changes requested):

iquip/apps/scheduler.py
BECATRUE: How did you check the problem?
In this branch, fetching the actual list from the proxy server isn't implemented yet, so I can't open the menu.

giwon2004: I made a dummy experiment by manually adding an experiment by `self.addExperiment`. 
Then I used right-click on it and clicked Edit.

Here I made the thread have nothing to do (command-ize the `request` part), but the error message `QThread destroyed while running` showed up after I quit the app while it did not show when the function `_thread_with_worker` is not called.

This showed the thread does not end when the call itself dies, but also showed the thread does not end _until the end_.

giwon2004:
I am not sure whether using the same `self.worker` works when the menu is clicked twice, and this needs a test.

BECATRUE:
> I am not sure whether using the same `self.worker` works when the menu is clicked twice, and this needs a test.

If so, it may be a simple solution that makes the menu or the button disable until `self.worker` finishes.

BECATRUE (approved):
LGTM!

giwon2004:
> If so, it may be a simple solution that makes the menu or the button disable until `self.worker` finishes.

No, this is about the case when the menu thread itself is called twice. (Making a new thread with an old worker)

PR #113, opened 2023-08-16
Update unittest for logger.py

### LINKED ISSUE #112: Implement unittest for file saving logger
Implement unittest for file saving logger

Currently logger unittest has error because implementation of logger.py is updated.
I will update unittest for logger as fast as possible.

Aijuh:
This closes #112 

I updated unittest for logger.py. 
Now, unittest and pylint has error at develop branch since I implemented unittest of old version of logger app and merge it.

This unittest does not have coverage rate 100% because of function `namer` inside `initLogger` in `LoggerApp`.
I don't have good idea about implementing unittest for function defined inside function.
So I am planning to make namer as private class method `_namer` if no idea is suggested.

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -21,7 +21,7 @@ class LoggingHandlerTest(unittest.TestCase):
         self.app = logger.LoggerApp(name="name", logFilePath=self.log_dir, parent=QObject())
 
     def tearDown(self):
-        self.app.removeHandler()
+        self.app.fileHandler.close()
         file_list = os.listdir(self.log_dir)
         for file in file_list:
             os.remove(os.path.join(self.log_dir, file))
@@ -47,7 +47,7 @@ class ConfirmClearingFrameTest(unittest.TestCase):
         self.app = logger.LoggerApp(name="name", logFilePath=self.log_dir, parent=QObject())
 
     def tearDown(self):
-        self.app.removeHandler()
+        self.app.fileHandler.close()
         file_list = os.listdir(self.log_dir)
         for file in file_list:
             os.remove(os.path.join(self.log_dir, file))
@@ -76,7 +76,7 @@ class LoggerAppTest(unittest.TestCase):
         self.app = logger.LoggerApp(name="name", logFilePath=self.log_dir, parent=QObject())
 
     def tearDown(self):
-        self.app.removeHandler()
+        self.app.fileHandler.close()
         file_list = os.listdir(self.log_dir)
         for file in file_list:
             os.remove(os.path.join(self.log_dir, file))
@@ -124,7 +124,7 @@ class LoggerAppTest(unittest.TestCase):
             app = logger.LoggerApp(name="name", logFilePath=self.log_dir, parent=QObject())
             app.loggerFrame.clearButton.clicked.emit()
             mocked_method.assert_called_once()
-            app.removeHandler()
+            app.fileHandler.close()
 
     def test_check_to_clear(self):
         with mock.patch("iquip.apps.logger.QWidget.show") as mocked_show:
@@ -136,7 +136,7 @@ class LoggerAppTest(unittest.TestCase):
             app = logger.LoggerApp(name="name", logFilePath=self.log_dir, parent=QObject())
             app.confirmFrame.buttonBox.accepted.emit()
             mocked_method.assert_called_once()
-            app.removeHandler()
+            app.fileHandler.close()
 
     @mock.patch("iquip.apps.logger.QTextEdit.clear")
     def test_clear_log(self, mocked_clear):
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -1,5 +1,6 @@
 """Unit tests for logger module."""
 
+import os
 import logging
 import unittest
 from unittest import mock
@@ -14,16 +15,24 @@ class LoggingHandlerTest(unittest.TestCase):
 
     def setUp(self):
         self.qapp = QApplication([])
+        self.logdir = "logfile-unittest"
+        if not os.path.exists(self.logdir):
+            os.mkdir(self.logdir)
+        self.app = logger.LoggerApp(name="name", logFilePath=self.logdir, parent=QObject())
 
     def tearDown(self):
+        self.app.removeHandler()
+        file_list = os.listdir(self.logdir)
+        for file in file_list:
+            os.remove(self.logdir+"/"+file)
+        os.rmdir(self.logdir)
         del self.qapp
 
     def test_emit(self):
         with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
-            app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
-            app.setLevel(app.frameHandler ,"DEBUG")
+            self.app.setLevel(self.app.frameHandler ,"DEBUG")
             test_log = logging.makeLogRecord({"name": "hello"})
-            app.frameHandler.emit(test_log)
+            self.app.frameHandler.emit(test_log)
             mocked_signal.emit.assert_called_once()
 
 
@@ -32,20 +41,27 @@ class ConfirmClearingFrameTest(unittest.TestCase):
 
     def setUp(self):
         self.qapp = QApplication([])
+        self.logdir = "logfile-unittest"
+        if not os.path.exists(self.logdir):
+            os.mkdir(self.logdir)
+        self.app = logger.LoggerApp(name="name", logFilePath=self.logdir, parent=QObject())
 
     def tearDown(self):
+        self.app.removeHandler()
+        file_list = os.listdir(self.logdir)
+        for file in file_list:
+            os.remove(self.logdir+"/"+file)
+        os.rmdir(self.logdir)
         del self.qapp
 
     def test_button_ok_clicked(self):
         with mock.patch("iquip.apps.logger.QWidget.close") as mocked_close:
-            app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
-            app.confirmFrame.buttonOKClicked()
+            self.app.confirmFrame.buttonOKClicked()
             mocked_close.assert_called_once()
 
     def test_button_cancel_clicked(self):
         with mock.patch("iquip.apps.logger.QWidget.close") as mocked_close:
-            app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
-            app.confirmFrame.buttonCancelClicked()
+            self.app.confirmFrame.buttonCancelClicked()
             mocked_close.assert_called_once()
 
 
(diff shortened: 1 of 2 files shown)


tests/test_logger.py
```diff
@@ -20,10 +20,10 @@ def tearDown(self):
 
     def test_emit(self):
         with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
-            app = logger.LoggerApp(name="name", parent=QObject())
-            app.setLevel("DEBUG")
+            app = logger.LoggerApp(name="name", logFilePath = "", parent=QObject())
```
giwon2004: We do not use whitespaces around `=` for keyword arguments.

There are more below, please change all of them.

giwon2004 (changes requested):
Only one thing!

tests/test_logger.py
Aijuh: Thank you for review.
I missed it.

Aijuh:
Still waiting for the idea about implementing unittest of function defined inside function. I also searched for it, but cannot find valid reference. Otherwise I will make function `namer` as method of LoggerApp.

Unittest build 3.9 sometimes has following error

> qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

This error seems to come from `explore.py`.

BECATRUE:
> Unittest build 3.9 sometimes has following error

I don't know the specific reason, but sometimes the unittest may be failed though it is implemented correctly.
Then, you can just enter the details and re-run it!

image

tests/test_logger.py
```diff
@@ -67,51 +72,53 @@ def test_set_level(self):
...
 
     def test_frames(self):
-        app = logger.LoggerApp(name="name", parent=QObject())
+        app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
         self.assertEqual(app.frames(), (app.loggerFrame,))
 
     def test_call_check_to_clear(self):
         with mock.patch("iquip.apps.logger.LoggerApp.checkToClear") as mocked_method:
-            app = logger.LoggerApp(name="name", parent=QObject())
+            app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
```
BECATRUE: Is it possible to move this statement outside of `with` block?

I think it looks better like `test_add_log()`.

tests/test_logger.py
```diff
@@ -67,51 +72,53 @@ def test_set_level(self):
...
         }
         root_logger = logging.getLogger()
-        app = logger.LoggerApp(name="name", parent=QObject())
+        app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
         for levelText, level in levels.items():
-            app.setLevel(levelText)
-            self.assertEqual(app.handler.level, level)
+            app.setLevel(app.frameHandler, levelText)
+            app.setLevel(app.fileHandler, levelText)
+            self.assertEqual(app.frameHandler.level, level)
```
BECATRUE, line 103: Please check `app.fileHandler.level`, too!

tests/test_logger.py
```diff
@@ -67,51 +72,53 @@ def test_set_level(self):
...
             self.assertEqual(root_logger.level, level)
-        prev_level = app.handler.level
+        prev_level = app.frameHandler.level
         non_level = "non_level"
-        app.setLevel(non_level)
-        self.assertEqual(prev_level, app.handler.level)
+        app.setLevel(app.frameHandler, non_level)
+        app.setLevel(app.fileHandler, non_level)
+        self.assertEqual(prev_level, app.frameHandler.level)
```
BECATRUE: Check `fileHandler`!

BECATRUE (changes requested):
Umm.. Can you check `namer()` though checking `self.fileHandler.namer`?

For example,
```python
app.initLogger()

name = "blah"
modified_name = app.frameHandler.namer(name)

self.assertEqual(modified_name, "blah.log")
```

I'm not sure, but it seems to implement the unittest for `namer()`.

tests/test_logger.py
Aijuh: It is not possible because logger app's method is mocked.
When I move it outside of `with` block, the `checkToClear` method didn't called.
I think if app is created before `with` statement, app's `checkToClear` method is not same as mocked method.

Aijuh:
I added code for unittest of namer and code for checking level of fileHandler.
I implemented unittest of namer by following suggestion from @BECATRUE.
Thank you.

giwon2004 (approved):

giwon2004:
LGTM!

tests/test_logger.py
```diff
@@ -20,10 +20,10 @@ def tearDown(self):
 
     def test_emit(self):
         with mock.patch("iquip.apps.logger._Signaller.signal") as mocked_signal:
-            app = logger.LoggerApp(name="name", parent=QObject())
-            app.setLevel("DEBUG")
+            app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
```
BECATRUE: It looks better to assign a non-empty string to `logFilePath`.
Is it possible?

tests/test_logger.py
BECATRUE: Okay, then remain it as it is!

tests/test_logger.py
```diff
@@ -67,51 +72,61 @@ def test_set_level(self):
...
             mocked_logger.info.called_once_with("Tried to clear logs by clicking clear button")
 
     def test_add_log(self):
+        app = logger.LoggerApp(name="name", logFilePath="", parent=QObject())
         with mock.patch("iquip.apps.logger.QTextEdit.insertPlainText") as mocked_insert:
             with mock.patch("iquip.apps.logger.QDateTime.currentDateTime") as mocked_time:
-                app = logger.LoggerApp(name="name", parent=QObject())
                 app.addLog("hello")
                 timeString = mocked_time().toString("yyyy-MM-dd HH:mm:ss")
                 msg = f"{timeString}: hello\n"
```
BECATRUE, line 153: Oh, now I realize it.
Please add an empty line between this class and `if __name__ == "__main__":`

BECATRUE (changes requested):

tests/test_logger.py
Aijuh: If I added `logFilepath = "test"`, it said directory `\\iquip\\test\\log_230821-144124.log'` is not exist.

BECATRUE: Oh, I see. Then if we set `logFilePath` to an empty string, where is the log file created, in the iquip directory or the called directory?

BECATRUE: Also, does the unittest make a log file? If so, the file should be removed after the test.

BECATRUE (changes requested):

Aijuh:
Good point. I should have checked that.
Since the test is running at the `iquip` directory, file is created at that directory with designated file name.
I will make code for removing that created dummy file.

kangz12345 (approved):
Just approving since I am a code owner (not reviewed).
Please mention me if you need my opinions.

Aijuh:
I added codes for removing dummy file created by file handler.
In order to remove file, I should first discard connection to the open file.

I want to define that codes inside destructor `__del__` and call it by `del app`. But when I call `del app` and try to remove file, destructor does not called and the file is still opened. So I cannot remove dummy file.
As I understand, destructor is called immediately when reference count becomes zero. But as I think no one refer to the created app, but the desctuctor does not called. I cannot figure out way to use `__del__` method.
Therefore I made another method inside class `LoggerApp` in `logger.py`.

tests/test_logger.py
```diff
@@ -14,16 +15,24 @@ class LoggingHandlerTest(unittest.TestCase):
...
+        self.logdir = "logfile-unittest"
+        if not os.path.exists(self.logdir):
+            os.mkdir(self.logdir)
+        self.app = logger.LoggerApp(name="name", logFilePath=self.logdir, parent=QObject())
 
     def tearDown(self):
+        self.app.removeHandler()
+        file_list = os.listdir(self.logdir)
+        for file in file_list:
+            os.remove(self.logdir+"/"+file)
```
BECATRUE: Please use `os.path.join()`.

tests/test_logger.py
```diff
@@ -14,16 +15,24 @@ class LoggingHandlerTest(unittest.TestCase):
 
     def setUp(self):
         self.qapp = QApplication([])
+        self.logdir = "logfile-unittest"
```
BECATRUE: How about `log_dir`?

BECATRUE: Please add a trailing "/" to show that the path is a directory explicitly.

BECATRUE: Also, I think the directory path should be a full path!

BECATRUE (changes requested):

BECATRUE:
> I added codes for removing dummy file created by file handler. In order to remove file, I should first discard connection to the open file.
> 
> I want to define that codes inside destructor `__del__` and call it by `del app`. But when I call `del app` and try to remove file, destructor does not called and the file is still opened. So I cannot remove dummy file. As I understand, destructor is called immediately when reference count becomes zero. But as I think no one refer to the created app, but the desctuctor does not called. I cannot figure out way to use `__del__` method. Therefore I made another method inside class `LoggerApp` in `logger.py`.

Umm... We don't use `removeHandler()` while implementing, so I think it should be removed.
However, as you said, it looks a difficult problem.

Did you try overriding the destructor of `LoggerApp` and implementing the contents of `removeHandler()` in the customized destructor?

tests/test_logger.py
Aijuh: I thought that Unit test is executed at the location of `iquip`.
So I make another directory for dummy log file at the `iquip`, like `iquip/logfile-unittest` and delete that directory.
I don't know the exact directory where unittest is executed. 

Setting full path as `os.getcwd()+self.log_dir` would be same as previous implementation. Do you prefer this way? Or did I missed something?

Aijuh:
@BECATRUE 
`LoggerApp`'s parent class is `Baseapp` in qiwis.py, which does not have `__del__` method.
I tested for it but it does not work. I read that `del` method deletes class during garbage collection in python. 

There is way to make destructor as `__del__` and directly call `self.app.__del__()`. But is incurs pylint error.

tests/test_logger.py
BECATRUE: As @Aijuh and I discussed yesterday, the current implementation doesn't seem to matter, so it is okay to remain it as it is.

Please just add a trailing "/"!

BECATRUE:
> @BECATRUE `LoggerApp`'s parent class is `Baseapp` in qiwis.py, which does not have `__del__` method. I tested for it but it does not work. I read that `del` method deletes class during garbage collection in python.
> 
> There is way to make destructor as `__del__` and directly call `self.app.__del__()`. But is incurs pylint error.

@Aijuh 
I see. Then, how about moving `removeHandler()` into `test_logger.py`?
I still want to remove the method from `LoggerApp`.

Aijuh:
Sorry for the late update, @BECATRUE. 
Only `fileHandler.close()` is needed to close the open file.
So I removed the method at `logger.py` and add code `fileHandler.close()` at `test_logger.py`.

BECATRUE:
> Sorry for the late update, @BECATRUE. Only `fileHandler.close()` is needed to close the open file. So I removed the method at `logger.py` and add code `fileHandler.close()` at `test_logger.py`.

Wow, what a simple solution!

BECATRUE (approved):
LGTM!

PR #125, opened 2023-09-12
`NDArrayViewer` and `CurvePlotViewer`

### LINKED ISSUE #122: `NDArrayViewer`
`NDArrayViewer` is an abstract interface for visualizing a `numpy.ndarray`.
It will be parameterized by the dimension of the array, e.g., it rejects the input whose dimension is not correct.
This issue introduces `NDArrayViewer` with `CurvePlotViewer`, which is a concrete subclass as an example.
`CurvePlotViewer` takes a single 1D array and its axis information, then plots it.

For the graphical features, I will exploit [`pyqtgraph`](https://www.pyqtgraph.org) as `artiq` uses it for applets.
Even though I designed an interface, I don't have any big plan for it.
At the moment, it won't include fancy features.
However, this dimension-flexible approach will allow various viewer implementations.

### Additional context
Direct conversion from h5 files is not considered here. Only interfaces for `numpy.ndarray` will be supported for now.

kangz12345:
This PR include a new module `iquip.apps.dataviewer` which contains `NDArrayViewer` and `CurvePlotViewer`.
It also introduces a new dataclass `AxisInfo`.

The basic idea for plotting an ndarray:
1. We are interested in target measurement data, e.g., photon counts.
2. We perform "scan" experiment, e.g., repeat 200shots and scan time, frequency, etc.
3. Each scanning parameter (shot, time, frequency, etc.) adds 1 dimension to the resulting ndarray.
4. To plot it, we need information for each axis, i.e., scanning parameter. `AxisInfo` represents this information.

As an example, I provide another branch: `kangz12345/122/ndarrayviewer-demo`.
It includes a plot widget and a push button.
If you click the button, it will randomly generate a 4-sequence sampled in {1, 2, 3}.
Then it multiplies `1e-3` to make its scale to m-.
The left (vertical) axis is labeled by "Voltage" with unit "V" (however, `pyqtgraph` automatically displays the unit as "mV" since the values are in that scale).
`AxisInfo` contains the scanning parameter information, which is time.
Screen Shot 2023-09-12 at 19 39 50

For our usual cases, we need to preprocess the data to make it a 1D array to use `CurvePlotViewer`. I will develop this in the next PRs.

NOTE: I will implement several features WITHOUT unit tests for few weeks, because we need such features as soon as possible. I will implement unit tests later, after finishing mandatory implementations.

This closes #122.

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -6,6 +6,7 @@ import logging
 from typing import Sequence, Optional
 
 import numpy as np
+import pyqtgraph as pg
 
 logger = logging.getLogger(__name__)
 
@@ -64,3 +65,30 @@ class NDArrayViewer(metaclass=abc.ABCMeta):
         for size, info in zip(data.shape, axes):
             if size != len(info.values):
                 raise ValueError(f"Size mismatch in {info}: expected {size} values")
+
+
+class CurvePlotViewer(NDArrayViewer):
+    """Plot viewer for visualizing a 2D curve.
+    
+    Attributes:
+        plotItem: The PlotItem for showing the curve plot.
+        plotWidget: The PlotWidget which contains the plotItem.
+        curve: The PlotDataItem which represents the curve plot.
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Args:
+            **kwargs: Passed as keyword arguments to instantiate a PlotItem.
+        """
+        super().__init__(ndim=1)
+        self.plotItem = pg.PlotItem(**kwargs)
+        self.plotWidget = pg.PlotWidget(plotItem=self.plotItem)
+        self.curve = self.plotItem.plot()
+
+    def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
+        """Extended."""
+        super().setData(data, axes)
+        axis = axes[0]
+        self.plotItem.setLabel(axis="bottom", text=axis.name, units=axis.unit)
+        self.curve.setData(axis.values, data)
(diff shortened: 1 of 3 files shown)


iquip/apps/dataviewer.py
```diff
@@ -0,0 +1,94 @@
...
+            ndim: See attribute docstring.
+        """
+        self.ndim = ndim
+
+    @abc.abstractmethod
+    def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
+        """Updates the data for the viewer.
+        
+        Args:
+            data: The new ndarray data. Its dimension should be self.dim.
```
BECATRUE: `self.ndim`!

iquip/apps/dataviewer.py
```diff
@@ -0,0 +1,94 @@
...
+    """Plot viewer for visualizing a 2D curve.
+    
+    Attributes:
+        plotItem: The PlotItem for showing the curve plot.
+        plotWidget: The PlotWidget which contains the plotItem.
+        curve: The PlotDataItem which represents the curve plot.
+    """
+
+    def __init__(self, **kwargs):
+        """
```
BECATRUE, line 80: It looks better to add "Extended", right?

BECATRUE (changes requested):

iquip/apps/dataviewer.py
kangz12345: I'm not sure 😄 actually it is overwritten, rather than overridden or extended.
Note that its arguments are different from its super's.

kangz12345:
@BECATRUE Please check this out!

BECATRUE (approved):
LGTM!

PR #131, opened 2023-09-13
Update the QIWIS version to 2.1.0

### LINKED ISSUE #127: Update the QIWIS version to 2.1.0
In according that the QIWIS version is updated to 2.1.0 in snu-quiqcl/qiwis#200, we will update the version IQUIP uses.
- As we can use the central position to show a frame, I will modify the position setting to show legibly.
- Make the ARTIQ proxy server settings, i.e. IP address and Port number global constants.
- Apply the background and icon images.

BECATRUE:
This closes #127.

### What I implemented
- Open a builder app with the center position.
- Add the proxy IP and PORT in the `config.json` and apply it.
- Apply the background and icon images.

### Result
image


config.json
```diff
@@ -27,5 +27,11 @@
             "show": true,
             "pos": "bottom"
         }
+    },
+    "constant": {
+        "proxy_ip": "172.22.22.10",
+        "proxy_port": "3390",
```
kangz12345: Since this is a public repo, please replace this with dummy values.
For example:
```
"proxy_ip": "xxx.xxx.xxx.xxx",
"proxy_port": "xxxxx",
```

kangz12345: BTW, why do you write the port number in `str`, instead of `int`?

kangz12345 (changes requested):

config.json
BECATRUE: There is no specific reason😅 Does it look better to set as `int`?

kangz12345: 😄 Yeah the port number is `int` in many cases.
Maybe later we might use some other ip/port, then it would use `int` for port numbers, so for consistency!

kangz12345 (approved):
LGTM!

PR #140, opened 2023-09-15
Improve unit test output

### LINKED ISSUE #126: Improve unit test output by buffering the stdout during the test
As @giwon2004 suggested, it would be better to buffer the stdout during the unit tests.
By using the [-b option](https://docs.python.org/3/library/unittest.html#cmdoption-unittest-b), we could achieve what we want.

kangz12345:
This closes #126.

This adds `-b` option to the unittest command, and also removes an unnecessary `print()` in `test_builder.py`.


BECATRUE (approved):
What a wonderful solution!

PR #142, opened 2023-09-15
Kangz12345/141/config template

### LINKED ISSUE #141: Only track config template and ignore config.json
Since the users will use different config files, so the current config.json file should be included in the .gitignore.
Instead, I want to rename the current config file to config_template.json.

kangz12345:
This closes #141.


BECATRUE (approved):
LGTM!

PR #146, opened 2023-09-15
Minimal `DataViewerApp` part 0: `SourceWidget`

### LINKED ISSUE #144: `DataViewerApp` with minimal features
#### 1. Source selection
- Realtime: "Sync" button to synchronize with the current artiq master. The RID information is shown.
- Remote: Type the desired RID in a spinbox to get the result data. The execution date is shown.
- Local: Will not be implemented in this issue.

#### 2. Data point
- Threshold spinbox (whose prefix is ">" which indicates that the threshold is exclusive).
- P1: Population of the state 1 based on the threshold.
- Average: Average photon count.
- Total: Total photon count.
P1/Average/Total are in a radio button group, so the selected value will be used for the main plot.
- Histogram plot (`BarGraphItem`) with the threshold vertical line.

image

#### 3. Main plot
- The main plot based on the selected value in "Data point" section.
- The plot type is switched based on the dimension of the data.
- If the user click a data point, its histogram is shown in "Data point" section.

### 4. Tools
- This will not be implemented in this issue.

See also: #143

## Subtasks
- #146 
- #151
- #156
- #160
- #164
- #176

kangz12345:
See #144 for details.

This PR implements `SourceWidget` which is the first part of the plan.
It provides an interface to select the data source type.
To the right of the radio buttons, an additional interface is provided.

Screen Shot 2023-09-15 at 18 15 29
Screen Shot 2023-09-15 at 18 15 41

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -195,7 +195,7 @@ class _RemotePart(QWidget):
         """Extended."""
         super().__init__(parent=parent)
         self.spinbox = QSpinBox(self)
-        self.spinbox.setRange(0, 999_999_999)
+        self.spinbox.setMaximum(2**31 - 1)
         self.spinbox.setButtonSymbols(QAbstractSpinBox.NoButtons)
         self.label = QLabel("Unknown", self)
         layout = QHBoxLayout(self)
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -173,7 +173,7 @@ class _RealtimePart(QWidget):
         button: Button for synchronizing with the current artiq master.
     """
 
-    def __init__(self, parent=None):
+    def __init__(self, parent: Optional[QWidget] = None):
         """Extended."""
         super().__init__(parent=parent)
         self.label = QLabel("No running experiment.", self)
@@ -191,7 +191,7 @@ class _RemotePart(QWidget):
         label: Label for showing the execution time of the experiment.
     """
 
-    def __init__(self, parent=None):
+    def __init__(self, parent: Optional[QWidget] = None):
         """Extended."""
         super().__init__(parent=parent)
         self.spinbox = QSpinBox(self)
@@ -220,7 +220,7 @@ class SourceWidget(QWidget):
         REALTIME = 0
         REMOTE = 1
 
-    def __init__(self, parent=None):
+    def __init__(self, parent: Optional[QWidget] = None):
         """Extended."""
         super().__init__(parent=parent)
         buttonGroupLayout = QVBoxLayout()
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -9,8 +9,8 @@ from typing import Sequence, Optional
 import numpy as np
 import pyqtgraph as pg
 from PyQt5.QtWidgets import (
-    QWidget, QLabel, QPushButton, QRadioButton, QButtonGroup,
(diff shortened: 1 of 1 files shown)


iquip/apps/dataviewer.py
```diff
@@ -156,3 +161,80 @@ def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
...
+    """Part widget for configuring realtime mode of the source widget.
+    
+    Attributes:
+        label: Label for showing information about the current experiment.
+          When it is synchronized with an experiment, it displays the RID
+          of the experiment. Otherwise, it shows "No running experiment.".
+        button: Button for synchronizing with the current artiq master.
+    """
+
+    def __init__(self, parent=None):
```
BECATRUE: Please set the type hint of `parent`, e.g., `parent: Optional[QWidget] = None`.

iquip/apps/dataviewer.py
```diff
@@ -156,3 +161,80 @@ def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
...
+    Attributes:
+        spinbox: Spinbox for RID input.
+        label: Label for showing the execution time of the experiment.
+    """
+
+    def __init__(self, parent=None):
+        """Extended."""
+        super().__init__(parent=parent)
+        self.spinbox = QSpinBox(self)
+        self.spinbox.setRange(0, 999_999_999)
```
BECATRUE: Setting the maximum value to just a random big number looks not good.

How about setting it to the maximum value of `int` type?
The max value of spin box is limitted to that of `int` type, so it looks more reasonable!

Additionally, the default minimum value is 0, and you don't need to set it.

iquip/apps/dataviewer.py
```diff
@@ -156,3 +161,80 @@ def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
...
+        self.spinbox.setButtonSymbols(QAbstractSpinBox.NoButtons)
```
BECATRUE, line 199: Why did you set this to `NoButtons`? Is it a temporary operation?

BECATRUE (changes requested):

iquip/apps/dataviewer.py
kangz12345: Okay then I will change it to `setMaximum(2**31 - 1)`.

iquip/apps/dataviewer.py
kangz12345: I don't think it requires +/- buttons. Note that we can use up/down keys to increase/decrease the current value of the spinbox.
Also, it is a temporary solution as well. Later I want to synchronize it to the result explorer app to select the RID from its list.

kangz12345:
I will rebase this branch on to `develop` after your review @BECATRUE.

kangz12345:
> I will rebase this branch on to `develop` after your review @BECATRUE.

I'm sorry, I will just rebase it now 😄

iquip/apps/dataviewer.py
BECATRUE: > Note that we can use up/down keys to increase/decrease the current value of the spinbox.

I couldn't think this way!

BECATRUE: > Also, it is a temporary solution as well. Later I want to synchronize it to the result explorer app to select the RID from its list.

👍

BECATRUE (approved):
LGTM!

PR #157, opened 2023-09-20
Single TTL controller: `TTLControllerWidget`

### LINKED ISSUE #153: Single TTL controller: `TTLControllerWidget`
For the first step of TTL controller project, I will develop a single TTL controller, which enables to set overriding and level.

`TTLControllerWidget` is a subclass of `QWidget`. It consists of the followings:
- ~~`TTLMonitorWidget` instance: To show the TTL status, but the actual function won't be implemented in this issue.~~
- Two labels for showing the (nick)name and the channel number.
- A button for setting the level when it is overriden.

See also: #152

BECATRUE:
This closes #153.

For test, please save the following code in the repository directory.
```python

from iquip.apps.monitor import TTLControllerWidget

qapp = QApplication([])
widget = TTLControllerWidget("Cooling Beam", 10)
widget.show()
qapp.exec_()
```

### DIFF
--- a/iquip/apps/monitor.py
+++ b/iquip/apps/monitor.py
@@ -2,12 +2,16 @@
 
 from typing import Optional
 
-from PyQt5.QtWidgets import QHBoxLayout, QLabel, QVBoxLayout, QWidget
+from PyQt5.QtWidgets import QHBoxLayout, QLabel, QPushButton, QVBoxLayout, QWidget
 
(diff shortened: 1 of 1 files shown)


BECATRUE:
As discussed with Jiyong, I removed the monitor part since we will not use this for a while.

BECATRUE:
Please recheck this! @kangz12345

kangz12345 (approved):
LGTM

PR #158, opened 2023-09-20
Frame with all TTL controllers: `TTLControllerFrame`

### LINKED ISSUE #154: Frame with all TTL controllers: `TTLControllerFrame`
Using `TTLControllerWidget`, I will develop a frame with all TTL controllers, which enables to set the override option of all TTLs.

`TTLControllerFrame` is a subclass of `QWidget`. It consists of the followings:
- Several `TTLControllerWidget` described in `config.json`.
- A button for setting the override option of all TTLs.

See also: #152

BECATRUE:
This closes #154.

For test, store the following code in the repository path.
```python

from iquip.apps.monitor import TTLControllerFrame

qapp = QApplication([])
widget = TTLControllerFrame({
    "AAA": 0, "BBB": 1, "CCC": 2, "DDD": 3,
    "EEE": 4, "FFF": 5, "GGG": 6, "HHH": 7
})
widget.show()
qapp.exec_()
```

### Result
_(updated)_

### DIFF
--- a/iquip/apps/monitor.py
+++ b/iquip/apps/monitor.py
@@ -63,6 +63,7 @@ class TTLControllerFrame(QWidget):
             self.ttlWidgets[name] = ttlWidget
             ttlWidgetLayout.addWidget(ttlWidget, row, column)
         self.overrideButton = QPushButton("Not Overridden", self)
+        self.overrideButton.setCheckable(True)
         # layout
         layout = QVBoxLayout(self)
         layout.addLayout(ttlWidgetLayout)
--- a/iquip/apps/monitor.py
+++ b/iquip/apps/monitor.py
@@ -2,7 +2,7 @@
 
 from typing import Dict, Optional
 
-from PyQt5.QtWidgets import QHBoxLayout, QLabel, QPushButton, QVBoxLayout, QWidget
+from PyQt5.QtWidgets import QGridLayout, QHBoxLayout, QLabel, QPushButton, QVBoxLayout, QWidget
 
 class TTLControllerWidget(QWidget):
     """Single TTL channel controller widget.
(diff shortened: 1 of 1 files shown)


iquip/apps/monitor.py
```diff
@@ -29,3 +35,77 @@ def __init__(self, name: str, channel: int, parent: Optional[QWidget] = None):
...
+    Attributes:
+        ttlWidgets: Dictionary with TTL controller widgets.
+          Each key is a TTL channel name, and its value is the corresponding TTLControllerWidget.
+        overrideButton: Button for setting the override.
+
+    Signals:
+        overrideChanged(override): Current override value is changed to override.
+    """
+
+    NUM_COLUMNS = 4
```
kangz12345: I think this should be an instance (constant) attribute, instead of class constant. Imagine we want several Frames and their number of columns different.

iquip/apps/monitor.py
```diff
@@ -29,3 +35,77 @@ def __init__(self, name: str, channel: int, parent: Optional[QWidget] = None):
...
+        self.ttlWidgets = {}
+        # widgets
+        ttlWidgetLayout = QGridLayout()
+        for idx, (name, channel) in enumerate(ttlInfo.items()):
+            ttlWidget = TTLControllerWidget(name, channel, self)
+            row = idx // TTLControllerFrame.NUM_COLUMNS
+            column = idx % TTLControllerFrame.NUM_COLUMNS
+            self.ttlWidgets[name] = ttlWidget
+            ttlWidgetLayout.addWidget(ttlWidget, row, column)
+        self.overrideButton = QPushButton("Not Overridden", self)
```
kangz12345: How about "Not Overriding"? I think "overriding" is more intuitive since it is overriding the ttl level.

kangz12345 (changes requested):

iquip/apps/monitor.py
BECATRUE: I see. Then I'll make sure to transfer the number of columns to an argument of constructor.

BECATRUE: Later we may change the number of columns actively, but I won't save it as an attribute in this PR because we only need it in the contructor right now.

kangz12345 (approved):
LGTM!

PR #160, opened 2023-09-20
Minimal `DataViewerApp` part 2: `MainPlotWidget`

kangz12345:
See #144 for details.

`MainPlotWidget` shows the data with a plot (1D) or an image (2D).
It provides a signal for each data point mouse click events.

You can test it using the following script:
```python
import iquip.apps.dataviewer as dv
import numpy as np

data_1d = np.random.random(10)
axes_1d = (dv.AxisInfo("Time", np.linspace(0, 9e-6, 10), "s"),)

data_2d = np.random.random((10, 10))
axes_2d = (
    dv.AxisInfo("Y", np.linspace(3e-3, 5e-3, 10), "m"),
    dv.AxisInfo("X", np.linspace(5e-3, 6e-3, 10), "m"),
)

qapp = QApplication([])
mainplots = []
for data, axes in ((data_1d, axes_1d), (data_2d, axes_2d)):
    mainplot = dv.MainPlotWidget()
    mainplots.append(mainplot)
    mainplot.setData(data, axes)
    mainplot.dataClicked.connect(lambda x: print(x))
    mainplot.show()
qapp.exec_()
```

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -86,9 +86,10 @@ class NDArrayViewer(metaclass=abc.ABCMeta):  # pylint: disable=too-few-public-me
         Args:
             sceneX: Scene x coordinate.
             sceneY: Scene y coordinate.
-            tolerance: Maximum Euclidean distance for aiming tolerance. If None,
-              the tolerance is infinity, i.e., the nearest data point is always
-              returned. Otherwise, it might return None if there is no data point
+            tolerance: Maximum Euclidean distance in the scene coordinate,
+              for aiming tolerance. If None, the tolerance is infinity,
+              i.e., the nearest data point is always returned.
+              Otherwise, it might return None if there is no data point
               within the tolerance.
         """
 
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -117,6 +117,22 @@ class CurvePlotViewer(NDArrayViewer):  # pylint: disable=too-few-public-methods
         self.plotItem.setLabel(axis="bottom", text=axis.name, units=axis.unit)
         self.curve.setData(axis.values, data)
 
+    def nearestDataPoint(
+        self, scenePos: pg.Point, tolerance: Optional[float] = None,
+    ) -> Optional[Tuple[int]]:
+        """Overridden."""
+        viewBox = self.plotItem.getViewBox()
+        viewPos = viewBox.mapSceneToView(scenePos)
+        viewRect = viewBox.viewRect()
+        sceneRect = viewBox.rect()
+        x, y = self.curve.getOriginalDataset()
+        dx, dy = viewPos.x() - x, viewPos.y() - y
+        rx, ry = sceneRect.width() / viewRect.width(), sceneRect.height() / viewRect.height()
+        distanceSquared = np.square(dx * rx) + np.square(dy * ry)
+        minIndex = np.argmin(distanceSquared)
+        if tolerance is None or distanceSquared[minIndex] <= np.square(tolerance):
+            return (minIndex,)
+        return None
 
 class HistogramViewer(NDArrayViewer):  # pylint: disable=too-few-public-methods
     """Histogram viewer showing a bar graph.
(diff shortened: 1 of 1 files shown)


kangz12345:
I think some commits went wrong... I fixed them but please review the code thoroughly!

iquip/apps/dataviewer.py
```diff
@@ -73,12 +77,25 @@ def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
             if size != len(info.values):
                 raise ValueError(f"Size mismatch in {info}: expected {size} values")
 
+    def nearestDataPoint(
+        self, scenePos: pg.Point, tolerance: Optional[float] = None,  # pylint: disable=unused-argument
```
BECATRUE: Please remove the trailing comma.

iquip/apps/dataviewer.py
```diff
@@ -165,6 +197,22 @@ def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
...
+        scenePos: pg.Point,
+        tolerance: Optional[float] = None,  # pylint: disable=unused-argument
+    ) -> Optional[Tuple[int, int]]:
+        """Overridden.
+        
+        ImageViewer does not use tolerance since it has a clear bounding box for
+          each data point.
+        """
+        dataPos = self.image.mapFromDevice(scenePos)
+        x, y = np.floor(dataPos.x()), np.floor(dataPos.y())
```
BECATRUE, line 215: Why did you use `np.floor()`, not `int()`?

iquip/apps/dataviewer.py
```diff
@@ -436,3 +484,77 @@ def _idToggledSlot(self, id_: int, checked: bool):
...
+        }
+        self.stack = QStackedWidget(self)
+        for plotType in MainPlotWidget.PlotType:
+            self.stack.addWidget(self.viewers[plotType].widget)
+        layout = QHBoxLayout(self)
+        layout.addWidget(self.stack)
+        # signal connection
+        for viewer in self.viewers.values():
+            viewer.plotItem.scene().sigMouseClicked.connect(
+                functools.partial(self._mouseClicked, viewer))
```
BECATRUE: The line breaking looks little strange. How about this?
```python
viewer.plotItem.scene().sigMouseClicked.connect(
    functools.partial(self._mouseClicked, viewer)
)
```

iquip/apps/dataviewer.py
```diff
@@ -73,12 +77,25 @@ def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
...
                 raise ValueError(f"Size mismatch in {info}: expected {size} values")
 
+    def nearestDataPoint(
+        self, scenePos: pg.Point, tolerance: Optional[float] = None,  # pylint: disable=unused-argument
+    ) -> Optional[Tuple[int, ...]]:
+        """Returns the index of the nearest data point.
+        
+        Args:
+            scenePos: Scene position coordinates.
+            tolerance: Maximum Euclidean distance in the scene coordinate,
```
BECATRUE, line 89: This is just a question! Is the unit of `tolerance` a pixel?

BECATRUE (changes requested):

iquip/apps/dataviewer.py
kangz12345: As I said before, trailing commas help reducing diffs in the future.
For example, if I want to add more arguments to `nearestDataPoint()`, I can just add another line for the new arguments, leaving the original line.
If I didn't use that trailing comma, I would have to put a comma at the end, so the line would be modified, making a diff on that line.

In fact, the most ideal code would be:
```python
# pylint: disable=unused-argument
def nearestDataPoint(
    self,
    scenePos: pg.Point,
    tolerance: Optional[float] = None,
) -> ...:
```

iquip/apps/dataviewer.py
kangz12345: I also used it at the first time, but noticed we forgot that `int()` truncates towards zero. That means -0.5 is mapped to 0, so they are misinterpreted as valid clicks.

iquip/apps/dataviewer.py
kangz12345: I think so..! "Scene position" seems to be in pixel.

iquip/apps/dataviewer.py
BECATRUE: Oh, I see. It looks more efficient! Then, please never mind😃

iquip/apps/dataviewer.py
BECATRUE: Wow, I forgot it perfectly! Thanks for noticing me.

BECATRUE (approved):
LGTM!

PR #176, opened 2023-09-25
Minimal `DataViewerApp` part 4: Final

### LINKED ISSUE #144: `DataViewerApp` with minimal features

kangz12345:
This closes #144.

I'm sorry for the large diffs. I will try to separate it if you feel too much.

Initial state:
Screen Shot 2023-09-26 at 11 58 10

After clicking "Sync":
Screen Shot 2023-09-26 at 14 15 26

## Dataset user guide
Users have to build their dataset in the following format:
```python
class Test(EnvExperiment):
    def prepare(self):
        self.set_dataset("name", [], broadcast=True)  # must not be in @kernel
        self.set_dataset("name.parameters", ["Param1", "Param2"], broadcast=True)
        self.set_dataset("name.units", ["Unit1", ""], broadcast=True)  # no unit for param2

    @kernel
    def run(self):
        cnt = 0  # photon count
        value1 = 1e-6  # param1 value
        value2 = 2  # param2 value
        self.append_to_dataset("name", [cnt, value1, value2])
```

Then in the GUI, enter the dataset name "name" in the line edit and click "Sync" button.
Then the axis combo boxes will work. After selecting the axes, click "Apply" button.

"Remote" mode is not implemented yet - will be implemented later (not in this PR)!

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -853,7 +853,7 @@ class DataViewerApp(qiwis.BaseApp):
         """Given the axis information, draws the main plot.
         
         Args:
-            axis: See SimpleScanDataPolicy.extract().
+            axis: See updateMainPlot().
         """
         self.axis = axis
         dataType = self.frame.dataPointWidget.dataType()
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -893,7 +893,7 @@ class DataViewerApp(qiwis.BaseApp):
         """Returns the data array at the given index.
         
         Args:
-            index: See selectDataPoint().
+            index: The index of the target data point, in the dataset array.
         """
         _, symbols = self.policy.symbolize(self.axis)
         data_indices = np.all(symbols.T == index, axis=1)
@@ -904,7 +904,7 @@ class DataViewerApp(qiwis.BaseApp):
         """Selects a data point at the given index.
         
         Args:
-            index: The index of the target data point, in the dataset array.
+            index: See dataPoint().
         """
         if self.policy is None:
             return
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -820,6 +820,7 @@ class DataViewerApp(qiwis.BaseApp):
         self.axis: Tuple[int, ...] = ()
         self.frame.syncRequested.connect(self.synchronize)
         self.frame.sourceWidget.axisApplied.connect(self.setAxis)
+        self.frame.mainPlotWidget.dataClicked.connect(self.selectDataPoint)
 
     @pyqtSlot()
     def synchronize(self):
@@ -858,6 +859,22 @@ class DataViewerApp(qiwis.BaseApp):
         dataType = self.frame.dataPointWidget.dataType()
         reduce = self._reduceFunction(dataType)
         self.frame.mainPlotWidget.setData(*self.policy.extract(axis, reduce))
+        self.selectDataPoint((0,) * len(axis))
+
+    def selectDataPoint(self, index: Tuple[int, ...]):
+        """Selects a data point at the given index.
+        
+        Args:
+            index: The index of the target data point, in the dataset array.
+        """
+        _, symbols = self.policy.symbolize(self.axis)
+        data_indices = np.all(symbols.T == index, axis=1)
+        data = self.policy.dataset[:, 0][data_indices].astype(int)
+        for dataType in DataPointWidget.DataType:
+            value = self._reduceFunction(dataType)(data)
+            self.frame.dataPointWidget.setValue(value, dataType)
+        bins, counts = np.unique(data, return_counts=True)
+        self.frame.dataPointWidget.setHistogramData(bins, counts)
 
     def frames(self) -> Tuple[DataViewerFrame]:
         """Overridden."""
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -878,6 +878,8 @@ class DataViewerApp(qiwis.BaseApp):
             axis: See SimpleScanDataPolicy.extract().
             dataType: Target data type.
         """
+        if self.policy is None:
+            return
         reduce = self._reduceFunction(dataType)
         data, axes = self.policy.extract(axis, reduce)
         self.frame.mainPlotWidget.setData(data, axes)
@@ -894,6 +896,8 @@ class DataViewerApp(qiwis.BaseApp):
         Args:
             index: The index of the target data point, in the dataset array.
         """
+        if self.policy is None:
+            return
         self.dataPointIndex = index
         _, symbols = self.policy.symbolize(self.axis)
         data_indices = np.all(symbols.T == index, axis=1)
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -321,6 +321,7 @@ class SourceWidget(QWidget):
         self.datasetEdit = QLineEdit(self)
         self.datasetEdit.setPlaceholderText("Dataset")
         self.axisBoxes = {axis: QComboBox(self) for axis in "XY"}
+        self.axisBoxes["Y"].setEnabled(False)
         self.axisApplyButton = QPushButton("Apply", self)
         datasetLayout = QHBoxLayout()
         datasetLayout.addWidget(self.datasetEdit)
@@ -368,8 +369,9 @@ class SourceWidget(QWidget):
                 previousText[axis] = combobox.currentText()
         items = [parameter if unit is None else f"{parameter} ({unit})"
                  for parameter, unit in zip(parameters, units)]
-        self.axisBoxes["X"].clear()
-        self.axisBoxes["X"].addItems(items)
+        for axis in "YX":
+            self.axisBoxes[axis].clear()
+            self.axisBoxes[axis].addItems(items)
         for axis, text in previousText.items():
             self.axisBoxes[axis].setCurrentText(text)
         self._handleApplyClicked()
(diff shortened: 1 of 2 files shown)


BECATRUE:
Could you attach a screenshot?

iquip/apps/dataviewer.py
```diff
@@ -1,4 +1,5 @@
 """App module for data viewers which displays result data using plot, etc."""
+# pylint: disable=too-many-lines
```
BECATRUE, line 1: Which is better to locate it before or after the module docstring?

iquip/apps/dataviewer.py
```diff
@@ -13,18 +14,33 @@
...
+    """Returns P1 given threshold and photon count array.
+    
+    Args:
+        threshold: If the photon count is strictly greater than threshold, it is
+          taken as 1 state.
+        array: The array of photon counts.
+    """
+    counts = np.bincount(array, minlength=threshold+1)
+    total = np.sum(counts)
+    p0 = np.sum(counts[:threshold+1])
```
BECATRUE: I think the following looks simpler, but is there any reason using `np.bincount()`?
```python
total = array.size
p1 = np.sum(array > threshold)
return p1 / total
```

iquip/apps/dataviewer.py
```diff
@@ -311,11 +344,73 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+        It updates the Y axis combobox status properly.
+
+        Args:
+            index: Currently selected combobox item index.
+        """
+        xBox, yBox = self.axisBoxes.values()
+        previousY = yBox.currentText()
+        yBox.clear()
+        for i in range(xBox.count()):
+            if i != index:
```
BECATRUE: How about checking the case where `i == index` and continuing that case?
It can prevent this loop from becoming deeper.

iquip/apps/dataviewer.py
```diff
@@ -311,11 +344,73 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+        """Called when the axis parameter apply button is clicked."""
+        axis = ()
+        xBox, yBox = self.axisBoxes.values()
+        x, y = xBox.currentIndex(), yBox.currentIndex()
+        if x >= 0:
+            axis = (x,)
+            if y >= 0:
+                if y >= x:
+                    y += 1
+                axis = (y, x)
```
BECATRUE: I understood this process, but it looks too complicated...
So how about disabling the Y axis which is the same as the current X axis, instead of removing it.

You can disable a specific item as follows.
```python
combo_box = QComboBox()

combo_box.addItem("Item 0")
combo_box.addItem("Item 1")
combo_box.addItem("Item 2")
combo_box.addItem("Item 3")
combo_box.addItem("Item 4")

combo_box.model().item(2).setEnabled(False)
```

iquip/apps/dataviewer.py
```diff
@@ -634,24 +733,213 @@ def datasetName(self) -> str:
...
+                                    params={"key": self.name},
+                                    timeout=10)
+            response.raise_for_status()
+        except requests.exceptions.RequestException:
+            logger.exception("Failed to fetch the dataset %s.", self.name)
+            return
+        dataset = np.array(response.json())
+        try:
+            response = requests.get(f"http://{self.ip}:{self.port}/dataset/master/",
+                                    params={"key": f"{self.name}.parameters"},
```
BECATRUE, line 782: Does this mean that a user should make a key, `{name}.parameters` in the dataset?

iquip/apps/dataviewer.py
```diff
@@ -634,24 +733,213 @@ def datasetName(self) -> str:
...
+        if self.policy is None:
+            return
+        self.dataPointIndex = index
+        _, symbols = self.policy.symbolize(self.axis)
+        data_indices = np.all(symbols.T == index, axis=1)
+        data = self.policy.dataset[:, 0][data_indices].astype(int)
+        for dataType in DataPointWidget.DataType:
+            value = self._reduceFunction(dataType)(data)
+            self.frame.dataPointWidget.setValue(value, dataType)
+        bins, counts = np.unique(data, return_counts=True)
```
BECATRUE, line 917: In the case where `data = [0, 0, 1, 1, 2, 4, 5, 10, 10]`, `bins` would be `[0, 1, 2, 4, 5, 10]`.
So the X axis of the histogram is no longer a continuous (non-negative) integer.

Did you intend it?

iquip/apps/dataviewer.py
```diff
@@ -634,24 +733,213 @@ def datasetName(self) -> str:
...
+        self.frame.dataPointWidget.setHistogramData(bins, counts)
+
+    @pyqtSlot(int)
+    def setThreshold(self, threshold: int):  # pylint: disable=unused-argument
+        """Given the threshold, updates the data point widget and the main plot.
+        
+        Args:
+            threshold: See DataPointWidget.setThreshold().
+        """
+        self.updateMainPlot(self.axis, self.frame.dataPointWidget.dataType())
```
BECATRUE: Isn't it necessary to update the main plot only if the current data type is `P1`?

iquip/apps/dataviewer.py
```diff
@@ -634,24 +733,213 @@ def datasetName(self) -> str:
...
+        self.selectDataPoint(self.dataPointIndex)
```
BECATRUE: Even if the threshold is changed, won't the data point plot be changed?

iquip/apps/dataviewer.py
```diff
@@ -634,24 +733,213 @@ def datasetName(self) -> str:
...
+        data_indices = np.all(symbols.T == index, axis=1)
+        data = self.policy.dataset[:, 0][data_indices].astype(int)
+        for dataType in DataPointWidget.DataType:
+            value = self._reduceFunction(dataType)(data)
+            self.frame.dataPointWidget.setValue(value, dataType)
+        bins, counts = np.unique(data, return_counts=True)
+        self.frame.dataPointWidget.setHistogramData(bins, counts)
+
+    @pyqtSlot(int)
+    def setThreshold(self, threshold: int):  # pylint: disable=unused-argument
```
BECATRUE: How about naming `threshold` with a leading underscope in order to bypass the pylint warning?

iquip/apps/dataviewer.py
```diff
@@ -634,24 +733,213 @@ def datasetName(self) -> str:
...
+        data, axes = self.policy.extract(axis, reduce)
+        self.frame.mainPlotWidget.setData(data, axes)
+        index = self.dataPointIndex
+        if data.ndim == len(index) and np.all(np.less(index, data.shape)):
+            self.selectDataPoint(index)
+        else:
+            self.selectDataPoint((0,) * data.ndim)
+
+    @pyqtSlot(tuple)
+    def selectDataPoint(self, index: Tuple[int, ...]):
```
BECATRUE, line 903: Why didn't you update the number of samples here?

I think you can update it as follows:
```python
self.frame.dataPointWidget.setNumberOfSamples(len(data))
```

BECATRUE (changes requested):
It wasn't easy to review, but the structure looks too wonderful😮

iquip/apps/dataviewer.py
kangz12345: Hmm... I think the pylint comment should go first 😄 I will swap the lines.

iquip/apps/dataviewer.py
kangz12345: Your solution looks way better! Thank you for the suggestion.

iquip/apps/dataviewer.py
kangz12345: What a good idea! I will apply that feature.

iquip/apps/dataviewer.py
kangz12345: Since I applied your suggestion (disabling item), this code is updated as well.

iquip/apps/dataviewer.py
kangz12345: Yes, for now users have to define such datasets and put information in it 😄

iquip/apps/dataviewer.py
kangz12345: Yes, the X axis data does not have to be continuous.

iquip/apps/dataviewer.py
kangz12345: You are right, I will update the code since it is inefficient.

iquip/apps/dataviewer.py
kangz12345: Do you mean it shouldn't be changed? To be honest I was too lazy to update the P1 spinbox, so I just updated it all.
I improved the code to update only the P1 value!

iquip/apps/dataviewer.py
kangz12345: Considering inheritance and overriding, I think we should not name the arguments like that (put an underscore to avoid pylint warning), because it will cause trouble when we call overridden methods with keyword arguments.
(Even if `DataViewerApp` might not be inherited, for consistency!)

For example:
```python
class A:
    def f(self, x, _y):
        """Uses x but not _y."""

class B(A):
    def f(self, _x, y):
        """Overridden, uses y but not _x."""

def test(obj: A):  # obj can be an instance of A or B
    obj.f(x=0, y=0)  # error, one can never call f() with keyword arguments
```

Anyway, I just removed the `threshold` argument since it is okay for slots to have less arguments than the connected signals.

iquip/apps/dataviewer.py
kangz12345: You are right! I totally forgot about that.

iquip/apps/dataviewer.py
BECATRUE: Oh, I understand right now! Please never mind😅

iquip/apps/dataviewer.py
BECATRUE: > Do you mean it shouldn't be changed?

Umm.. I thought even if we change the threshold spin box, the data point index wouldn't be changed, so the data point plot will.

But, as you said, we need to update the P1 value! I totally forgot it😅

iquip/apps/dataviewer.py
BECATRUE: Oh, I see!

BTW, your solution looks so cool!

iquip/apps/dataviewer.py
```diff
@@ -311,11 +341,74 @@ def __init__(self, parent: Optional[QWidget] = None):
...
+        Args:
+            See SimpleScanDataPolicy.
+        """
+        previousText: Dict[str, str] = {}
+        for axis, combobox in self.axisBoxes.items():
+            if combobox.currentIndex() >= 0:
+                previousText[axis] = combobox.currentText()
+        items = [parameter if unit is None else f"{parameter} ({unit})"
+                 for parameter, unit in zip(parameters, units)]
+        for axis in "YX":
```
BECATRUE, line 372: It is just a question! Why "YX", not "XY"?

kangz12345: 😄 because we have to update the Y item list before X is cleared, since clearing X would trigger `currentIndexChanged` signal!

BECATRUE (approved):
LGTM!

BTW, I think we need an info label about `index` in data point plot, in order to identify the current index.

PR #189, opened 2023-10-06
Set default index -1

### LINKED ISSUE #186: Select the most recent data point when the dataset is updated
When the dataset is updated in the dataviewer, it seems helpful to select the most recent data point automatically, especially for realtime mode (which is the only mode for now).

Support -1 index, and select the -1-th datapoint after every dataset update.

kangz12345:
This closes #186.
I think this is enough, but I will fix the bugs if any occurs :)


BECATRUE (approved):
Cool!

But, I wonder if it works well as you intended even when the dimension is two.

kangz12345:
You are right, that issue should be considered later 😄

BECATRUE:
Now, you can merge this!

PR #191, opened 2023-10-10
Scheduler: Fetch the actual scheduled queue and apply it in the view

### LINKED ISSUE #185: Scheduler: Fetch the actual scheduled queue and apply it in the view
This is a sub-issue of #181.

After implementing the basic model/view architecture In #182, I will implement the followings:
- Fetch the actual scheduled queue from the proxy server.
- Show it in view actively.

#### _ScheduleThread: `QThread` subclass for fetching the scheduled queue

It runs repeatly, in other words, as soon as a thread is finished, the another one is created and started.
Every time the thread is finished, the view is updated to the fetched scheduled queue.

BECATRUE:
This closes #185 and for details, please see the issue.

Now, it fetches the scheduled queue and shows it in real time.

### Screenshot
image

### DIFF
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -2,7 +2,7 @@
 
 import enum
 import logging
-from typing import Any, Callable, Iterable, Optional, Sequence, Tuple
+from typing import Any, Callable, Optional, Sequence, Tuple
 
 import requests
 from PyQt5.QtCore import (
@@ -38,7 +38,7 @@ class _ScheduleThread(QThread):
         updatedTime: Optional[float],
         ip: str,
         port: int,
-        callback: Callable[[bool, float, Iterable[SubmittedExperimentInfo]], None],
+        callback: Callable[[bool, float, Sequence[SubmittedExperimentInfo]], None],
         parent: Optional[QObject] = None
     ):  # pylint: disable=too-many-arguments
         """Extended.
@@ -130,7 +130,7 @@ class ScheduleModel(QAbstractTableModel):
         """Overridden.
 
         Returns:
-            DisplayRole: Column-th info field of row-th experiment in scheduleList.
+            DisplayRole: Column-th info field of row-th experiment in the schedule.
         """
         if not index.isValid() or role != Qt.DisplayRole:
             return QVariant()
@@ -206,7 +206,7 @@ class SchedulerApp(qiwis.BaseApp):
         self,
         isChanged: bool,
         updatedTime: float,
-        schedule: Iterable[SubmittedExperimentInfo]
+        schedule: Sequence[SubmittedExperimentInfo]
     ):
         """Updates schedulerFrame.scheduleModel using the given schedule.
         
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -1,8 +1,10 @@
 """App module for showing the scheduled queue for experiments."""
 
 import enum
-from typing import Any, Callable, Dict, Optional, Tuple
+import logging
+from typing import Any, Callable, Iterable, Optional, Tuple
 
+import requests
 from PyQt5.QtCore import (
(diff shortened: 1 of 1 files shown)


iquip/apps/scheduler.py
```diff
@@ -1,14 +1,95 @@
...
+    """
+
+    fetched = pyqtSignal(bool, float, list)
+
+    def __init__(
+        self,
+        ip: str,
+        port: int,
+        updatedTime: Optional[float],
+        callback: Callable[[bool, float, Iterable[SubmittedExperimentInfo]], None],
```
kangz12345: Since the third argument of `fetched` is `list`, you should require `callback` to accept `List` rather than `Iterable`.
You are imposing too strong condition (it is enough for `callback` to accept lists, not tuples, etc.).

Note that `Callable[[List], None]` _contains_ `Callable[[Iterable], None]`. Don't be confused!

iquip/apps/scheduler.py
```diff
@@ -109,13 +190,53 @@ class SchedulerApp(qiwis.BaseApp):
...
+    ):
+        """Updates schedulerFrame.scheduleModel using the given schedule.
+        
+        Args:
+            See _ScheduleThread signals section.
+        """
+        if isChanged:
+            model = self.schedulerFrame.scheduleModel
+            model.beginResetModel()
+            model.scheduleList = schedule
```
kangz12345: Since the `ScheduleModel.scheduleList` is a `list` based on its first definition, `schedule` type hint should be `List`.
Or, based on its usage, it should be `Sequence` as it must allow arbitrary indexing.

Anyway, I think it is not a good idea to assign `schedule` directly to `model.scheduleList` since it is a mutable data structure and it came from outside (given by an argument), which means there could be many other _modifiers_.

Of course `SubmittedExperimendInfo` itself is also mutable, but I think that is okay since we would barely modify them. On the other hand, lists are prone to be modified and we might make mistakes easily.

My suggestion is to clear the `model.scheduleList` first and then update it using e.g., `extend()`. You can always adapt your own way to resolve issues though!

kangz12345 (changes requested):

iquip/apps/scheduler.py
BECATRUE: I have a question! Why does the former contain the latter? I think `Iterable` contains `List`...

iquip/apps/scheduler.py
BECATRUE: Oh, it's a good point! Then, how about using just `copy()` or `deepcopy()`?

BECATRUE:
As @kangz12345 and I discussed, I made the schedule list in model private, and offered a method to set it.
Thus, we can handle the schedule data more safely.

iquip/apps/scheduler.py
```diff
@@ -26,31 +26,31 @@ class _ScheduleThread(QThread):
...
     fetched = pyqtSignal(bool, float, list)
 
     def __init__(
         self,
+        updatedTime: Optional[float],
         ip: str,
         port: int,
-        updatedTime: Optional[float],
-        callback: Callable[[bool, float, Iterable[SubmittedExperimentInfo]], None],
+        callback: Callable[[bool, float, Sequence[SubmittedExperimentInfo]], None],
```
kangz12345: Still, `Callable[[..., List[...]], ...]` is better since the `fetched` always gives a `list`.
You can then connect `Callable[[..., Sequence[...]], ...]` as its slot, just like you did now.

kangz12345 (changes requested):

iquip/apps/scheduler.py
BECATRUE: Then, is it better to modify `updateScheduleModel()` type hint as well?

kangz12345 (approved):
LGTM

PR #196, opened 2023-10-12
Highlight selected data point

### LINKED ISSUE #177: Highlight the currently selected data point in data viewer
When a data point is selected, currently it is hard to identify which data point is selected.
Therefore, I want to highlight it by drawing a horizontal line and a vertical line in the plot, at the data point.

When `DataViewerApp.selectDataPoint()` is called, adjust the position of the two highlight lines.

kangz12345:
This closes #177.

The feature is implemented only for 1D data.

Screen Shot 2023-10-12 at 21 39 11

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -715,6 +715,10 @@ class MainPlotWidget(QWidget):
                 functools.partial(self._mouseClicked, viewer),
             )
 
+    def viewer(self) -> NDArrayViewer:
+        """Returns the current viewer."""
+        return self.viewers[self.stack.currentIndex()]
+
     def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
         """Sets the data to plot.
 
@@ -980,6 +984,7 @@ class DataViewerApp(qiwis.BaseApp):
         """
         if self.policy is None:
             return
+        self.frame.mainPlotWidget.viewer().highlight(index)
         self.dataPointIndex = index
         data = self.dataPoint(index)
         for dataType in DataPointWidget.DataType:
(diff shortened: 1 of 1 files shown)


iquip/apps/dataviewer.py
```diff
@@ -127,6 +139,7 @@ def __init__(self, **kwargs):
         self.plotItem = pg.PlotItem(**kwargs)
         self.widget = pg.PlotWidget(plotItem=self.plotItem)
         self.curve = self.plotItem.plot()
+        self.lines: Optional[Tuple[pg.InfiniteLine, pg.InfiniteLine]] = None
```
BECATRUE, line 140: Please write the docstring in terms of this!

iquip/apps/dataviewer.py
```diff
@@ -110,6 +110,18 @@ def nearestDataPoint(
         """
         return None
 
+    def highlight(self, index: Optional[Tuple[int, ...]]):
+        """Highlights the data point at the given index.
+        
+        A vertical and a horizontal lines appear at the data point.
+        If there was already a highlight lines, they move to the new point, i.e.,
+          there are at most one highlighted data point at once.
```
BECATRUE: Umm.. I think this explanation should be moved into `CurvePlotViewer` because we may want to implement this differently according to subclasses of `NDArrayViewer` and now you overrode it in only `CurvePlotViewer`.

BECATRUE (changes requested):
Even if you change the selected dataset, the highlight-lines would still remain. Is it okay?

iquip/apps/dataviewer.py
kangz12345: It makes sense, it does not have to be two lines for the other viewers. I will move that docstring.

kangz12345:
> Even if you change the selected dataset, the highlight-lines would still remain. Is it okay?

Yes, because new data point will be selected when the dataset is changed.
(I think there is already a bug when the new dataset is empty though)

BECATRUE (approved):
Cool!

PR #202, opened 2023-10-16
Fetching schedule stops permanently after an error occurs

### LINKED ISSUE #200: Fetching schedule stops permanently after an error occurs
### Describe the bug
After a request error such as a connection error occurs, fetching procedure stops permanently and we cannot restart it.
Please refer to the [code](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/scheduler.py#L79).

Just merge this case into the timeout case.

BECATRUE:
This closes #200.

When a request error such as a connection error occurs, it will restart fetching procedure.
Also, I set the default `updatedTime` to -1 instead of `None` as it is a signal parameter and PyQt doesn't support to emit `None`.

### DIFF
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -73,11 +73,10 @@ class _ScheduleThread(QThread):
                                     timeout=10)
             response.raise_for_status()
             response = response.json()
-        except requests.exceptions.Timeout:
+        except requests.exceptions.RequestException as e:
+            if not isinstance(e, requests.exceptions.Timeout):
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
LGTM!
However, I think we should later implement another way to trigger the updating loop (e.g., restart updating button).
Consider the case where some problems occurred during weekend so it retries over and over again for tens of hours.
For now, this patch seems good enough.

BECATRUE:
I see! As you said, I will handle this in more details.

PR #204, opened 2023-10-16
Convert negative index to positive in `dataPoint()`

### LINKED ISSUE #203: `selectDataPoint` fails when a dataset is loaded
### Describe the bug
When a new dataset is loaded, an exception occurs.

### To Reproduce
1. Type a dataset name.
2. Turn on polling for a while.
3. Set an appropriate axis.
4. Click "Apply" button.

As I debugged, [DataViewerApp.dataPoint()](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/dataviewer.py#L984) doesn't support `index` with `(-1,)`.

Once loaded, `DataViewerApp.selectDataPoint()` is called in [here](DataViewerApp), and it calls `dataPoint()` in [here](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/dataviewer.py#L998) with `index = (-1,)`.
And it causes an error in `data_indices = np.all(symbols.T == index, axis=1)` since `symbols.T` never has `(-1,)`.

kangz12345:
This resolves #203.

Applied the method suggested by @BECATRUE.

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -980,8 +980,9 @@ class DataViewerApp(qiwis.BaseApp):
         Args:
             index: The index of the target data point, in the dataset array.
         """
-        _, symbols = self.policy.symbolize(self.axis)
-        data_indices = np.all(symbols.T == index, axis=1)
+        parameters, symbols = self.policy.symbolize(self.axis)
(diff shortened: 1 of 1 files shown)


BECATRUE (approved):
LGTM!

PR #208, opened 2023-10-20
Modification based dataset realtime update

### LINKED ISSUE #207: Modification-based realtime update in data viewer
After snu-quiqcl/artiq-proxy#87 is fully implemented, the data viewer can take only modifications in each polling, instead of the whole dataset repeatedly.
Then longer datasets will be able to be loaded without any crash.

Renew the GUI components:
- dataset selection combobox: show available dataset names
- sync start/stop button: remove the current polling UI and add synchronization UI

Synchronization procedure:
1. When starting sync, get the current whole dataset with the timestamp of that response once.
2. Now start long-polling loop:
3. Get the modifications (if not timeout) since the timestamp and apply them to the current plot, and update the timestamp variable with the received one.
4. Repeat the step 3 until stopping sync.
5. If sync stopped or an unexpected error occurred (not timeout), stop the polling loop and do not update the plot.

kangz12345:
I apologize for the large diffs...
It was an urgent patch so I didn't break it to several PRs.

To summarize this PR:
1. Change polling strategy: periodic polling -> long polling
2. Change dataset update strategy: fetch whole dataset every time -> only modifications since the last fetch
3. Change threading strategy: Spawn a new thread for every fetch -> Spawn a thread for each sync (spawn once start sync, destroy after stop sync) and run a long polling loop
4. Synchronize fetching thread and GUI thread with a `QMutex` and a `QWaitCondition`
5. _NOT_ changed the data structure or plot strategy: it still updates the plot for the whole dataset when the dataset is updated (even by modifications)

More description for the thread synchronization:
- Without any synchronization, the fetching thread will report new modifications continuously while the GUI suffers from organizing and plotting the dataset.
- As the dataset size grows, the GUI latency becomes longer and the `modified` signals start to be queued.
- Then the GUI event queue will be full of `modified` signals, so the GUI might freeze or glitches may appear.
- Therefore, using a mutex and wait condition, the fetcher thread should wait for the GUI thread to update the plot before fetching the next modifications.

As I tested, the most of the latency was because of the large data transfer.
Therefore, I didn't change the dataset organizing and plotting policy.

This branch is already tested and used in the experiment PC.
It works well without any glitch up to 200,000 data points (tested so far, will work for even more!) (the previous version crashed for 20,000 data points).

This branch works after snu-quiqcl/artiq-proxy#93 is done (already done in a not-merged branch).

This closes #207.

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -858,8 +858,8 @@ class _DatasetFetcherThread(QThread):
             return default
         return response.json()
 
-    def run(self):
-        """Overridden."""
+    def _get_dataset(self):
+        """Fetches the target dataset and emits the signal."""
         rawDataset = self._get("dataset/master/", {"key": self.name})
         if rawDataset is None:
             return
@@ -873,6 +873,10 @@ class _DatasetFetcherThread(QThread):
         else:
             units = [unit if unit else None for unit in rawUnits]
         self.fetched.emit(dataset, parameters, units)
+
+    def run(self):
+        """Overridden."""
+        self._get_dataset()
 
 
 class DataViewerApp(qiwis.BaseApp):
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -898,6 +898,23 @@ class _DatasetFetcherThread(QThread):
         if timestamp < 0:
             self.stopped.emit("Failed to get dataset.")
             return
+        url = "dataset/master/modifications"
+        params = {"key": self.name, "timestamp": timestamp, "timeout": 10}
+        while self._running:
+            response = self._get(url, params, timeout=12)
+            if response is None:
+                self.stopped.emit("Failed to get modifications.")
+                return
+            timestamp, modifications = response
+            if timestamp < 0:
+                timestamp = self._initialize()
+                if timestamp < 0:
+                    self.stopped.emit("Dataset is deleted.")
+                    return
+            elif modifications:
+                self.modified.emit(modifications)
+            params["timestamp"] = timestamp
+        self.stopped.emit("Stopped synchronizing.")
 
 
 class DataViewerApp(qiwis.BaseApp):
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -931,8 +931,6 @@ class DataViewerApp(qiwis.BaseApp):
         self.frame = DataViewerFrame()
         self.thread: Optional[_DatasetFetcherThread] = None
         self.policy: Optional[SimpleScanDataPolicy] = None
-        self._data: Optional[np.ndarray] = None
-        self._axes: Tuple[AxisInfo, ...] = ()
         self.axis: Tuple[int, ...] = ()
         self.dataPointIndex: Tuple[int, ...] = ()
         self.frame.syncRequested.connect(self.synchronize)
@@ -1026,7 +1024,6 @@ class DataViewerApp(qiwis.BaseApp):
             return
         reduce = self._reduceFunction(dataType)
         data, axes = self.policy.extract(axis, reduce)
-        self._data, self._axes = data, axes
         self.frame.mainPlotWidget.setData(data, axes)
         index = self.dataPointIndex
         if data.ndim == len(index) and np.all(np.less(index, data.shape)):
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -794,7 +794,6 @@ class DataViewerFrame(QSplitter):
         self.addWidget(leftWidget)
         self.addWidget(mainPlotBox)
         self.addWidget(toolBox)
-        realtimePart = self.sourceWidget.stack.widget(SourceWidget.ButtonId.REALTIME)
         # signal connection
         remotePart = self.sourceWidget.stack.widget(SourceWidget.ButtonId.REMOTE)
         remotePart.ridEditingFinished.connect(self.dataRequested)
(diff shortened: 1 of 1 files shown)


iquip/apps/dataviewer.py
```diff
@@ -820,69 +810,127 @@ class _DatasetFetcherThread(QThread):
...
+    def _initialize(self) -> float:
+        """Fetches the target dataset to initialize the local dataset.
+        
+        Returns:
+            The received timestamp or -1 if it failed.
+        """
+        response = self._get("dataset/master/", {"key": self.name})
+        if response is None or response[0] < 0:
+            return -1
+        timestamp, rawDataset = response
```
BECATRUE, line 888: How about moving this above the previous _if_ statement, and replacing `response[0]` with `timestamp`?

iquip/apps/dataviewer.py
```diff
@@ -820,69 +810,127 @@ class _DatasetFetcherThread(QThread):
...
+        dataset = np.array(rawDataset)
+        numberOfParameters = dataset.shape[1] if dataset.ndim > 1 else 0
```
BECATRUE, line 890: It is just a question, but if `dataset.ndim = 0`, what happens in the data viewer?
I can't imagine this case🥲

BECATRUE: Also, which is better between `numParams` and the origianl one?

iquip/apps/dataviewer.py
```diff
@@ -820,69 +810,127 @@ class _DatasetFetcherThread(QThread):
...
+            {"key": f"{self.name}.parameters"},
+            (0, list(map(str, range(numberOfParameters)))),
+        )
+        rawUnits = self._get("dataset/master/", {"key": f"{self.name}.units"})
+        if rawUnits is None:
+            units = [None] * (numberOfParameters)
         else:
-            units = [unit if unit else None for unit in response.json()]
-        self.fetched.emit(dataset, parameters, units)
+            units = [unit if unit else None for unit in rawUnits[1]]
```
BECATRUE, line 900: Is it different from `units = rawUnits[1]`?

iquip/apps/dataviewer.py
```diff
@@ -820,69 +810,127 @@ class _DatasetFetcherThread(QThread):
...
+    def run(self):
+        """Overridden."""
+        timestamp = self._initialize()
+        if timestamp < 0:
+            self.stopped.emit("Failed to get dataset.")
+            return
+        url = "dataset/master/modification"
+        params = {"key": self.name, "timestamp": timestamp, "timeout": 10}
+        while self._running:
+            response = self._get(url, params, timeout=12)
```
BECATRUE, line 917: Is it sufficient to give only 2 seconds for transmission?

iquip/apps/dataviewer.py
```diff
@@ -820,69 +810,127 @@ class _DatasetFetcherThread(QThread):
...
+            if response is None:
+                self.stopped.emit("Failed to get modifications.")
+                return
+            timestamp, modifications = response
+            if timestamp < 0:
+                timestamp = self._initialize()
+                if timestamp < 0:
```
BECATRUE, line 924: Is it right that I understood?

First, if the dataset was removed, try to get it again.
Then, if there is the new dataset, run without error.
Otherwise, i.e. there is no dataset still, emit `stopped` signal.

iquip/apps/dataviewer.py
```diff
@@ -936,6 +1007,29 @@ def setDataset(
...
+            return
+        appended = np.vstack(tuple(m["x"] for m in modifications if m["action"] == "append"))
+        if self.policy.dataset.size == 0:
+            self.policy.dataset = appended
+        else:
+            self.policy.dataset = np.concatenate((self.policy.dataset, appended))
+        if not self.axis:
+            return
+        self.updateMainPlot(self.axis, self.frame.dataPointWidget.dataType())
+        self.thread.mutex.lock()
```
BECATRUE, line 1029: Why is it necessary? Is it already locked?

BECATRUE (changes requested):

iquip/apps/dataviewer.py
kangz12345: Then I have to separate `is None` check.

BECATRUE: Oh, I see! Then, it's okay to remain it as it is.

iquip/apps/dataviewer.py
kangz12345: When `ndim=0`, maybe the data policy can handle that case, but the viewer will not be able to.
It would crash I guess, and it is hard to define how it should react to such a case.
For now, I have no plans to support that case.

kangz12345: > Also, which is better between `numParams` and the origianl one?

That's a good and popular abbreviation, but I used the full version since I already used `parameters`.

iquip/apps/dataviewer.py
kangz12345: Since artiq only supports single-type lists, the unit list only has strings.
So no-unit is `""` instead of `None`.

iquip/apps/dataviewer.py
kangz12345: I'm not sure.. we have to measure the time to find the sufficient time.
However, it is okay to ignore the response in our protocol.

iquip/apps/dataviewer.py
kangz12345: Also, since the current version has a problem, this might be updated soon.
The problem is that a user should wait for up to 12 seconds after clicking the stop sync button.

iquip/apps/dataviewer.py
kangz12345: Yes, that's exactly how the `last_deleted` timestamp works.
Datasets are deleted when it is explicitly deleted or overwritten.

iquip/apps/dataviewer.py
kangz12345: This will prevent the wait condition from notifying before the fetcher thread start waiting.

kangz12345:
There comes a little bit of change soon, because of a bug found last week. It is resolved in the lab PC, so I will apply it to this branch soon.

iquip/apps/dataviewer.py
BECATRUE: Oh, I see! Your implementation is correct.

BECATRUE (approved):
LGTM

kangz12345:
> There comes a little bit of change soon, because of a bug found last week. It is resolved in the lab PC, so I will apply it to this branch soon.

I will create another PR for the updates.

PR #212, opened 2023-10-23
Safe fetcher thread

### LINKED ISSUE #209: Safe control of dataset synchronization thread
The current implementation of #208 does not protect the dataset synchronization thread (`_DatasetFetcherThread`) spawning.
In other words, since the user can click the Sync ON/OFF button rapidly, it can spawn multiple fetcher threads or induce some illegal situations.
Therefore, I want to protect the UI to ensure that the thread can be exist at most 1 at once.

When the button is toggled, it immediately disables the button.
If the button is now "ON", spawn a thread and enable the button back.
If the button is now "OFF", stop the thread and _DO NOT_ enable the button.
After the thread actually finishes, the `finished` signal will trigger to enable the button back.

Also, more logging will be added.

kangz12345:
This closes #209.

This also resolves some issues:
1. Missing a trailing slash "/" in "dataset/master/modification/". This induces a redirection for each request which slightly decrease the performance.
2. Deadlock occurred when there have been data modifications and `self.axis` is empty.

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -986,7 +986,12 @@ class DataViewerApp(qiwis.BaseApp):
         realtimePart: _RealtimePart = self.frame.sourceWidget.stack.widget(
             SourceWidget.ButtonId.REALTIME
         )
-        realtimePart.label.setText("Start synchronizing.")
+        if self.thread is not None and self.thread.isRunning():
+            logger.error("Tried to start dataset fetcher thread which is already running.")
+            realtimePart.setStatus(message="Stopping the running thread...")
+            self.thread.stop()
+            return
+        realtimePart.setStatus(message="Start synchronizing.")
         self.thread = _DatasetFetcherThread(
             self.frame.datasetName(),
             self.constants.proxy_ip,  # pylint: disable=no-member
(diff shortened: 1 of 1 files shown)


iquip/apps/dataviewer.py
```diff
@@ -296,19 +297,28 @@ def __init__(self, parent: Optional[QWidget] = None):
         layout.addWidget(self.label)
         # signal connection
         self.button.toggled.connect(self._buttonToggled)
+        self.button.clicked.connect(functools.partial(self.button.setEnabled, False))
```
BECATRUE, line 300: `self.button.clicked.connect(lambda: self.button.setEnabled(False))` may be another solution!

iquip/apps/dataviewer.py
```diff
@@ -967,16 +977,23 @@ def _toggleSync(self, checked: bool):
         """
         if checked:
             self.synchronize()
-        elif self.thread is not None:
+            return
+        try:
             self.thread.stop()
```
BECATRUE, line 982: When can a `RuntimeError` be occurred?

BECATRUE (changes requested):

iquip/apps/dataviewer.py
kangz12345: Although I prefer `partial()` to `lambda`, in this case `lambda` might be a better solution since it strictly and explicitly show that it takes no argument, and `clicked` signal has 1 boolean argument.
`partial()` makes it obscure that if the boolean argument is consumed or not (of course we know that `self.button.setEnabled()` takes a single argument and it is given by `False` already, so the partial function will not take any more arguments).
However, I want to use `partial()` when I can ;)

iquip/apps/dataviewer.py
kangz12345: Somehow if the `thread` is deleted by `deleteLater()` and then `thread.stop()` is called.

BECATRUE (approved):
LGTM!

PR #214, opened 2023-10-23
Improve thread signals and slots

### LINKED ISSUE #213: Delete finished threads and remove callbacks from arguments
Currently `iquip` creates a lot of threads and do not manage them at all.
Moreover, we are passing the callback slots as arguments of thread constructors.
I think this restricts the signal-slot connection of threads, e.g., number of signals and slots.
Therefore, I want to make the threads more stable in terms of signal-slots.

- Connect `deleteLater()` to the `finished` signal.
- Remove callbacks from arguments and connect them outside the thread.

### Additional context
We may consider using [`QThreadPool`](https://doc.qt.io/qt-5/qthreadpool.html) but it uses [`QRunnable`](https://doc.qt.io/qt-5/qrunnable.html) instead of [`QThread`](https://doc.qt.io/qt-5/qthread.html) which is too abstract and not a `QObject` (so it does not have signals).

kangz12345:
This closes #213.

Since this has many changes (not just additions), it must be tested and reviewed thoroughly.
I tested this on the lab PC.

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -516,6 +516,7 @@ class BuilderApp(qiwis.BaseApp):
             self.onReloaded,
             self
         )
+        self.experimentInfoThread.finished.connect(self.experimentInfoThread.deleteLater)
         self.experimentInfoThread.start()
 
     def onReloaded(
@@ -576,6 +577,7 @@ class BuilderApp(qiwis.BaseApp):
             self.onSubmitted,
             self
         )
+        self.experimentSubmitThread.finished.connect(self.experimentSubmitThread.deleteLater)
         self.experimentSubmitThread.start()
 
     def onSubmitted(self, rid: int):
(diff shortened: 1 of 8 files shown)


iquip/apps/builder.py
```diff
@@ -356,22 +356,19 @@ def __init__(
         schedOpts: Dict[str, Any],
         ip: str,
         port: int,
-        callback: Callable[[int], None],
         parent: Optional[QObject] = None
     ):  # pylint: disable=too-many-arguments
         """Extended.
         
         Args:
             experimentPath, experimentArgs, schedOpts, ip, port: See the attributes section.
```
BECATRUE: Now, it is okay to write only "See the attributes section.".

iquip/apps/explorer.py
```diff
@@ -161,6 +159,7 @@ def lazyLoadFile(self, experimentFileItem: QTreeWidgetItem):
             self._addFile,
             self
         )
+        self.fileFinderThread.finished.connect(self.fileFinderThread.deleteLater)
```
BECATRUE, line 162: Please update `_FileFinderThread` call, and connect `fetched` signal to `self._addFile()`.

BECATRUE (changes requested):

kangz12345:
I tested the branch on the lab PC.

BECATRUE (approved):
LGTM!

PR #217, opened 2023-11-14
Update fetching schedule to suit ARTIQ-PROXY

### LINKED ISSUE #216: Update fetching schedule to suit ARTIQ-PROXY
The mechanism offering schedules was updated in snu-quiqcl/artiq-proxy#102, hence we need to suit this protocol.

This implementation will be similar to #208, but much simpler.

BECATRUE:
This closes #216.

Now, the updated scheduler app works well to communicate with the proxy server.

The implementation is similar to that of dataviewer app, but much simpler.
The schedule is almost shorter than datasets, so I didn't lock the GUI even if the previously fetched schedule is being drawn.
Also, if fetching schedules is stopped abnormally, we cannot restart it now.

If necessary, I will handle it when using web socket in snu-quiqcl/artiq-proxy#105.

### DIFF
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -1,4 +1,4 @@
-"""App module for showing the scheduled queue for experiments."""
+"""App module for showing the schedule for experiments."""
 
 import enum
 import functools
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
LGTM

PR #220, opened 2023-11-16
Enable to restart fetching schedules

### LINKED ISSUE #218: Enable to restart fetching schedules
Now, if fetching schedules is paused due to an abnormal cause, it CANNOT be restarted and we should restart IQUIP entirely.
So I will make a button to restart it only if it is paused.

Similar to [dataviewer](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/dataviewer.py#L1006), I will connect the `finished` signal of fetching thread to a function that enable the button.
Once the button is clicked, it becomes disabled immediately and restarts fetching thread.

BECATRUE:
This closes #218.

At the bottom of GUI, the restart button is added.

image

### DIFF
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -213,7 +213,7 @@ class SchedulerFrame(QWidget):
     Attributes:
         scheduleView: The table view for showing the schedule.
         scheduleModel: The model for handling the schedule.
-        button: The button for restarting fetching schedules. When the button is clicked,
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
LGTM.
Please merge this if you tested it.

BECATRUE:
I checked it works well when a timeout exception occurs since it requests to the wrong proxy url.

PR #224, opened 2023-11-30
Select a class name when opening a builder

### LINKED ISSUE #223: Select a class name when opening a builder
As I said in snu-quiqcl/artiq-proxy#102, we need to support to designate a specific class name.

For this, I will add a `QInputDialog` to select a class name just after the open button in explorer app is cliecked.

BECATRUE:
This closes #223.

I updated the followings:
- `thread.py`: `ExperimentInfoThread` to emit with all experiments class info of the given experiment path.
- `apps/explorer.py`: After the experiment info is fetched, show a `QInputDialog`.
- `apps/builder.py`: Similar to `explorer.py`.
- Unittests for explorer and builder app.

Now, once we clicked the open button in explorer app, a `QInputDialog` appears.

image

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -341,6 +341,7 @@ class _ExperimentSubmitThread(QThread):
     
     Attributes:
         experimentPath: The path of the experiment file.
+        experimentClsName: The class name of the experiment.
         experimentArgs: The arguments of the experiment.
         schedOpts: The scheduler options; pipeline, priority, and timed.
         ip: The proxy server IP address.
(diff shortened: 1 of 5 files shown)


iquip/apps/builder.py
```diff
@@ -377,11 +380,11 @@ def run(self):
 
         Whenever the experiment is submitted well regardless of whether it runs successfully or not,
         the server returns the run identifier.
-        After submitted, the submitted signal is emitted.
```
kangz12345, line 380: Why is this line removed?

BECATRUE: I think the description of `submit` signal in the class docstring is enough.
`submitted(rid): The experiment is submitted. The rid is a run identifier.`

In addition, we didn't write signal descriptions in `run()` in dataviwer like [this](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/dataviewer.py#L893).

kangz12345 (approved):
LGTM

PR #226, opened 2023-12-01
Fix y-axis range of main plot in dataviewer from 0 to 1 when P1 data type

### LINKED ISSUE #225: Fix y-axis range of main plot in dataviewer from 0 to 1 when P1 data type
According to requests by an experimenter, I will fix the y-axis range of main plot in dataviewer app from 0 to 1, when the current data type is P1.

Pass the data type to `MainPlotWidget.setData()`.

BECATRUE:
This closes #225.

Now, once the P1 data type button is clicked, the y-axis range is set from 0 to 1.

image

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -691,6 +691,7 @@ class DataPointWidget(QWidget):
             checked: Whether the button is now checked or not.
         """
         if checked:
+            self.buttonGroup.button(id_).setChecked(True)
             self.dataTypeChanged.emit(DataPointWidget.DataType(id_))
 
 
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -745,7 +745,9 @@ class MainPlotWidget(QWidget):
         """Returns the current viewer."""
         return self.viewers[self.stack.currentIndex()]
 
-    def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
+    def setData(
+        self, data: np.ndarray, axes: Sequence[AxisInfo], dataType: DataPointWidget.DataType
+    ):
         """Sets the data to plot.
 
         If the dimension of data is 1, CURVE plot will be shown. If it is 2,
@@ -753,6 +755,7 @@ class MainPlotWidget(QWidget):
         
         Args:
             data, axes: See NDArrayViewer.setData().
+            dataType: See DataPointWidget.DataType.
         """
         if data.ndim == 1:
             plotType = MainPlotWidget.PlotType.CURVE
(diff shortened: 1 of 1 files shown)


iquip/apps/dataviewer.py
```diff
@@ -691,6 +691,7 @@ def _idToggledSlot(self, id_: int, checked: bool):
             checked: Whether the button is now checked or not.
         """
         if checked:
+            self.buttonGroup.button(id_).setChecked(True)
```
kangz12345: What is this for?

iquip/apps/dataviewer.py
```diff
@@ -1157,7 +1163,7 @@ def updateMainPlot(self, axis: Sequence[int], dataType: DataPointWidget.DataType
             return
         reduce = self._reduceFunction(dataType)
         data, axes = self.policy.extract(axis, reduce)
-        self.frame.mainPlotWidget.setData(data, axes)
+        self.frame.mainPlotWidget.setData(data, axes, dataType)
```
kangz12345: I think we should `setYRange()` here, i.e., `self.frame.mainPlotWidget.viewer().plotItem.setYRange(0, 1)`,
since the `MainPlotWidget` is independent to the `DataType`.
What do you think of this?

kangz12345 (changes requested):

iquip/apps/dataviewer.py
BECATRUE: I thought I needed this when I was testing before, but when I looked back, I don't think so.
Please ignore it😅

iquip/apps/dataviewer.py
BECATRUE: I see! I updated it as you suggested.

iquip/apps/dataviewer.py
```diff
@@ -762,11 +762,11 @@ def setData(self, data: np.ndarray, axes: Sequence[AxisInfo]):
             logger.error("MainPlotWidget does not support %d-dim data", data.ndim)
             return
         self.viewers[plotType].setData(data, axes)
-        self.stack.setCurrentIndex(plotType)
         if self.autoRangeBox.isChecked():
             plotItem = self.viewers[plotType].plotItem
             bounds = plotItem.getViewBox().childrenBoundingRect(items=None)
             plotItem.setXRange(bounds.left(), bounds.right())
+        self.stack.setCurrentIndex(plotType)
```
kangz12345, line 769: I think this line is just moved from line 765.
If it is not intended, please move it back to line 765, since it makes an unnecessary diff.

kangz12345 (changes requested):

iquip/apps/dataviewer.py
BECATRUE: As I think, it looks better to show the plot after all modifications are done.
What do you think of it?

kangz12345: Oh I agree with that! Let's go with your decision.

kangz12345 (approved):

PR #233, opened 2024-01-03
Aijuh/232/range scan

### LINKED ISSUE #232: Add scan widget only for RangeScan
Adds RangeScan widget for display.

Make new class for RangeScan layout. This layout will be called by _ScanEntry class using scannable argument info. 
temp2
Use spin box for displaying number and responding to keyboard or mouse input.
Add randomize button to decide whether or not to shuffle scannable sequence.

Aijuh:
This closes #232.

Adds scan widget that shows Rangescan variable.
Detailed information about scanning variable is [here](https://github.com/snu-quiqcl/artiq-proxy/discussions/111).

Below is the example image of RangeScan widget.
image

In terms of design, I used Gridlayout for RangeScan.
Below is the original scan widget at branch [AIJUH/229/scan-widget](https://github.com/snu-quiqcl/iquip/blob/AIJUH/229/scan-widget/iquip/apps/builder.py).
image


BECATRUE:
First, please check Pylint errors.

Aijuh:
Pylint error fixed by adding following lines.
```
# pylint: disable=too-many-instance-attributes
# pylint: disable=too-many-statements
```

BECATRUE:
I'm sorry for late check..
And there's too much content, so could you break it down into smaller pieces and reduce it to about 100 lines?

BECATRUE:
Also, in `RangeScan`, there is a `seed` parameter.
Did you intend not to implement this?

Aijuh:
In terms of seed variable, it seems that original artiq code did not implement this. 
How about adding new widget to select seed value in iquip?
If an user enters value at seed widget, use entered seed value. Otherwise, just use default seed value using timing.

BECATRUE:
I agree with you.
Even if the ARTIQ dashboard doesn't support the seed variable, this feature is supported in ARTIQ code, thus supporting it looks better.
Before implementing, however, you need to check how to send the seed variable to artiq-master.

PR #235, opened 2024-01-04
Fetch schedule using websocket

### LINKED ISSUE #234: Fetch schedule using web socket
According to snu-quiqcl/artiq-proxy#113, we need to update the mechanism fetching the schedule.

The implementation may be similar to the example code in snu-quiqcl/artiq-proxy#114.

BECATRUE:
Now, the scheduler app fetches schedule on real-time through a web socket.

~~Also, to show arguments correctly in the last column, I modified `ScheduleModel.data()`.~~
(I will handle this in #240.)

This closes #234.

### DIFF
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -172,7 +172,7 @@ class ScheduleModel(QAbstractTableModel):
             return QVariant()
         row, column = index.row(), index.column()
         infoField = ScheduleModel.InfoFieldId(column).name.lower()
-        return getattr(self._schedule[row], infoField)
+        return str(getattr(self._schedule[row], infoField))
 
     def headerData(self, section: int, orientation: Qt.Orientation, role: Qt.ItemDataRole) -> Any:
         """Overridden.
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -2,14 +2,17 @@
 
 import enum
 import functools
+import json
 import logging
 from typing import Any, Optional, Sequence, Tuple
 
 import requests
+import websockets
 from PyQt5.QtCore import (
     pyqtSignal, pyqtSlot, QAbstractTableModel, QModelIndex, QObject, Qt, QThread, QVariant
 )
 from PyQt5.QtWidgets import QAction, QPushButton, QTableView, QVBoxLayout, QWidget
+from websockets.sync.client import connect
 
 import qiwis
 from iquip.protocols import SubmittedExperimentInfo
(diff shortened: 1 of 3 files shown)


iquip/apps/scheduler.py
```diff
@@ -2,14 +2,17 @@
...
 import logging
 from typing import Any, Optional, Sequence, Tuple
 
 import requests
+import websockets
 from PyQt5.QtCore import (
     pyqtSignal, pyqtSlot, QAbstractTableModel, QModelIndex, QObject, Qt, QThread, QVariant
 )
 from PyQt5.QtWidgets import QAction, QPushButton, QTableView, QVBoxLayout, QWidget
+from websockets.sync.client import connect
```
kangz12345, line 15: I think since you already imported `websockets`, you don't have to import `connect` again (also it is not used many times).
Moreover, this module (`scheduler`) is not focused that much in `websockets` so `connect()` itself might seem confusing (although in this code you used proper context manager name so one can infer that it is for websockets easily).

iquip/apps/scheduler.py
```diff
@@ -23,68 +26,56 @@ class DeleteType(enum.Enum):
...
-                    content=expid.get("content", None),
-                    arguments=expid["arguments"]
-                ))
-            self.fetched.emit(schedule)
-            params["timestamp"] = timestamp
+        try:
+            with connect(self.url) as websocket:
+                while True:
+                    response = websocket.recv()
+                    response = json.loads(response)
```
kangz12345: How about just merging these two lines?
For example: `response = json.loads(websocket.recv())`

kangz12345 (changes requested):
Is this tested with the proper artiq proxy?

BECATRUE:
> Is this tested with the proper artiq proxy?

Sure! I tested if this PR works with snu-quiqcl/artiq-proxy#114.

iquip/apps/scheduler.py
BECATRUE: I see! Then, I will import only `websockets` module.

iquip/apps/scheduler.py
BECATRUE: Okay! I will fix it.

iquip/apps/scheduler.py
BECATRUE: Umm.. I don't know the exact reason, but if importing only this module, `connect()` cannot be found.

image

Temporarily, I left it as it was.

BECATRUE:
I replaced the infinite while loop with a for loop based on [websockets docs](https://websockets.readthedocs.io/en/stable/reference/sync/client.html#websockets.sync.client.ClientConnection.__iter__).

iquip/apps/scheduler.py
kangz12345: Oh, I see. Then it should be imported with full path.
In that case, let's keep the current code. It is readable enough!

kangz12345 (approved):
LGTM

PR #241, opened 2024-01-10
Update scheduler arguments and copy experiment info

### LINKED ISSUE #240: Show arguments in scheduler neatly
Now, the scheduler app shows all the decimal points of arguments, it may be too long.
Thus, I will show them neatly.

In addition, I will make an user copy the specific submitted experiment info.

To allow copy, I will use [QClipboard](https://doc.qt.io/qtforpython-5/PySide2/QtGui/QClipboard.html).

BECATRUE:
I implemented the followings:
- Show number arguments up to 9 decimal points (ns scale)
- Make an user copy a specific experiment info through right-click

The example result is as below.

image

This closes #240.

### DIFF
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -285,10 +285,12 @@ class SchedulerApp(qiwis.BaseApp):
         """Sets a copy action in schedulerFrame.scheduleView and its header."""
         view = self.schedulerFrame.scheduleView
         header = view.verticalHeader()
-        action = QAction("Copy", self)
-        action.triggered.connect(self.copyExperimentInfo)
-        view.addAction(action)
-        header.addAction(action)
+        viewAction = QAction("Copy", view)
+        headerAction = QAction("Copy", header)
+        viewAction.triggered.connect(self.copyExperimentInfo)
+        headerAction.triggered.connect(functools.partial(self.copyExperimentInfo, True))
+        view.addAction(viewAction)
+        header.addAction(headerAction)
 
     @pyqtSlot()
     def copyExperimentInfo(self, allInfo: bool = False):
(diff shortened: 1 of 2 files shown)


iquip/apps/scheduler.py
```diff
@@ -256,9 +276,27 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+    def setCopyAction(self):
+        """Sets a copy action in schedulerFrame.scheduleView."""
+        view = self.schedulerFrame.scheduleView
+        action = QAction("Copy", view)
+        action.triggered.connect(self.copyExperimentInfo)
+        view.addAction(action)
+
+    @pyqtSlot()
+    def copyExperimentInfo(self):
+        """Copies the selected experiment info to the system clipboard."""
```
kangz12345: Is it possible to copy the item at the selected column only?
For example only the RID or the arguments, etc.

BECATRUE: Sure! Indeed, actions can be applied to each index i.e., row and column, respectively.
However, I thought we don't need to copy each item, hence I applied same actions to all columns.

According to your suggestion, I will make one copy each item.
And I will add a function that copy the specific experiment info when right-click a row number.

kangz12345: Is the first column with indices the header?

BECATRUE: ~~No! The columns indicate only the table contents, and we can obtain the header through `view.verticalHeader()`.
Please refer to [setCopyAction](https://github.com/snu-quiqcl/iquip/blob/BECATRUE/240/update-scheduler-arguments/iquip/apps/scheduler.py#L287).~~

Yes! It means the header.

kangz12345 (approved):

Issue #238, opened 2024-01-05
Add codes for scannable argument processing


Aijuh:
### Feature you want to implement
Add codes at _ScanEntry class for scannable argument processing.

### How the feature is implemented
Below is the example of Scannable argument.
![image](https://github.com/snu-quiqcl/iquip/assets/43592775/1548e2ec-fd01-47ff-b2de-7e1c5de6e07d)

Scannable argument can be seperated into two parts. One is global arguments, the other is scan type dependent arguments ([ref](https://github.com/snu-quiqcl/artiq-proxy/discussions/111)).
By following original artiq convention, I will rename global arguments as `procdesc` and scan type dependent arguments as `state`.
I will add code for processing scannable argument Info from experiment file to attain desired `procdesc` and `state`.
### Additional context
Next PR will be adding acutal scannable widget for RangeScan display (#232).

Aijuh:
Closed

PR #242, opened 2024-01-15
Add code for argument information processing

Aijuh:
This closes issue #238. 

Main part of this PR is to add code for processing argument information at _ScanEntry class.

## PR description
Scannable argInfo has following information.

image

`default` is an dictionary that contains argument of each scan type. If scan type differs, its key and value are different. This information is called as `state` in this PR.
`unit`, `scale`, and etc are common to all scan type. This information is called as `procdesc` in this PR.
This information is used when widget for each scan type is created. 

For this PR, I did not implemented widget for each scan type. But this widget will use `state` and `procdesc` at the following PRs.

## Example image
image

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -3,12 +3,14 @@
 import json
 import logging
 from typing import Any, Dict, Optional, Tuple, Union
+from collections import OrderedDict
 
 import requests
 from PyQt5.QtCore import QDateTime, QObject, Qt, QThread, pyqtSignal, pyqtSlot
 from PyQt5.QtWidgets import (
-    QCheckBox, QComboBox, QDateTimeEdit, QDoubleSpinBox, QHBoxLayout, QLabel, QLineEdit,
-    QListWidget, QListWidgetItem, QPushButton, QVBoxLayout, QWidget
+    QButtonGroup, QCheckBox, QComboBox, QDateTimeEdit, QDoubleSpinBox, QGridLayout, QHBoxLayout,
+    QLabel, QLineEdit, QListWidget, QListWidgetItem, QPushButton, QRadioButton, QStackedWidget,
+    QVBoxLayout, QWidget
 )
 
 import qiwis
@@ -291,13 +293,16 @@ class _DateTimeEntry(_BaseEntry):
         return None
 
 
-# TODO(AIJUH): Add feature for argInfo processing and other scan type classes.
 class _ScanEntry(QWidget):
     """Entry class for a scannable object.
     
     Attributes:
         name: The name of the scannable object.
-        argInfo: The infomation of the arguments.
+        state: The dictionary that describes argument information of each scan type.
+        stackWidget: The QstackWidget that contains widget of each scan type.
+        layout: The layout of _ScanEntry widget.
+        widgets: The dictionary that contains widget of each scan type inside stackWidget.
+        radioButtons: The dictionary that contains buttons of each scan type for stackWidget.
     """
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


BECATRUE:
Oh, this PR is also too long.
Could you separate it into several PRs?

Aijuh:
I deleted all the scan type classes.
I will add RangeScan type first at the next PR.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
 
         Args:
             name: See the attributes section.
-            argInfo: See the attributes section.
+            argInfo: Each key and its value are:
+              default: The dictionary that describes arguments of a scannable object.
```
BECATRUE: I think "the default scannable object" is better instead of "a scannable object", because it describes "default" options.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+              unit: The unit of the number value.
+              scale: The scale factor that is multiplied to the number value.
+              global_step: The step between values changed by the up and down button.
+              global_min: The minimum value. (default=float("-inf"))
+              global_max: The maximum value. (default=float("inf"))
+                If min > max, then they are swapped.
```
BECATRUE, line 327: Is this supported by ARTIQ, or will you implement this function?

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
         self.name = name
-        self.argInfo = argInfo
+        self.state = self.get_state(argInfo)
+        self.stack = QStackedWidget(self)
+        self.layout = QGridLayout(self)
+        self.layout.addWidget(QLabel(name, self), 0, 0)
+        self.layout.addWidget(self.stack, 1, 1)
+
+    def get_state(self, argInfo):
+        """Gets a state dictionary that describes key parameters of a scannable object.
```
BECATRUE: How about "the default scannable object"?

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+
+        Creates a default state dictionary and updates it using the argInfo.
+
+        Args:
+            argInfo: A dictionary that contains argument information of each scannable type.
```
BECATRUE: Since you already wrote the description of `argInfo`, please write "argInfo: See argInfo in __ init __()." instead.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+        self.state = self.get_state(argInfo)
+        self.stack = QStackedWidget(self)
+        self.layout = QGridLayout(self)
+        self.layout.addWidget(QLabel(name, self), 0, 0)
+        self.layout.addWidget(self.stack, 1, 1)
+
+    def get_state(self, argInfo):
+        """Gets a state dictionary that describes key parameters of a scannable object.
+
+        Creates a default state dictionary and updates it using the argInfo.
```
BECATRUE: It looks unnecessary as the first line is sufficient.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
         super().__init__(parent=parent)
         self.name = name
-        self.argInfo = argInfo
+        self.state = self.get_state(argInfo)
+        self.stack = QStackedWidget(self)
+        self.layout = QGridLayout(self)
+        self.layout.addWidget(QLabel(name, self), 0, 0)
+        self.layout.addWidget(self.stack, 1, 1)
+
+    def get_state(self, argInfo):
```
BECATRUE: Please add the type hinting in terms of `argInfo` and return value.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
```
BECATRUE, line 317: I realized this now. Please remove one tab before `):` like below.
```python
def __init__(
  self,
  name: str,
  argInfo: Dict[str, Any],
  parent: Optional[QWidget] = None
):
```

iquip/apps/builder.py
```diff
@@ -291,13 +291,16 @@ def value(self) -> Optional[str]:
...
 
-# TODO(AIJUH): Add feature for argInfo processing and other scan type classes.
+# TODO(AIJUH): Add other scan type classes.
 class _ScanEntry(QWidget):
     """Entry class for a scannable object.
     
     Attributes:
         name: The name of the scannable object.
-        argInfo: The infomation of the arguments.
+        state: The dictionary that describes argument information of each scan type.
```
BECATRUE: I think you should write more description about the structure of `state`.
However, if you improve it in next PR, it is okay to remain it in this PR.

iquip/apps/builder.py
```diff
@@ -291,13 +291,16 @@ def value(self) -> Optional[str]:
         return None
 
 
-# TODO(AIJUH): Add feature for argInfo processing and other scan type classes.
+# TODO(AIJUH): Add other scan type classes.
 class _ScanEntry(QWidget):
```
BECATRUE, line 295: I have a question. Why did you inherit from `QWidget` directly, not `_BaseEntry`?

iquip/apps/builder.py
```diff
@@ -291,13 +291,16 @@ def value(self) -> Optional[str]:
...
     """Entry class for a scannable object.
     
     Attributes:
         name: The name of the scannable object.
-        argInfo: The infomation of the arguments.
+        state: The dictionary that describes argument information of each scan type.
+        stackWidget: The QstackWidget that contains widget of each scan type.
```
BECATRUE: "widgets"

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+              global_max: The maximum value. (default=float("inf"))
+                If min > max, then they are swapped.
+              ndecimals: The number of displayed decimals.
         """
         super().__init__(parent=parent)
         self.name = name
-        self.argInfo = argInfo
+        self.state = self.get_state(argInfo)
+        self.stack = QStackedWidget(self)
+        self.layout = QGridLayout(self)
```
BECATRUE, line 334: Why did you use `QGridLayout`?
Could you explain your plan briefly?

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+                    state[ty]["sequence"] = default["sequence"]
+                else:
+                    logger.warning("unknown default type: %s", ty)
+        return state
+
+    def get_procdesc(self, argInfo):
+        """Gets a prodesc dictionary that describes common parameters of the scannable object.
+
+        Args:
+            argInfo: A dictionary that contains argument information of the scannable object.
```
BECATRUE: Please update `argInfo` docstring like the review above.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+                    for key in "center span step randomize seed".split():
+                        state[ty][key] = default[key]
+                elif ty == "ExplicitScan":
+                    state[ty]["sequence"] = default["sequence"]
+                else:
+                    logger.warning("unknown default type: %s", ty)
+        return state
+
+    def get_procdesc(self, argInfo):
+        """Gets a prodesc dictionary that describes common parameters of the scannable object.
```
BECATRUE: "procdesc"

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+                elif ty == "CenterScan":
+                    for key in "center span step randomize seed".split():
+                        state[ty][key] = default[key]
+                elif ty == "ExplicitScan":
+                    state[ty]["sequence"] = default["sequence"]
+                else:
+                    logger.warning("unknown default type: %s", ty)
+        return state
+
+    def get_procdesc(self, argInfo):
```
BECATRUE: Please write type hintings.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+        """Gets a prodesc dictionary that describes common parameters of the scannable object.
+
+        Args:
+            argInfo: A dictionary that contains argument information of the scannable object.
+        """
+        procdesc = {
+            "unit": argInfo["unit"],
+            "scale": argInfo["scale"],
+            "global_step": argInfo["global_step"],
+            "ndecimals": argInfo["ndecimals"],
```
BECATRUE: In according to your docstring, please move `ndecimals` to the last.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+            "global_min": argInfo["global_min"],
+            "global_max":  argInfo["global_max"]
+        }
+        return procdesc
```
BECATRUE, line 382: I think you can return this dict directly, not defining a variable.
However it is just a coding style, do as you please.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+        """Gets a state dictionary that describes key parameters of a scannable object.
+
+        Creates a default state dictionary and updates it using the argInfo.
+
+        Args:
+            argInfo: A dictionary that contains argument information of each scannable type.
+        """
+        scale = argInfo["scale"]
+        state = {
+            "selected": "NoScan",
```
BECATRUE, line 346: Is it a temporary setting?

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+            "NoScan": {"value": 0.0, "repetitions": 1},
+            "RangeScan": {"start": 0.0, "stop": 100.0*scale, "npoints": 10,
```
BECATRUE: Please add a space before and after an operator like `100.0 * scale`.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+                          "randomize": False, "seed": None},
+            "CenterScan": {"center": 0.*scale, "span": 100.*scale,
```
BECATRUE: Please replace "span" with "step" same as #111.

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+                           "step": 10.*scale, "randomize": False,
+                           "seed": None},
+            "ExplicitScan": {"sequence": []}
+        }
+        if "default" in argInfo:
```
BECATRUE, line 355: Is it possible that the default is not given?

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+            defaults = argInfo["default"]
+            if not isinstance(defaults, list):
```
BECATRUE, line 357: When is its type not a list?

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+                    state[ty]["npoints"] = default["npoints"]
+                    state[ty]["randomize"] = default["randomize"]
+                    state[ty]["seed"] = default["seed"]
+                elif ty == "CenterScan":
+                    for key in "center span step randomize seed".split():
+                        state[ty][key] = default[key]
+                elif ty == "ExplicitScan":
+                    state[ty]["sequence"] = default["sequence"]
+                else:
+                    logger.warning("unknown default type: %s", ty)
```
BECATRUE: Please start with an upper case.

BECATRUE: And I think "scan type" is better instead of "default type".

iquip/apps/builder.py
```diff
@@ -309,11 +312,83 @@ def __init__(
...
+                           "step": 10.*scale, "randomize": False,
+                           "seed": None},
+            "ExplicitScan": {"sequence": []}
+        }
+        if "default" in argInfo:
+            defaults = argInfo["default"]
+            if not isinstance(defaults, list):
+                defaults = [defaults]
+            state["selected"] = defaults[0]["ty"]
+            for default in reversed(defaults):
```
BECATRUE: Why did you reverse `defaults`?

Also, I think this for loop can be shorten as follows:
```python
for default in defaults:
  ty = default.pop("ty")
  if ty not in ["NoScan", "RangeScan", "CenterScan", "ExplicitScan"]:
    logger.warning("Unknown scan type: %s", ty)
  state[ty] = default
```

BECATRUE (changes requested):

iquip/apps/builder.py
Aijuh: This min, max value will be used in QSpinBox, which has following properties ([ref](https://doc.qt.io/qt-6/qspinbox.html#minimum-prop)).
They will be adjusted automatically.

iquip/apps/builder.py
Aijuh: You are right.

iquip/apps/builder.py
Aijuh: `_BaseEntry` has value method, which returns selected value. When I first thought of it, I didn't inherit from `_BaseEntry` because `_ScanEntry` doesn't return a single value.

Moreover, since `_ScanEntry` use QGridLayout for design, I did not start from `_BaseEntry`.

iquip/apps/builder.py
Aijuh: I changed docstring as below. 
Detailed info of dictonary for each scannble type will be described in next PR. They will be added at each scannable type Entry class.

```
        state: Each key and its value are:
          "NoScan": The dictionary that contains argument info of NoScan scannable type.
          "RangeScan": The dictionary that contains argument info of RangeScan scannable type.
          "CenterScan": The dictionary that contains argument info of CenterScan scannable type.
          "ExplicitScan": The dictionary that contains argument info of ExplicitScan scannable type.
```

iquip/apps/builder.py
Aijuh: I followed original artiq code in terms of design.

iquip/apps/builder.py
Aijuh: I will reconsider it at the next PR.

iquip/apps/builder.py
Aijuh: corret

iquip/apps/builder.py
Aijuh: Do you want me to change variable name?
What do you want to say with this #111 reference?

iquip/apps/builder.py
Aijuh: As I tested, `default` is a list type that contains dictionary. 
But there might be the case that I don't know, so I followed original artiq code.

iquip/apps/builder.py
Aijuh: I wanted to make sure that each dictionary for each scan type in `state` contains only designated keys.
But since each scannable entry will access this dictionary with specific key, this implementation seems good.

Aijuh: I just used artiq code for reversing, but I also think it is meaningless.

iquip/apps/builder.py
Aijuh: With reference to [this](https://github.com/snu-quiqcl/artiq-proxy/discussions/111), scan object always has default.
Does argInfo for other non scan object has default?

Aijuh: As I checked, it seems that argInfo always has `default`

iquip/apps/builder.py
Aijuh: This PR has too many content.
Can you explain me in more detail?

Aijuh:
Thank you for the review @BECATRUE. Even if I double checked docstrings and fixed my codes, there were many things to improve. I changed all the point you gave me.

In terms of using layout, using QGridLayout is necessary for the design.
image
In the example image above, button box occupy the upper row. And QLabel and QSpinbox share one row. This design can be done by using QGridLayout. So I followed original artiq code in terms of design.

BECATRUE:
I think you didn't apply all reviews such as type hintings or docstrings.
Please check it again!

iquip/apps/builder.py
BECATRUE: Oh, please remove two blanks before and after "init".
I added those to prevent it from turning into bold according to markdown syntax.

iquip/apps/builder.py
```diff
@@ -297,8 +297,13 @@ class _ScanEntry(QWidget):
     
     Attributes:
         name: The name of the scannable object.
-        state: The dictionary that describes argument information of each scan type.
-        stackWidget: The QstackWidget that contains widget of each scan type.
+        state: Each key and its value are:
```
BECATRUE: I think there are too many columns, hence how about "Each key and its value are as follows."?

iquip/apps/builder.py
BECATRUE: I see. There seems to be several differences between `_BaseEntry` and `_ScanEntry`.

iquip/apps/builder.py
BECATRUE: Please update to `__init__()`.

iquip/apps/builder.py
```diff
@@ -354,39 +357,27 @@ def get_state(self, argInfo):
...
+            argInfo: See argInfo in __ init __().
         """
         procdesc = {
             "unit": argInfo["unit"],
             "scale": argInfo["scale"],
             "global_step": argInfo["global_step"],
-            "ndecimals": argInfo["ndecimals"],
             "global_min": argInfo["global_min"],
-            "global_max":  argInfo["global_max"]
+            "global_max":  argInfo["global_max"],
```
BECATRUE: Please remove one blank.

iquip/apps/builder.py
BECATRUE: Oh, I'm really sorry that I confused you.

I intended to say snu-quiqcl/artiq-proxy#111.
I want just to replace the positions of "span" and "step" as follows:
```python
{"center": ..., "step": ..., "span": ..., ...}
```

iquip/apps/builder.py
BECATRUE: Then, did you leave it in case you didn't know?

BECATRUE (changes requested):
Both "scannable" and "scan" are used, so how about unifying them into one?

Aijuh:
I will follow https://github.com/snu-quiqcl/artiq-proxy/discussions/111 for the order of variable in later PR's docstring.

```
                   "seed": None},
    "ExplicitScan": {"sequence": []}
}
if "default" in argInfo:

```
For this code, I just leave it because there might be the case that default is not given. If this case happens, we can select scan type and its parameters at our GUI.

iquip/apps/builder.py
```diff
@@ -291,29 +291,95 @@ def value(self) -> Optional[str]:
...
+        Args:
+            argInfo: See argInfo in __init__().
+        """
+        scale = argInfo["scale"]
+        state = {
+            "selected": "NoScan",
+            "NoScan": {"value": 0.0, "repetitions": 1},
+            "RangeScan": {"start": 0.0, "stop": 100.0 * scale, "npoints": 10,
+                          "randomize": False, "seed": None},
+            "CenterScan": {"center": 0. * scale,"step": 10. * scale,
```
BECATRUE: Please add a space before "step".

BECATRUE (changes requested):
Final!

BECATRUE (approved):
LGTM!

PR #243, opened 2024-01-18
Fetch dataset information using web socket

### LINKED ISSUE #236: Fetch dataset information using web socket
According to snu-quiqcl/artiq-proxy#115, I will use web socket to fetch dataset information.

#### In `_DatasetListThread`
- Remove the corresponding button and start this when the app is started.
- `run()`: Connect to the proxy server and receive dataset list consistently.

#### In `_DatasetFetcherThread`
- `__init__()`: Initialize a web socket object.
- `_initialize()`: Leave it almost the same, but remove `timestamp`.
- `run()`: Connect to the proxy server and receive dataset modification consistently.
- `stop()`: Close the web socket object.

BECATRUE:
Now, the dataviewer fetches dataset through web socket implemented in snu-quiqcl/artiq-proxy#117.

In GUI, the only modification is that the button to update dataset list is removed.

image

I tested that
- the dataset list is updated in real-time.
- the selected dataset is fetched in real-time.
- clicking the stop button is reflected immediately.

For test, the following example experiment may be helpful.

```python
import time
from artiq.experiment import *

class DatasetTest(EnvExperiment):
    """Dataset Test"""

    def build(self):
        self.setattr_device("core")

    @kernel
    def run(self):
        self.set_dataset("test_a", [], broadcast=True)
        self.set_dataset("test_a.parameters", ["shot"], broadcast=True)
        self.set_dataset("test_a.units", [""], broadcast=True)
        time.sleep(5)

        for cnt in range(10):
            self.append_to_dataset("test_a", [cnt, cnt])
            time.sleep(1)
        time.sleep(5)

        self.set_dataset("test_a", [], broadcast=True)
        time.sleep(1)
        for cnt in range(10):
            self.append_to_dataset("test_a", [cnt, cnt])
            time.sleep(1)

```

This closes #236.

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -942,6 +942,7 @@ class _DatasetFetcherThread(QThread):
 
     def _initialize(self):
         """Fetches the target dataset to initialize the local dataset."""
+        self.websocket = connect(self.url)
         self.websocket.send(json.dumps(self.name))
         rawDataset = json.loads(self.websocket.recv())
         if not rawDataset:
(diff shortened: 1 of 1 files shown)


iquip/apps/dataviewer.py
```diff
@@ -876,29 +872,28 @@ def __init__(self, ip: str, port: int, parent: Optional[QObject] = None):
...
         """Overridden."""
         try:
-            response = requests.get(self.url, timeout=5)
-            response.raise_for_status()
-        except requests.exceptions.RequestException:
-            logger.exception("Failed to GET %s", self.url)
-            return
-        self.fetched.emit(self._filter(response.json()))
+            with connect(self.url) as websocket:
+                for response in websocket:
```
kangz12345, line 893: Comment for later consideration:
when the server or client quits, does the other side know it?
In other words, if the server is closed, does the client still wait for the next message (and vice versa)?

BECATRUE: As I tested, the client and server are good at detecting the event that the other is terminated even if it is abnormal.

If the client is terminated, in the server, the web socket is closed and the function is terminated.
In the opposite case, in the client, the event is caught by a `try-except` statement e.g. `except WebSocketException`.

Hence, I don't think anything you're worried about is going to happen.

BECATRUE:
I realized that this app fetches the dataset list from the proxy server regardless of whether the current status is "realtime" or "remote". The correct implementation is that it fetches the list only when the current status is "realtime", but I will handle this in the next PR, #258, hence please never mind.

kangz12345 (approved):
LGTM

PR #248, opened 2024-01-30
Threaded motorized stage manager

kangz12345:
This PR implements `StageManager` which manages several stage RPC clients and controls them in a thread (other than the main GUI thread) to prevent GUI from freezing.

I tested it working, using the following code:

```python
app = QApplication([])

thread = QThread()
manager = StageManager()
manager.moveToThread(thread)
thread.start()

manager.connectStage.emit("test", ("ip", port, "serial_number"))
proxy = StageProxy(manager, "test")

widget = StageWidget()  # will be introduced in the next PR (already implemented)
widget.moveBy.connect(proxy.moveBy)
widget.show()

app.exec_()
```

However, I forgot to save the actual test script, so there might be mistakes in the example code above.
Please don't hesitate to ask.


iquip/apps/stage.py
```diff
@@ -0,0 +1,163 @@
...
+              the previous one is closed and replaced with the new one.
+            info: RPC target information tuple.
+        """
+        client = self._clients.get(key, None)
+        if client is not None:
+            client.close_rpc()
+        self._clients[key] = Client(*info)
+
+    @pyqtSlot(str, float)
+    @use_client
```
BECATRUE, line 121: It looks too clear!

iquip/apps/stage.py
```diff
@@ -0,0 +1,163 @@
...
+    If the key does not exist, function is not called at all.
+    If an OSError occurs while running function, the RPC client is closed and
+        removed from the client dictionary. 
+    
+    Args:
+        function: Decorated function. It should take a Client object as the
+            first argument.
+    """
+    @functools.wraps(function)
+    def wrapped(self: "StageManager", key: str, *args, **kwargs):
```
BECATRUE: Referred to [stackoverflow](https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class), and [PEP 563](https://peps.python.org/pep-0563/#enabling-the-future-behavior-in-python-3-7), you might be able to use `__future__` module.

Please check it!

iquip/apps/stage.py
```diff
@@ -0,0 +1,163 @@
...
+        Args:
+            key: String key for identifying the client.
+        """
+        client = self._clients.get(key, None)  # pylint: disable=protected-access
+        if client is None:
+            logger.error("Failed to get client %s.", key)
+            return
+        try:
+            function(self, client, *args, **kwargs)
+        except OSError:
```
BECATRUE: Why did you check only `OSError`?
According to [Client](https://github.com/m-labs/sipyco/blob/master/sipyco/pc_rpc.py#L70), there seems to be possibilities that `ValueError` and `AttributeError` are raised.

iquip/apps/stage.py
```diff
@@ -0,0 +1,163 @@
...
+            logger.exception("Error occurred while running %s with client %s.", function, key)
```
BECATRUE: Did you intend `function`, instead of `function.__name__`?

BECATRUE (changes requested):

iquip/apps/stage.py
kangz12345: Good to know! Thank you, I will apply it.

iquip/apps/stage.py
kangz12345: Good idea, I will apply that.

iquip/apps/stage.py
kangz12345: That's correct! I guess it would be good to catch them all. Even though `AttributeError` is the programmer's responsibility, catching it will helpful for stability.

iquip/apps/stage.py
```diff
@@ -0,0 +1,163 @@
...
+    
+    Usage:
+        proxy = StageProxy(manager, key)
+        proxy.signal(x, y)  # equivalent to: manager.signal.emit(key, x, y)
+    """
+
+    def __init__(self, manager: StageManager, key: str):
+        """
+        Args:
+            manager: StageManager object where the target stage client object lives.
```
BECATRUE: The `manager` and `key` should be moved to the attributes section.

BECATRUE (changes requested):

BECATRUE (approved):
LGTM

PR #249, opened 2024-01-31
Implement RangeScan

Aijuh:
This closes #233.
##

RangeScan is implemented as below. Any change of values in RangeScan widget is saved at `state` dictionary inside ScanEntry. 

image

##
When submitting an experiment as a scannable, the example format looks like this:
`{'num_var1': 0.0, 'num_var2': 9.999999999999999e-05, 'Scan1': {'ty': 'RangeScan', 'start': 2, 'stop': 11.199999999999996, 'npoints': 5, 'randomize': False, 'seed': None}, 'Scan2': {'ty': 'RangeScan', 'start': 4.0, 'stop': 3.0, 'npoints': 5, 'randomize': True, 'seed': None}}`

The artiq master get experiment information as below:
`{'log_level': 30, 'file': 'simple_simulation.py', 'class_name': 'SimpleSimulation', 'arguments': {'n1': 0.0, 'n2': 4.9999999999999996e-06, 'a': {'npoints': 5, 'randomize': False, 'seed': None, 'start': 6.0, 'stop': 5.0, 'ty': 'RangeScan'}, 'b': {'npoints': 4, 'randomize': False, 'seed': None, 'start': 2.0, 'stop': 9.5, 'ty': 'RangeScan'}}, 'repo_rev': None}`

If you concentrate on argument format, you can see that they are in same format. 
As I know, artiq-proxy gets expInfo from local and send it to master as a chunk. So it seems that no change in artiq-proxy is needed to deal with scannable.

## Caveat
If we change number inside QSpinBox by pressing up or down button, the number will not be accurate. 
rangescan sidebar

For example, If you change `stop` value as 10.5 by pressing up button 5 times, actual number is 10.499999999999998, not 10.5.
So it is desirable to change number by entering number inside a QSpinBox.


Aijuh:
Inside `_RangeScan` class, there are functions inside `__init__`. I don't think it seems good, but for now I followed original artiq code.

BECATRUE:
First, didn't we decide to include the seed entry?

iquip/apps/builder.py
```diff
@@ -330,10 +330,12 @@ def __init__(
         super().__init__(parent=parent)
         self.name = name
         self.state = self.get_state(argInfo)
-        self.stack = QStackedWidget(self)
```
BECATRUE, line 333: Will you use a stack strucutre in later?

BECATRUE: As I know, you planned to use stack for stacking each scan type entry.

iquip/apps/builder.py
```diff
@@ -683,3 +709,118 @@ def onSubmitted(self, rid: int):
     def frames(self) -> Tuple[Tuple[str, BuilderFrame]]:
         """Overridden."""
         return (("", self.builderFrame),)
+
+
+class _RangeScan(QWidget):
```
BECATRUE: Please move this class before or after `_ScanEntry`.

iquip/apps/builder.py
```diff
@@ -330,10 +330,12 @@ def __init__(
         super().__init__(parent=parent)
         self.name = name
         self.state = self.get_state(argInfo)
-        self.stack = QStackedWidget(self)
+        procdesc = self.get_procdesc(argInfo)
+        self.state = self.get_state(argInfo)
```
BECATRUE: Please remove it because you already wrote this before.

iquip/apps/builder.py
BECATRUE: If you plan that, leaving it looks better.

iquip/apps/builder.py
```diff
@@ -381,6 +383,11 @@ def get_procdesc(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
         }
         return procdesc
 
+    def scannable_info(self) -> Dict[str, Any]:
+        """Gets a ditctionary of scannable arguments from _ScanEntry."""
```
BECATRUE: "dictionary"

BECATRUE: In addition, I think you can make this class inherrited from `_BaseEntry` and rename `scannable_info()` to `value()`.

Of course, as you said, the layout type is different, but it might be resolved easily through modifying `_BaseEntry`.

iquip/apps/builder.py
```diff
@@ -683,3 +709,118 @@ def onSubmitted(self, rid: int):
...
+            state: Each key and its value as follows.
+              start: The start point for the RangeScan sequence.
+              stop: The end point for the RangeScan sequence.
+              npoints: The number of points in the RangeScan squence.
+              randomize: A boolean value that decides whether to shuffle the RangeScan sequence.
+        """
+        super().__init__(parent=parent)
+        scale = procdesc["scale"]
+
+        def apply_properties(widget):
```
BECATRUE: How about making it a method, not an internal function?

Also, please add the type hinting.

iquip/apps/builder.py
```diff
@@ -683,3 +709,118 @@ def onSubmitted(self, rid: int):
...
+            """Adds properties to the SpinBox widget.
+
+            Attributes:
+               widget: A QDOubleSpinWidget that has properties to set. 
+            """
+            widget.setDecimals(procdesc["ndecimals"])
+            if procdesc["global_min"] is not None:
+                widget.setMinimum(procdesc["global_min"]/scale)
```
BECATRUE: Please add a space before and after an operator.

iquip/apps/builder.py
```diff
@@ -683,3 +709,118 @@ def onSubmitted(self, rid: int):
...
+
+        def update_randomize():
+            """Updates the current randomize boolean value from _RangeScan argument.
+            
+            If the checked state of randomize button changes, this is called.
+            """
+            state["randomize"] = randomize.isChecked()
+            randomize.setChecked(state["randomize"])
+
+        start.valueChanged.connect(update_start)
```
BECATRUE: Why did you add a slot?
Isn't it sufficient to get only the final state?

BECATRUE (changes requested):

iquip/apps/builder.py
Aijuh: I missed wrong spelling....

If you want to use `BaseEntry`, its layout should change from `QHBoxLayout` to `QGridLayout`. Of course GridLayout has more option and it is possible. But one should change all the layout of basic Entries. Do you want that or not?

Aijuh:
I am not sure about seed argument. Artiq-dashboard did not implemented it inside gui.
But it still use seed value when shuffle sequence later. 

I don't think entering seed value every time is not good idea. I will search for seed arguments inside artiq for more detail.

iquip/apps/builder.py
```diff
@@ -330,10 +332,12 @@ def __init__(
         super().__init__(parent=parent)
         self.name = name
         self.state = self.get_state(argInfo)
+        procdesc = self.get_procdesc(argInfo)
         self.stack = QStackedWidget(self)
+        self.rangeWidget = _RangeScan(procdesc, self.state["RangeScan"])
         self.layout = QGridLayout(self)
         self.layout.addWidget(QLabel(name, self), 0, 0)
-        self.layout.addWidget(self.stack, 1, 1)
+        self.layout.addWidget(self.rangeWidget, 0, 1)
```
BECATRUE: Could you explain your future plan? I think it might be updated to `self.stack` instead of `self.rangeWidget`.

BECATRUE:
> I don't think entering seed value every time is not good idea. I will search for seed arguments inside artiq for more detail.

I also agree with you. Hence, I think the best solution is the followings.
First, the default option should be given automatically.
However, if a user wants to give a seed, the GUI should be able to support it.

Could you find whether it is possible?

iquip/apps/builder.py
BECATRUE: Umm.. I came up with an idea! How about not modifying `_BaseEntry` and put `QGridLayout` into `self.layout`?
The example is as below.

```python
class _ScanEntry(_BaseEntry):

  def __init__(...):
    ...
    layout = QGridLayout(self)
    ...
    self.layout.addLayout(layout)
```

As I understand, you seem to want to leave the same column as the name label empty, so I think it will be the same.

BECATRUE (changes requested):
I will add other reviews after we decide if `_ScanEntry` is inheritted from `_BaseEntry`.

iquip/apps/builder.py
Aijuh: Yes I will change it to `self.stack`.

Aijuh:
It works well as below.
image

Inside QHLayout, name and QGridLayout is inserted. And I will add more widget at QGridLayout.

iquip/apps/builder.py
```diff
@@ -327,13 +329,14 @@ def __init__(
                 If min > max, then they are swapped.
               ndecimals: The number of displayed decimals.
         """
```
BECATRUE, line 321: I think you can write `def __init__(...):` in just a line.

iquip/apps/builder.py
```diff
@@ -327,13 +329,14 @@ def __init__(
...
-        super().__init__(parent=parent)
-        self.name = name
+        super().__init__(name, {}, parent=parent)
```
BECATRUE: Did you intend to pass `{}` for `argInfo`? Please see other entry classes like `_BooleanEntry`.

iquip/apps/builder.py
```diff
@@ -292,7 +293,7 @@ def value(self) -> Optional[str]:
 
 
 # TODO(AIJUH): Add other scan type classes.
-class _ScanEntry(QWidget):
+class _ScanEntry(_BaseEntry):
     """Entry class for a scannable object.
     
     Attributes:
```
BECATRUE, line 299: You should remove `name` and `layout` docstring because it is an attribute of `_BaseEntry`.

iquip/apps/builder.py
```diff
@@ -304,6 +305,7 @@ class _ScanEntry(QWidget):
           "CenterScan": The dictionary that contains argument info of CenterScan scannable type.
           "ExplicitScan": The dictionary that contains argument info of ExplicitScan scannable type.
```
BECATRUE: I'm saying this late, but these four lines (of each scanabble type) look similar.
How about modifying it as follows.
```
"NoScan", "RangeScan", "CenterScan", and "ExplicitScan": The dictionary that contains argument info of the corresponding scannable type.
```

iquip/apps/builder.py
```diff
@@ -327,13 +329,14 @@ def __init__(
...
         self.stack = QStackedWidget(self)
-        self.layout = QGridLayout(self)
-        self.layout.addWidget(QLabel(name, self), 0, 0)
-        self.layout.addWidget(self.stack, 1, 1)
+        self.rangeWidget = _RangeScan(procdesc, self.state["RangeScan"])
+        layout = QGridLayout(self)
+        layout.addWidget(self.rangeWidget, 0, 1)
+        self.layout.addLayout(layout)
 
     def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
```
BECATRUE: If you pass `argInfo` to the constructor of its super class, you can get `argInfo` through `self.argInfo`.

iquip/apps/builder.py
```diff
@@ -304,6 +305,7 @@ class _ScanEntry(QWidget):
           "CenterScan": The dictionary that contains argument info of CenterScan scannable type.
           "ExplicitScan": The dictionary that contains argument info of ExplicitScan scannable type.
         stackWidget: The QstackWidget that contains widgets of each scannable type.
+        rangeWidget: The Widget that will be removed at next PR.
```
BECATRUE: Please write it with TODO comment.

iquip/apps/builder.py
```diff
@@ -327,13 +329,14 @@ def __init__(
                 If min > max, then they are swapped.
               ndecimals: The number of displayed decimals.
         """
-        super().__init__(parent=parent)
-        self.name = name
+        super().__init__(name, {}, parent=parent)
         self.state = self.get_state(argInfo)
+        procdesc = self.get_procdesc(argInfo)
         self.stack = QStackedWidget(self)
```
BECATRUE: Can you add a TODO comment with a brief plan?

iquip/apps/builder.py
```diff
@@ -327,13 +329,14 @@ def __init__(
...
-        self.layout = QGridLayout(self)
-        self.layout.addWidget(QLabel(name, self), 0, 0)
-        self.layout.addWidget(self.stack, 1, 1)
+        self.rangeWidget = _RangeScan(procdesc, self.state["RangeScan"])
+        layout = QGridLayout(self)
```
BECATRUE: To avoid confusion between `layout` and `self.layout`, it looks better to rename `layout` to something else.

iquip/apps/builder.py
```diff
@@ -381,6 +384,131 @@ def get_procdesc(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
         }
         return procdesc
 
+    def value(self) -> Dict[str, Any]:
+        """Gets a dictionary of scannable arguments from _ScanEntry."""
```
BECATRUE: You should write "Overridden." becase it is an overridden method. Please see `value()` in other entry classes.

BECATRUE (changes requested):
I reviewed only `_ScanEntry` class.

Aijuh:
I fixed codes with respect to feedback.
Even if the code worked well, the following message occured at terminal. 
It is because we are adding QVlayout to the predefined QHlayout of ScanEntry.

image

BECATRUE:
The referred warning should be resolved.
Is it resolved by modifying `scanLayout = QVBoxLayout(self)` to `scanLayout = QVBoxLayout()`?

iquip/apps/builder.py
```diff
@@ -292,27 +293,19 @@ def value(self) -> Optional[str]:
...
           "selected": The name of the selected scannable type.
-          "NoScan": The dictionary that contains argument info of NoScan scannable type.
-          "RangeScan": The dictionary that contains argument info of RangeScan scannable type.
-          "CenterScan": The dictionary that contains argument info of CenterScan scannable type.
-          "ExplicitScan": The dictionary that contains argument info of ExplicitScan scannable type.
+          "NoScan", "RangeScan", "CenterScan", and "ExplicitScan": The dictionary that contains 
+            argument info of the corresponding scannable type.
         stackWidget: The QstackWidget that contains widgets of each scannable type.
-        layout: The layout of _ScanEntry widget.
         radioButtons: The dictionary that contains buttons of each scannable type for stackWidget.
```
BECATRUE: It should be removed because it doesn't exist!

iquip/apps/builder.py
```diff
@@ -292,27 +293,19 @@ def value(self) -> Optional[str]:
...
-        name: The name of the scannable object.
         state: Each key and its value are as follows.
           "selected": The name of the selected scannable type.
-          "NoScan": The dictionary that contains argument info of NoScan scannable type.
-          "RangeScan": The dictionary that contains argument info of RangeScan scannable type.
-          "CenterScan": The dictionary that contains argument info of CenterScan scannable type.
-          "ExplicitScan": The dictionary that contains argument info of ExplicitScan scannable type.
+          "NoScan", "RangeScan", "CenterScan", and "ExplicitScan": The dictionary that contains 
+            argument info of the corresponding scannable type.
         stackWidget: The QstackWidget that contains widgets of each scannable type.
```
BECATRUE: "QStackedWidget"

BECATRUE: And its name is wrong. Please modify it.

iquip/apps/builder.py
```diff
@@ -327,21 +320,23 @@ def __init__(
...
+        scanLayout = QVBoxLayout(self)
+        scanLayout.addWidget(self.rangeWidget)
+        self.layout.addLayout(scanLayout)
 
-    def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
+    def get_state(self) -> Dict[str, Any]:
         """Gets a dictionary that describes default parameters of all scannable types.
 
         Args:
             argInfo: See argInfo in __init__().
```
BECATRUE: Please remove it.

BECATRUE: And please modify `get_procdesc()` likewise.

iquip/apps/builder.py
```diff
@@ -381,6 +376,134 @@ def get_procdesc(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
         }
         return procdesc
 
+    def value(self) -> Dict[str, Any]:
+        """Overridden.
+        
+        Returns a dictionary of scannable arguments from _ScanEntry.
```
BECATRUE: I think "from _ScanEntry" is unnecessary because this method is in that class.

How about "Returns the dictionary about the selected scannable argument"?

BECATRUE (changes requested):
Please check the docstring and some typos!

iquip/apps/builder.py
```diff
@@ -292,27 +293,18 @@ def value(self) -> Optional[str]:
...
-          "NoScan": The dictionary that contains argument info of NoScan scannable type.
-          "RangeScan": The dictionary that contains argument info of RangeScan scannable type.
-          "CenterScan": The dictionary that contains argument info of CenterScan scannable type.
-          "ExplicitScan": The dictionary that contains argument info of ExplicitScan scannable type.
-        stackWidget: The QstackWidget that contains widgets of each scannable type.
-        layout: The layout of _ScanEntry widget.
-        radioButtons: The dictionary that contains buttons of each scannable type for stackWidget.
+          "NoScan", "RangeScan", "CenterScan", and "ExplicitScan": The dictionary that contains 
+            argument info of the corresponding scannable type.
+        stack: The QstackedWidget that contains widgets of each scannable type.
```
BECATRUE: Please modify "QstackWidget" to "QStackedWidget"!

BECATRUE: And instead of "stack", "stackWidget" looks better because it implies that this attribute is a widget.

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+            "global_min": self.argInfo["global_min"],
+            "global_max": self.argInfo["global_max"],
+            "ndecimals": self.argInfo["ndecimals"]
         }
         return procdesc
 
+    def value(self) -> Dict[str, Any]:
+        """Overridden.
+        
+        Returns the dictionary of a selected scannable arguments.
```
BECATRUE: "the selected"

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        """
+        selected = self.state["selected"]
+        return self.state[selected]
+
+
+class _RangeScan(QWidget):
+    """A RangeScan widget in Scannable Entry.
```
BECATRUE: Please use lower cases as possible, except to name of class or variable.

BECATRUE: For example, "A widget for range scan in _ScanEntry class.".

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+
+    Attributes: See the docstring of Args at __init__ for more detail.
```
BECATRUE: You should write the most detailed docstring possible in attribute section because here is an explanation for overall class properties and an argument section is an explanation only for a method.

BECATRUE: Hence, please reorganize those docstrings.

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        npoints: QSpinBox for npoints arguments inside state.
+        randomize: QCheckBox for randomize arguments inside state.
+        layout: The layout of _RangeScan widget.
+    """
+    def __init__(
+        self,
+        procdesc: Dict[str, Any],
+        state: Dict[str, Any],
+        parent: Optional[QWidget] = None
+        ): # pylint: disable=too-many-statements
```
BECATRUE: This is just question! After you shortened the length of this method, is this Pylint comment necessary?

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self.start.valueChanged.connect(self.update_start)
+        self.npoints.valueChanged.connect(self.update_npoints)
+        self.stop.valueChanged.connect(self.update_stop)
+        self.randomize.stateChanged.connect(self.update_randomize)
+
+    def apply_properties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
+        """Adds properties to the SpinBox widget.
+
+        Attributes:
+            widget: A QDOubleSpinWidget that has properties to set.
```
BECATRUE: Please add a docstring about `procdesc`.

BECATRUE: "QDoubleSpinWidget"

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self.randomize.stateChanged.connect(self.update_randomize)
+
+    def apply_properties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
+        """Adds properties to the SpinBox widget.
+
+        Attributes:
+            widget: A QDOubleSpinWidget that has properties to set. 
+        """
+        widget.setDecimals(procdesc["ndecimals"])
+        if procdesc["global_min"] is not None:
```
BECATRUE: Accessing each value using `procdesc[...]` looks messy.
How about implementing like `_NumberEntry`?

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+            widget.setMinimum(procdesc["global_min"] / self.scale)
+        else:
+            widget.setMinimum(float("-inf"))
+        if procdesc["global_max"] is not None:
+            widget.setMaximum(procdesc["global_max"] / self.scale)
+        else:
+            widget.setMaximum(float("inf"))
```
BECATRUE: I just wonder why you set the default maximum value to "inf", because in `_NumberEntry`, I set it to 99.99.

iquip/apps/builder.py
BECATRUE: I'd like to unify the implementation way in `_NumberEntry` and this method, if there is no more efficient way.

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        if procdesc["global_step"] is not None:
+            widget.setSingleStep(procdesc["global_step"] / self.scale)
+        if procdesc["unit"]:
+            widget.setSuffix(" " + procdesc["unit"])
```
BECATRUE: I think the space is unnecessary. Is there a certain reason?

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+
+    def update_start(self, value: float):
```
BECATRUE: Where are these four following methods used?

BECATRUE (changes requested):

iquip/apps/builder.py
Aijuh: I had no reason. I will set range as 0~99.99.

iquip/apps/builder.py
Aijuh: That seems good.

iquip/apps/builder.py
Aijuh: These functions are connected by signals.
```
        self.start.valueChanged.connect(self.update_start)
        self.npoints.valueChanged.connect(self.update_npoints)
        self.stop.valueChanged.connect(self.update_stop)
        self.randomize.stateChanged.connect(self.update_randomize)

```

Aijuh: At line 438

Aijuh:
> The referred warning should be resolved. Is it resolved by modifying `scanLayout = QVBoxLayout(self)` to `scanLayout = QVBoxLayout()`?

Previous warning message was solved in this way..

iquip/apps/builder.py
BECATRUE: > Please modify "QstackWidget" to "QStackedWidget"!

It isn't applied yet. @Aijuh

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+    """A widget for range scan in _ScanEntry.
+
+    Attributes:
+        scale: The scale factor that is multiplied to the number value.
+        state: Each key and its value as follows.
+          start: The start point for the RangeScan sequence.
+          stop: The end point for the RangeScan sequence.
+          npoints: The number of points in the RangeScan squence.
+          randomize: A boolean value that decides whether to shuffle the RangeScan sequence.
+        start: QDoubleSpinBox for start arguments inside state.
```
BECATRUE: I think naming a widget variable with its type, e.g., "StartSpinBox" because one might think "start" implies a value, not a widget.
What do you think of it?

BECATRUE: And "start argument", not "start arguments" looks correct, because it is singular.
Hence, "QDoubleSpinBox for the start argument in state." looks sufficient.

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        stop: QDoubleSpinBox for stop arguments inside state.
+        npoints: QSpinBox for npoints arguments inside state.
+        randomize: QCheckBox for randomize arguments inside state.
+        layout: The layout of _RangeScan widget.
```
BECATRUE: This is just my coding style, but I prefer not to mention the name of class to which it belongs in its docstring if possible.
Hence, how about "The outermost layout."?

iquip/apps/builder.py
Aijuh: I missed upper case at 'S'. Thank you.

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+    """
+    def __init__(
+        self,
+        procdesc: Dict[str, Any],
+        state: Dict[str, Any],
+        parent: Optional[QWidget] = None
+        ):
+        """Extended.
+
+        Args: See the docstring of Attributes for more detail.
```
BECATRUE: Please write the docstring in the original format like [dataviewer.py](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/dataviewer.py#L939).

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+            procdesc: Each key and its value are as follows.
+              unit: The unit of the number value.
+              scale: See scale docstring at class Attributes.
+              global_step: The step between values changed by the up and down button.
+              global_min: The minimum value. (default=float("-inf"))
```
BECATRUE: Please update the default value.

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self.npoints = QSpinBox(self)
+        self.npoints.setMinimum(1)
+        self.npoints.setMaximum((1 << 31) - 1)
+        self.npoints.setValue(state["npoints"])
+        self.stop = QDoubleSpinBox(self)
+        self.apply_properties(self.stop, procdesc)
+        self.stop.setValue(state["stop"] / self.scale)
+        self.randomize = QCheckBox("Randomize", self)
+        self.randomize.setChecked(state["randomize"])
+        #layout
```
BECATRUE: "# layout"

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        super().__init__(parent=parent)
+        self.scale = procdesc["scale"]
+        self.state = state
+        self.layout = QGridLayout(self)
+        self.start = QDoubleSpinBox(self)
+        self.apply_properties(self.start, procdesc)
+        self.start.setValue(state["start"] / self.scale)
+        self.npoints = QSpinBox(self)
+        self.npoints.setMinimum(1)
+        self.npoints.setMaximum((1 << 31) - 1)
```
BECATRUE: This is a question! Is there any reason to set the maximum value to this?

iquip/apps/builder.py
BECATRUE: I have noticed that this question has not yet been resolved. @Aijuh

iquip/apps/builder.py
BECATRUE: I mentioned you just before to the previous review.

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self.npoints.valueChanged.connect(self.update_npoints)
+        self.stop.valueChanged.connect(self.update_stop)
+        self.randomize.stateChanged.connect(self.update_randomize)
+
+    def apply_properties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
+        """Adds properties to the SpinBox widget.
+
+        Attributes:
+            widget: A QDoubleSpinWidget that has properties to set.
+            procdesc: See the docstring at __init__ for more detail.
```
BECATRUE: Please write this docstring in this format, [builder.py](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/builder.py#L342).

iquip/apps/builder.py
```diff
@@ -365,22 +355,146 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self.layout.addWidget(QLabel("stop:", self), 2, 0)
+        self.layout.addWidget(self.stop, 2, 1)
+        self.layout.addWidget(self.randomize, 3, 1)
+        self.start.valueChanged.connect(self.update_start)
+        self.npoints.valueChanged.connect(self.update_npoints)
+        self.stop.valueChanged.connect(self.update_stop)
+        self.randomize.stateChanged.connect(self.update_randomize)
+
+    def apply_properties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
+        """Adds properties to the SpinBox widget.
```
BECATRUE: If you want to write upper letters, it should be a valid class or variable name.
In other words, you should write "the spin box widget" or "the QDoubleSpinBox widget".

iquip/apps/builder.py
BECATRUE: I meant that `widget.setSuffix(unit)` looks sufficient.

iquip/apps/builder.py
Aijuh: It was at original artiq code. I don't know exact number that is needed for experiments.

BECATRUE:
Please revert `config_template.json`.
It should be exist.

@Aijuh

iquip/apps/builder.py
```diff
@@ -314,8 +314,8 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
               unit: The unit of the number value.
               scale: The scale factor that is multiplied to the number value.
               global_step: The step between values changed by the up and down button.
-              global_min: The minimum value. (default=float("-inf"))
-              global_max: The maximum value. (default=float("inf"))
+              global_min: The minimum value. (default=float("0"))
+              global_max: The maximum value. (default=float("99.99"))
```
BECATRUE: Like [builder.py](https://github.com/snu-quiqcl/iquip/blob/develop/iquip/apps/builder.py#L166), `float()` is unnecessary.

iquip/apps/builder.py
```diff
@@ -400,52 +400,52 @@ def __init__(
         ):
         """Extended.
 
-        Args: See the docstring of Attributes for more detail.
+        Args: See the attributes section for more details.
```
BECATRUE: Please write just "Args:".

iquip/apps/builder.py
```diff
@@ -400,52 +400,52 @@ def __init__(
...
             procdesc: Each key and its value are as follows.
               unit: The unit of the number value.
-              scale: See scale docstring at class Attributes.
+              scale: See scale docstring at attributes section.
```
BECATRUE: "scale: See the attributes section."

iquip/apps/builder.py
```diff
@@ -400,52 +400,52 @@ def __init__(
...
               global_step: The step between values changed by the up and down button.
-              global_min: The minimum value. (default=float("-inf"))
-              global_max: The maximum value. (default=float("inf"))
+              global_min: The minimum value. (default=0.0)
+              global_max: The maximum value. (default=99.99)
                 If min > max, then they are swapped.
               ndecimals: The number of displayed decimals.
-            state: See the state docstring at class Attributes.
+            state: See the state docstring at attributes section.
```
BECATRUE: "state: See the attributes section."

iquip/apps/builder.py
```diff
@@ -400,52 +400,52 @@ def __init__(
         ):
         """Extended.
 
-        Args: See the docstring of Attributes for more detail.
+        Args: See the attributes section for more details.
             procdesc: Each key and its value are as follows.
```
BECATRUE: I noticed that the docstrings of all items are duplicated with `_ScanEntry.__init__()`.
So, you should remove the duplicated ones.

Also, please match the docstrings to the existing format exactly.

iquip/apps/builder.py
```diff
@@ -400,52 +400,52 @@ def __init__(
...
+        self.randomizeCheckBox.stateChanged.connect(self.update_randomize)
 
     def apply_properties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
-        """Adds properties to the SpinBox widget.
+        """Adds properties to the spin box widget.
 
         Attributes:
             widget: A QDoubleSpinWidget that has properties to set.
-            procdesc: See the docstring at __init__ for more detail.
+            procdesc: See the __init__ section.
```
BECATRUE: "procdesc: See procdesc in __init__().

BECATRUE (changes requested):
Please check the previous review about those four slots.

config_template.json
```diff
@@ -80,4 +80,4 @@
         "icon_path": "resources/icon.jpg",
         "background_path": "resources/background.jpg"
     }
-}
+}
```
BECATRUE: Could you check why there is a difference?

iquip/apps/builder.py
```diff
@@ -322,26 +314,24 @@ def __init__(
               unit: The unit of the number value.
               scale: The scale factor that is multiplied to the number value.
               global_step: The step between values changed by the up and down button.
-              global_min: The minimum value. (default=float("-inf"))
-              global_max: The maximum value. (default=float("inf"))
+              global_min: The minimum value. (default=0)
```
BECATRUE: Please write "0.0" to indicate that its type is `float`.

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+
+class _RangeScan(QWidget):
+    """A widget for range scan in _ScanEntry.
+
+    Attributes:
+        scale: The scale factor that is multiplied to the number value.
+        state: Each key and its value as follows.
+          start: The start point for the RangeScan sequence.
+          stop: The end point for the RangeScan sequence.
+          npoints: The number of points in the RangeScan squence.
```
BECATRUE: "sequence"

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self,
+        procdesc: Dict[str, Any],
+        state: Dict[str, Any],
+        parent: Optional[QWidget] = None
+        ):
+        """Extended.
 
         Args:
-            argInfo: See argInfo in __init__().
+            procdesc: keys are as follows.
```
BECATRUE: Like other dictionaries, please write "Each key and its value as follows.".

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+              unit, scale, global_step, global_min, global_max, ndecimals: 
+                See the argInfo at init() in _ScanEntry class.
```
BECATRUE: `See argInfo in _ScanEntry.__init__().`

Please check if the method name is correct.

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self.layout.addWidget(self.stopSpinBox, 2, 1)
+        self.layout.addWidget(self.randomizeCheckBox, 3, 1)
+        self.startSpinBox.valueChanged.connect(self.update_start)
+
+    def apply_properties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
+        """Adds properties to the spin box widget.
+
+        Attributes:
+            widget: A QDoubleSpinWidget that has properties to set.
+            procdesc: See the procdesc in init().
```
BECATRUE: Please remove "the".

And I'm sorry for the previous review.
> "procdesc: See procdesc in __init__()."

Due to the markdown syntax, I intended `__init__()`, but it is shown to `init()`.
Please fix it.

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        self.randomizeCheckBox.setChecked(state["randomize"])
+        # layout
+        self.layout.addWidget(QLabel("start:", self), 0, 0)
+        self.layout.addWidget(self.startSpinBox, 0, 1)
+        self.layout.addWidget(QLabel("npoints:", self), 1, 0)
+        self.layout.addWidget(self.npointsSpinBox, 1, 1)
+        self.layout.addWidget(QLabel("stop:", self), 2, 0)
+        self.layout.addWidget(self.stopSpinBox, 2, 1)
+        self.layout.addWidget(self.randomizeCheckBox, 3, 1)
+        self.startSpinBox.valueChanged.connect(self.update_start)
```
BECATRUE: Please remove this, too.

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+            minVal = 0.0
+        if maxVal is None:
+            maxVal = 99.99
+        widget.setMinimum(minVal / self.scale)
+        widget.setMaximum(maxVal / self.scale)
+        widget.setSuffix(unit)
+        widget.setSingleStep(step / self.scale)
+
+    def get_scan_args(self) -> Dict[str, Any]:
+        """Returns argument of range scan."""
```
BECATRUE: Returns the arguments of the range scan.

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        scanArgs = {
+        "start": self.startSpinBox.value(),
```
BECATRUE: Please add an indentation.

BECATRUE: And, I think returning directly is okay, instead of defining a variable.

iquip/apps/builder.py
```diff
@@ -365,21 +355,112 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        selected = self.state["selected"]
+        return self.state[selected]
+
+
+class _RangeScan(QWidget):
+    """A widget for range scan in _ScanEntry.
+
+    Attributes:
+        scale: The scale factor that is multiplied to the number value.
+        state: Each key and its value as follows.
```
BECATRUE: As I understand, we don't have to make `state` an attribute anymore, because we won't track it.
What do you think of it?

BECATRUE (changes requested):

config_template.json
Aijuh: I just copy and pasted file at original branch. I also don't know.
Since you updated something at config_template, I updated it.

iquip/apps/builder.py
Aijuh: I should have checked other cases.

iquip/apps/builder.py
Aijuh: I added it to test the code at last step and forget to remove it..

iquip/apps/builder.py
Aijuh: correct.

config_template.json
BECATRUE: I updated that config file in "develop" branch, hence you must not copy it but should revert this file to the initial state of your branch.

In other words, there should be no change in this PR.

BECATRUE: Or you may fix it by manually checking this "Files changed" tab.

iquip/apps/builder.py
BECATRUE: I'm really sorry.. Grammatically, "Each key and its value are as follows." is correct.

iquip/apps/builder.py
```diff
@@ -365,21 +355,108 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        state: Dict[str, Any],
+        parent: Optional[QWidget] = None
+        ):
+        """Extended.
 
         Args:
-            argInfo: See argInfo in __init__().
+            procdesc: Each key and its value as follows.
+              unit, scale, global_step, global_min, global_max, ndecimals: 
+                See argInfo at _ScanEntry.__init__().
```
BECATRUE: Do you have any reason to use "at" instead of "in"?

iquip/apps/builder.py
```diff
@@ -645,6 +722,23 @@ def argumentsFromListWidget(self, listWidget: QListWidget) -> Dict[str, Any]:
             args[widget.name] = widget.value()
         return args
 
+    def scannableFromListWidget(self, listWidget: QListWidget) -> Dict[str, Any]:
```
BECATRUE: Why did you define another method, not using `argumentsFromListWidget`?
They look identical.

iquip/apps/builder.py
```diff
@@ -653,7 +747,9 @@ def submit(self):
         """
         try:
             experimentArgs = self.argumentsFromListWidget(self.builderFrame.argsListWidget)
+            scanArgs = self.scannableFromListWidget(self.builderFrame.scanListWidget)
             schedOpts = self.argumentsFromListWidget(self.builderFrame.schedOptsListWidget)
+            experimentArgs.update(scanArgs)
```
BECATRUE: A `update()` never causes a `ValueError`, thus please move it out of the `try-except` statement.

BECATRUE (changes requested):
Almost done!

iquip/apps/builder.py
Aijuh: No reason..

Aijuh:
When describing the dictionary, I changed the format to `Each key and its value are:`.

iquip/apps/builder.py
```diff
@@ -632,7 +709,7 @@ def argumentsFromListWidget(self, listWidget: QListWidget) -> Dict[str, Any]:
         """Gets arguments from the given list widget and returns them.
         
         Args:
-            listWidget: The QListWidget containing _BaseEntry instances.
+            listWidget: The QListWidget containing _BaseEntry instances or _ScanEntry instances.
```
BECATRUE: `_ScanEntry` is a subclass of `_BaseEntry`, thus please remove it.

iquip/apps/builder.py
BECATRUE: Please check this.

iquip/apps/builder.py
```diff
@@ -365,21 +355,108 @@ def get_state(self, argInfo: Dict[str, Any]) -> Dict[str, Any]:
...
+        """
+        selected = self.state["selected"]
+        return self.state[selected]
+
+
+class _RangeScan(QWidget):
+    """A widget for range scan in _ScanEntry.
+
+    Attributes:
+        scale: The scale factor that is multiplied to the number value.
```
BECATRUE: This docstring is also duplicated.

BECATRUE (changes requested):

BECATRUE (approved):
👍

PR #255, opened 2024-03-05
Implement NoScan and add BaseScan

### LINKED ISSUE #256: Add NoScan Entry at builder gui
First, add a NoScan class. 
Before, the builder GUI only supported RangeScan. When NoScan class is added, feature for selecting the type of scan we want is needed. So this feature will also be added.
Since Scan classes share common method and some variables, it seems that base class for all Scan classes is needed.

1. NoScan has two important parameters, `value` and `repetitions`. Use `QDoubleSpinBox` for control `value`, and `QSpinBox` for `repetitions`. It has similar structure with `RangeScan` class.
2. Since we added NoScan class, ScanEntry class should be updated. Use `QStackedWidget` to store each scan type entry. And Use `QRadioButton` to make selection button gui. All radio buttons are added at `QHBoxLayout`. If one of button is checked by user, method `scanTypeToggled` is called and update current widget at ScanEntry. Original iquip code used `QButtonGroup`, but as I checked there was no reason to use it.
3. Add BaseScan class that is the base class for Scan classes. Since all Scan classes use `applyProperties` method to initialize there widget, BaseScan class can have that method.

Aijuh:
This closes #256
##
1. `_NoScan` class has been added. It has two arguments, `value` and `repetitions`. 
2. Implment scan type selection in the `_ScanEntry` class. You can choose RangeScan or NoScan at our iquip GUI. Two widgets are contained at `stackWidget`. `radioButtons` is used to select between two scan type.
3. Added BaseScan that is base class for all scan widgets. Since some attributes and method are duplicated, it seems more efficient way.
##
Below is the image of current figure of builder GUI.
image

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -9,8 +9,8 @@ import requests
 from PyQt5.QtCore import QDateTime, QObject, Qt, QThread, pyqtSignal, pyqtSlot
 from PyQt5.QtWidgets import (
     QAbstractButton, QButtonGroup, QCheckBox, QComboBox, QDateTimeEdit, QDoubleSpinBox, QGridLayout,
-    QHBoxLayout, QLabel, QLineEdit, QListWidget, QListWidgetItem, QPushButton, QRadioButton, QSpinBox,
-    QStackedWidget, QVBoxLayout, QWidget
+    QHBoxLayout, QLabel, QLineEdit, QListWidget, QListWidgetItem, QPushButton, QRadioButton,
+    QSpinBox, QStackedWidget, QVBoxLayout, QWidget
 )
 
 import qiwis
@@ -307,6 +307,7 @@ class _ScanEntry(_BaseEntry):
         scanWidgets: The dictionary that contains each scan widget.
     """
     @unique
+    # pylint: disable=invalid-name
     class ScanType(IntEnum):
         """Enum class for mapping id to each scannable type."""
         NoScan = 0
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -2,15 +2,14 @@
 
 import json
 import logging
-from collections import OrderedDict
 from typing import Any, Dict, Optional, Tuple, Union
 
 import requests
 from PyQt5.QtCore import QDateTime, QObject, Qt, QThread, pyqtSignal, pyqtSlot
 from PyQt5.QtWidgets import (
-    QCheckBox, QComboBox, QDateTimeEdit, QDoubleSpinBox, QGridLayout, QHBoxLayout, QLabel,
-    QLineEdit, QListWidget, QListWidgetItem, QPushButton, QRadioButton, QSpinBox, QStackedWidget,
-    QVBoxLayout, QWidget
+    QButtonGroup, QCheckBox, QComboBox, QDateTimeEdit, QDoubleSpinBox, QGridLayout, QHBoxLayout,
+    QLabel, QLineEdit, QListWidget, QListWidgetItem, QPushButton, QRadioButton, QSpinBox,
+    QStackedWidget, QVBoxLayout, QWidget
 )
 
 import qiwis
@@ -302,10 +301,13 @@ class _ScanEntry(_BaseEntry):
           "selected": The name of the selected scannable type.
           "NoScan", "RangeScan", "CenterScan", and "ExplicitScan": The dictionary that contains 
             argument info of the corresponding scannable type.
-        stackWidget: The QStackedWidget that contains widgets of each scannable type.
-        rangeWidget: TODO(AIJUH): The Widget that will be removed at next PR.
+        stackedWidget: The QStackedWidget that contains widgets of each scannable type.
+        scanButtonGroup: The QButtonGroup that groups QRadiobuttons for selecting scan widget.
+        scanWidgetsDict: The dictionary that contains each scan widget.
+        scanButtonsDict: The dictionary that describes id of QRadiobutton at scanButtonGroup.
     """
     def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget] = None):
+        # pylint: disable=too-many-arguments
         """Extended.
 
         Args:
@@ -321,30 +323,33 @@ class _ScanEntry(_BaseEntry):
               ndecimals: The number of displayed decimals.
         """
         super().__init__(name, argInfo, parent=parent)
-        self.state = self.getState()
-        procdesc = self.getProcdesc()
-        self.stackWidget = QStackedWidget(self)
-        self.widgets = OrderedDict()
-        self.widgets["NoScan"] = _NoScan(procdesc, self.state["NoScan"])
-        self.widgets["RangeScan"] = _RangeScan(procdesc, self.state["RangeScan"])
-        for widget in self.widgets.values():
-            self.stackWidget.addWidget(widget)
-        self.radioButtons = OrderedDict()
-        self.radioButtons["NoScan"] = QRadioButton("No scan")
-        self.radioButtons["RangeScan"] = QRadioButton("Range")
+        self.state = self.stateManager()
+        procdesc = self.procdescManager()
+        self.stackedWidget = QStackedWidget(self)
         buttonLayout = QHBoxLayout()
-        for _, b in enumerate(self.radioButtons.values()):
-            buttonLayout.addWidget(b)
-            b.toggled.connect(self.scanTypeToggled)
+        self.scanButtonGroup = QButtonGroup(self)
+        scanDict = {"NoScan": (_NoScan, "No scan"), "RangeScan": (_RangeScan, "Range")}
+        self.scanWidgetsDict = {}
+        self.scanButtonsDict = {}
+        for buttonId, (ty, (scanCls, buttonName)) in enumerate(scanDict.items()):
+            scanWidget = scanCls(procdesc, self.state[ty])
+            self.scanWidgetsDict[ty] = scanWidget
+            self.stackedWidget.addWidget(scanWidget)
+            button = QRadioButton(buttonName)
+            self.scanButtonsDict[ty] = buttonId
+            buttonLayout.addWidget(button)
+            self.scanButtonGroup.addButton(button, buttonId)
+        self.scanButtonGroup.buttonToggled.connect(self.scanTypeToggled)
         selected = self.state["selected"]
-        self.radioButtons[selected].setChecked(True)
+        selectedButton = self.scanButtonGroup.button(self.scanButtonsDict[selected])
+        selectedButton.setChecked(True)
         # layout
         scanLayout = QVBoxLayout()
         scanLayout.addLayout(buttonLayout)
-        scanLayout.addWidget(self.stackWidget)
+        scanLayout.addWidget(self.stackedWidget)
         self.layout.addLayout(scanLayout)
 
-    def getState(self) -> Dict[str, Any]:
+    def stateManager(self) -> Dict[str, Any]:
         """Gets a dictionary that describes default parameters of all scannable types."""
         scale = self.argInfo["scale"]
         state = {
@@ -370,7 +375,7 @@ class _ScanEntry(_BaseEntry):
                     state[ty] = default
         return state
 
-    def getProcdesc(self) -> Dict[str, Any]:
+    def procdescManager(self) -> Dict[str, Any]:
         """Gets a procdesc dictionary that describes common parameters of the scannable object."""
         procdesc = {
             "unit": self.argInfo["unit"],
@@ -395,42 +400,44 @@ class _ScanEntry(_BaseEntry):
         
         Once the checked button at radiobutton group changed, this is called.
         """
-        for ty, button in self.radioButtons.items():
+        for ty, buttonId in self.scanButtonsDict.items():
+            button = self.scanButtonGroup.button(buttonId)
             if button.isChecked():
-                self.stackWidget.setCurrentWidget(self.widgets[ty])
+                self.stackedWidget.setCurrentWidget(self.scanWidgetsDict[ty])
                 self.state["selected"] = ty
                 break
 
 
 class _BaseScan(QWidget):
-    """Base class for all scannable widget.
+    """Base class for all scannable widgets.
 
     Attributes:
         scale: See argInfo in _ScanEntry.__init__().
         layout: The outermost layout.
+        procdesc: Each key and its value are:
+          unit, scale, global_step, global_min, global_max, ndecimals: 
+            See argInfo in _ScanEntry.__init__().
     """
     def __init__(self, procdesc: Dict[str, Any], parent: Optional[QWidget] = None):
         """Extended.
 
         Args:
-            procdesc: Each key and its value are:
-              unit, scale, global_step, global_min, global_max, ndecimals: 
-                See argInfo in _ScanEntry.__init__().
+            procdesc: See procdesc in __init__().
         """
         super().__init__(parent=parent)
         self.scale = procdesc["scale"]
         self.layout = QGridLayout(self)
+        self.procdesc = procdesc
 
-    def applyProperties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
+    def applyProperties(self, widget: QDoubleSpinBox):
         """Adds properties to the spin box widget.
 
         Attributes:
             widget: A QDoubleSpinWidget that has properties to set.
-            procdesc: See procdesc in __init__().
         """
-        ndecimals, minVal, maxVal, step, unit = map(procdesc.get, ("ndecimals",
-                                                                   "global_min", "global_max",
-                                                                   "global_step", "unit"))
+        ndecimals, minVal, maxVal, step, unit = map(self.procdesc.get, ("ndecimals",
+                                                                        "global_min", "global_max",
+                                                                        "global_step", "unit"))
         widget.setDecimals(ndecimals)
         if minVal is None:
             minVal = 0.0
@@ -441,7 +448,7 @@ class _BaseScan(QWidget):
         widget.setSuffix(unit)
         widget.setSingleStep(step / self.scale)
 
-    def getScanArgs(self) -> Dict[str, Any]:
+    def scanArguments(self) -> Dict[str, Any]:
         """Returns the arguments of the scannable object.
         
         This must be overridden in the subclass.
@@ -450,7 +457,7 @@ class _BaseScan(QWidget):
 
 
 class _NoScan(_BaseScan):
-    """An widget for no scan in _ScanEntry.
+    """Widget for no scan in _ScanEntry.
 
     Attributes:
         valueSpinBox: QDoubleSpinBox for value argument inside state.
@@ -465,26 +472,25 @@ class _NoScan(_BaseScan):
         """Extended.
 
         Args:
-            procdesc: See the __init__() in _BaseScan.
-            state:
-              valueSpinBox: The repeated value in the NoScan sequence.
-              repetitionsSpinBox: A number to repeat the value in the NoScan sequence.
+            state: Each key and its value are:
+              value: The repeated value in the NoScan sequence.
+              repetitions: The number to repeat the value in the NoScan sequence.
         """
         super().__init__(procdesc, parent=parent)
         self.valueSpinBox = QDoubleSpinBox(self)
-        self.applyProperties(self.valueSpinBox, procdesc)
-        self.valueSpinBox.setValue(state["value"]/self.scale)
+        self.applyProperties(self.valueSpinBox)
+        self.valueSpinBox.setValue(state["value"] / self.scale)
         repetitionsSpinBox = QSpinBox(self)
         repetitionsSpinBox.setMinimum(1)
         repetitionsSpinBox.setMaximum((1 << 31) - 1)
         repetitionsSpinBox.setValue(state["repetitions"])
-        #layout
+        # layout
         self.layout.addWidget(QLabel("Value:", self), 0, 0)
         self.layout.addWidget(self.valueSpinBox, 0, 1)
         self.layout.addWidget(QLabel("Repetitions:"), 1, 0)
         self.layout.addWidget(repetitionsSpinBox, 1, 1)
 
-    def getScanArgs(self) -> Dict[str, Any]:
+    def scanArguments(self) -> Dict[str, Any]:
         """Overridden.
         
         Returns the arguments of the no scan.
@@ -513,23 +519,22 @@ class _RangeScan(_BaseScan):
         """Extended.
 
         Args:
-            procdesc: See the __init__() in _BaseScan.
             state: Each key and its value are:
               start: The start point for the RangeScan sequence.
               stop: The end point for the RangeScan sequence.
               npoints: The number of points in the RangeScan sequence.
-              randomize: A boolean value that decides whether to shuffle the RangeScan sequence.
+              randomize: The boolean value that decides whether to shuffle the RangeScan sequence.
         """
         super().__init__(procdesc, parent=parent)
         self.startSpinBox = QDoubleSpinBox(self)
-        self.applyProperties(self.startSpinBox, procdesc)
+        self.applyProperties(self.startSpinBox)
         self.startSpinBox.setValue(state["start"] / self.scale)
         self.npointsSpinBox = QSpinBox(self)
         self.npointsSpinBox.setMinimum(1)
         self.npointsSpinBox.setMaximum((1 << 31) - 1)
         self.npointsSpinBox.setValue(state["npoints"])
         self.stopSpinBox = QDoubleSpinBox(self)
-        self.applyProperties(self.stopSpinBox, procdesc)
+        self.applyProperties(self.stopSpinBox)
         self.stopSpinBox.setValue(state["stop"] / self.scale)
         self.randomizeCheckBox = QCheckBox("Randomize", self)
         self.randomizeCheckBox.setChecked(state["randomize"])
@@ -542,7 +547,7 @@ class _RangeScan(_BaseScan):
         self.layout.addWidget(self.stopSpinBox, 2, 1)
         self.layout.addWidget(self.randomizeCheckBox, 3, 1)
 
-    def getScanArgs(self) -> Dict[str, Any]:
+    def scanArguments(self) -> Dict[str, Any]:
         """Overridden.
         
         Returns the arguments of the range scan.
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -308,7 +308,7 @@ class _ScanEntry(_BaseEntry):
     """
     def __init__(
         self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget] = None
-        ): # pylint: disable=too-many-arguments
+        ): # pylint: disable=too-many-locals
         """Extended.
 
         Args:
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -331,16 +331,19 @@ class _ScanEntry(_BaseEntry):
         scanDict = {"NoScan": (_NoScan, "No scan"), "RangeScan": (_RangeScan, "Range")}
         self.scanWidgets = {}
         self.scanButtonsDict = {}
+        self.idToButton = {}
         for buttonId, (ty, (scanCls, buttonName)) in enumerate(scanDict.items()):
             scanWidget = scanCls(procdesc, self.state[ty])
             self.scanWidgets[ty] = scanWidget
             self.stackedWidget.addWidget(scanWidget)
             button = QRadioButton(buttonName)
             self.scanButtonsDict[ty] = buttonId
+            self.idToButton[buttonId] = ty
             buttonLayout.addWidget(button)
             self.scanButtonGroup.addButton(button, buttonId)
-        self.scanButtonGroup.buttonToggled.connect(self.scanTypeToggled)
+        self.scanButtonGroup.buttonClicked.connect(self.scanTypeClicked)
         selected = self.state["selected"]
+        self.stackedWidget.setCurrentWidget(self.scanWidgets[selected])
         selectedButton = self.scanButtonGroup.button(self.scanButtonsDict[selected])
         selectedButton.setChecked(True)
         # layout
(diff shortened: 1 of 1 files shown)


BECATRUE:
Can you indicate the relative issue in PR contents, like "This is related to #{issue}."?

BECATRUE:
Then, you should write how to implement it in as possible as detailed in the issue, before implementing it.
From the next PR, please keep this in your mind.

iquip/apps/builder.py
```diff
@@ -322,11 +323,27 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
         super().__init__(name, argInfo, parent=parent)
         self.state = self.get_state()
         procdesc = self.get_procdesc()
-        # TODO(AIJUH): Add features for stack widget.
         self.stackWidget = QStackedWidget(self)
```
BECATRUE: This is just a suggestion.
As the class name is `QStackedWidget`, how about renaming it to `stackedWidget`?

If you prefer to the original one, don't keep in your mind.

BECATRUE:
I'm sorry that I couldn't check that methods and variables related to PyQt should have names with camel case, due to PyQt customary practice.

Please rename them implemented in the previous and this PRs.

iquip/apps/builder.py
```diff
@@ -322,11 +323,27 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
...
+            self.stackWidget.addWidget(widget)
+        self.radioButtons = OrderedDict()
+        self.radioButtons["NoScan"] = QRadioButton("No scan")
+        self.radioButtons["RangeScan"] = QRadioButton("Range")
+        scan_type = QButtonGroup(self)
+        buttonLayout = QHBoxLayout()
+        for _, b in enumerate(self.radioButtons.values()):
+            buttonLayout.addWidget(b)
+            scan_type.addButton(b)
+            b.toggled.connect(self.scan_type_toggled)
```
BECATRUE: Umm.. I think you can write this code in only one loop.

For example,
```python
self.widgets = {}
scanButtonGroup = QButtonGroup(self)
buttonLayout = QHBoxLayout()
for ty, (scanCls, buttonName) in {"NoScan": (_NoScan, "No scan"), ...}.items():
  scanWidget = scanCls(procdesc, self.state[ty])
  self.widgets[ty] = scanWidget
  self.stackWidget.addWidget(scanWidget)
  button = QRadioButton(buttonName)
  scanButtonGroup.addButton(button, id)  # id is defined in IntEnum class.
scanButtonGroup.buttonToggled.connect(...)  # this is called whenever any button is toggled.
```

BECATRUE: I think it has some advantages:
1. `OrderedDict` is unnecessary.
2. Easy to manage each widget.
3. Avoid to cause an error due to a typo.

BECATRUE: And please check the followings:
1. The name of `widgets` is too vague. How about "scanWidgets"?
2. If you use an id and enum class to identify which button is toggled, it seems that `radioButtons` is unnecessary.

Aijuh:
I understood. I will make new issue and convert to camel case.

Aijuh:
@BECATRUE  I updated it.

BECATRUE:
> I understood. I will make new issue and convert to camel case.

You already fixed it and won't make a new issue, right?

iquip/apps/builder.py
```diff
@@ -320,16 +321,30 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
...
+        self.state = self.getState()
+        procdesc = self.getProcdesc()
         self.stackWidget = QStackedWidget(self)
-        self.rangeWidget = _RangeScan(procdesc, self.state["RangeScan"])
+        self.widgets = OrderedDict()
+        self.widgets["NoScan"] = _NoScan(procdesc, self.state["NoScan"])
+        self.widgets["RangeScan"] = _RangeScan(procdesc, self.state["RangeScan"])
+        for widget in self.widgets.values():
+            self.stackWidget.addWidget(widget)
+        self.radioButtons = OrderedDict()
```
BECATRUE: Why did you remove the button group?
For radio buttons to operate in the same group, isn't it necessary?

iquip/apps/builder.py
```diff
@@ -320,16 +321,30 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
...
+        self.radioButtons["NoScan"] = QRadioButton("No scan")
+        self.radioButtons["RangeScan"] = QRadioButton("Range")
+        buttonLayout = QHBoxLayout()
+        for _, b in enumerate(self.radioButtons.values()):
+            buttonLayout.addWidget(b)
+            b.toggled.connect(self.scanTypeToggled)
```
BECATRUE: This code can be simplified.

For example,
```python
# in this class
class ScanType(IntEnum):
  NoScan = 0
  RangeScan = 1

# in this method
self.widgets = {}
buttonGroup = QButtonGroup(self)
buttonLayout = QHBoxLayout()
for ty, (scanCls, buttonName) in {"NoScan": (_NoScan, "No scan"), ...}.items()
  scanWidget = scanCls(procdesc, self.state[ty])
  self.widgets[ty] = scanWidget
  self.stackWidget.addWidget(scanWidget)
  button = QRadioButton(buttonName)
  buttonGroup.addButton(button, _ScanEntry.ScanType[ty].value)
  buttonLayout.addWidget(button)
buttonGroup.buttonClicked.connect(...)
```

I think it has some advantages:
1. Shorten the code
2. Easy to change the order of the scan widgets.
3. Make the management easy and avoid to cause an error due to typos.
4. Use only one signal, i.e. `buttonGroup.buttonClicked`.

Of course, the code may have some errors, hence please recheck it.

BECATRUE: And, please check the followings:
1. The name of `self.widgets` looks too vague. How about `self.scanWidgets`?
2. Is `self.radioButtons` necessary?

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+            if button.isChecked():
+                self.stackWidget.setCurrentWidget(self.widgets[ty])
+                self.state["selected"] = ty
+                break
+
 
-class _RangeScan(QWidget):
-    """A widget for range scan in _ScanEntry.
+class _BaseScan(QWidget):
+    """Base class for all scannable widget.
```
BECATRUE: "a scannable widget" or "all scannble widgets".

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        """
+        super().__init__(procdesc, parent=parent)
+        self.valueSpinBox = QDoubleSpinBox(self)
+        self.applyProperties(self.valueSpinBox, procdesc)
+        self.valueSpinBox.setValue(state["value"]/self.scale)
+        repetitionsSpinBox = QSpinBox(self)
+        repetitionsSpinBox.setMinimum(1)
+        repetitionsSpinBox.setMaximum((1 << 31) - 1)
+        repetitionsSpinBox.setValue(state["repetitions"])
+        #layout
```
BECATRUE: "# layout"

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        Args:
+            procdesc: Each key and its value are:
+              unit, scale, global_step, global_min, global_max, ndecimals: 
+                See argInfo in _ScanEntry.__init__().
+        """
+        super().__init__(parent=parent)
+        self.scale = procdesc["scale"]
+        self.layout = QGridLayout(self)
+
+    def applyProperties(self, widget: QDoubleSpinBox, procdesc: Dict[str, Any]):
```
BECATRUE: You moved this method to the super class, so it looks better to make `procdesc` an attribute.
Then, you don't need to call this method with it.

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        if minVal is None:
+            minVal = 0.0
+        if maxVal is None:
+            maxVal = 99.99
+        widget.setMinimum(minVal / self.scale)
+        widget.setMaximum(maxVal / self.scale)
+        widget.setSuffix(unit)
+        widget.setSingleStep(step / self.scale)
+
+    def getScanArgs(self) -> Dict[str, Any]:
```
BECATRUE: I remembered that Jiyong said not to start a method name with "get", according to the PyQt rule.
So, can you rename these three methods, `getState()`, `getProcdesc()`, and `getScanArgs`.

For example, use `scanArguments()` instead of it.
(I think `scanArguments` is better that `scanArgs` because we have used it as the full name.)

However, I'm worried to rename `getState()` to `state()` because it already exists as an attribute.
Hence you need to come up with a better idea.

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        """Returns the arguments of the scannable object.
+        
+        This must be overridden in the subclass.
+        """
+        raise NotImplementedError
+
+
+class _NoScan(_BaseScan):
+    """An widget for no scan in _ScanEntry.
```
BECATRUE: "A widget". But, I think "a" is unnecessary like other classes.

iquip/apps/builder.py
```diff
@@ -396,27 +513,23 @@ def __init__(
         """Extended.
 
         Args:
-            procdesc: Each key and its value are:
-              unit, scale, global_step, global_min, global_max, ndecimals: 
-                See argInfo in _ScanEntry.__init__().
+            procdesc: See the __init__() in _BaseScan.
```
BECATRUE: As I think, any explanation about `procdesc` is unnecessary becuase "Extended." indicates that there are more explanations in `_BaseScan.__init__()`.

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        self,
+        procdesc: Dict[str, Any],
+        state: Dict[str, Any],
+        parent: Optional[QWidget] = None
+        ):
+        """Extended.
+
+        Args:
+            procdesc: See the __init__() in _BaseScan.
+            state:
```
BECATRUE: Please add "Each key and its value are:".
And the name of keys looks wrong.

iquip/apps/builder.py
```diff
@@ -396,27 +513,23 @@ def __init__(
...
         Args:
-            procdesc: Each key and its value are:
-              unit, scale, global_step, global_min, global_max, ndecimals: 
-                See argInfo in _ScanEntry.__init__().
+            procdesc: See the __init__() in _BaseScan.
             state: Each key and its value are:
               start: The start point for the RangeScan sequence.
               stop: The end point for the RangeScan sequence.
               npoints: The number of points in the RangeScan sequence.
               randomize: A boolean value that decides whether to shuffle the RangeScan sequence.
```
BECATRUE: Can you start with "The" like other keys.

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        state: Dict[str, Any],
+        parent: Optional[QWidget] = None
+        ):
+        """Extended.
+
+        Args:
+            procdesc: See the __init__() in _BaseScan.
+            state:
+              valueSpinBox: The repeated value in the NoScan sequence.
+              repetitionsSpinBox: A number to repeat the value in the NoScan sequence.
```
BECATRUE: Can you start with "The" like the other.

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        """
+        super().__init__(procdesc, parent=parent)
+        self.valueSpinBox = QDoubleSpinBox(self)
+        self.applyProperties(self.valueSpinBox, procdesc)
+        self.valueSpinBox.setValue(state["value"]/self.scale)
```
BECATRUE: Please add a space before and after the operator.

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        repetitionsSpinBox = QSpinBox(self)
```
BECATRUE: Did you intend "self.repetitionsSpinBox"?

iquip/apps/builder.py
```diff
@@ -375,17 +390,119 @@ def value(self) -> Dict[str, Any]:
...
+        self.layout.addWidget(repetitionsSpinBox, 1, 1)
+
+    def getScanArgs(self) -> Dict[str, Any]:
+        """Overridden.
+        
+        Returns the arguments of the no scan.
+        """
+        return {
+            "value": self.valueSpinBox.value(),
+            "repetitions": self.repetitionsSpinBox.value()
```
BECATRUE, line 512: Does it work well..? I think there is no `self.repetitionsSpinBox`.

BECATRUE (changes requested):

iquip/apps/builder.py
Aijuh: I remembered that you send me the link for rules for coding pyQt. But I cannot find it.
Can you send the link if you have it?
If not, you don't have to. Of course I will fix this case.

iquip/apps/builder.py
Aijuh: In previous code it works well without button group. Since we operate on QRadioButton seperately.

iquip/apps/builder.py
Aijuh: I will think in this way.

iquip/apps/builder.py
Aijuh: After I finished testing, I found that the name of Qspinbox for `repetitions` is just repetitions.
There was mistake while I updating it.

iquip/apps/builder.py
Aijuh: Yes...

iquip/apps/builder.py
Aijuh: I didn't find it.

iquip/apps/builder.py
Aijuh: I see.

iquip/apps/builder.py
Aijuh: Thank for giving entire code.

iquip/apps/builder.py
BECATRUE: It looks natural that they are tied by a button group.
However, you already fixed to use it, thus please never mind.

iquip/apps/builder.py
```diff
@@ -301,10 +301,14 @@ class _ScanEntry(_BaseEntry):
...
-        rangeWidget: TODO(AIJUH): The Widget that will be removed at next PR.
+        stackedWidget: The QStackedWidget that contains widgets of each scannable type.
+        scanButtonGroup: The QButtonGroup that groups QRadiobuttons for selecting scan widget.
+        scanWidgetsDict: The dictionary that contains each scan widget.
+        scanButtonsDict: The dictionary that describes id of QRadiobutton at scanButtonGroup.
     """
-    def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget] = None):
+    def __init__(
+        self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget] = None
+        ): # pylint: disable=too-many-locals
```
BECATRUE: I think this Pylint comment can be moved to the definition above.
And the definition can be written in a line.

iquip/apps/builder.py
BECATRUE: The reason I suggested to use an enum class is that it looks inefficient to check all buttons in `scanTypeToggled()`.
According to [QButtonGroup](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QButtonGroup.html#detailed-description), the `buttonClicked` signal is emitted with id of the clicked button, so you can make code more efficiently.

For example,
```python
def scanButtonClicked(self, id: int):
  ty = _ScanEntry.ScanType(id).name
  stackedWidget = self.scanWidgetsDict[ty]
  ...
```

iquip/apps/builder.py
```diff
@@ -320,16 +324,33 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
...
-        # TODO(AIJUH): Add features for stack widget.
-        self.stackWidget = QStackedWidget(self)
-        self.rangeWidget = _RangeScan(procdesc, self.state["RangeScan"])
+        self.state = self.stateManager()
+        procdesc = self.procdescManager()
+        self.stackedWidget = QStackedWidget(self)
+        buttonLayout = QHBoxLayout()
+        self.scanButtonGroup = QButtonGroup(self)
+        scanDict = {"NoScan": (_NoScan, "No scan"), "RangeScan": (_RangeScan, "Range")}
+        self.scanWidgetsDict = {}
```
BECATRUE: It looks sufficient to name it to `scanWidgets`.
In other apps, we have hardly indicated "Dict".

iquip/apps/builder.py
```diff
@@ -320,16 +324,33 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
...
+        selectedButton.setChecked(True)
+        # layout
         scanLayout = QVBoxLayout()
-        scanLayout.addWidget(self.rangeWidget)
+        scanLayout.addLayout(buttonLayout)
+        scanLayout.addWidget(self.stackedWidget)
         self.layout.addLayout(scanLayout)
 
-    def get_state(self) -> Dict[str, Any]:
+    def stateManager(self) -> Dict[str, Any]:
```
BECATRUE: First, we have usually named a method to start with a verb, except to a slot.
Also `*Manager` feels like it means a manager object.

Then, how about just `initState()` or something better?

iquip/apps/builder.py
```diff
@@ -375,17 +396,120 @@ def value(self) -> Dict[str, Any]:
...
+        layout: The outermost layout.
+        procdesc: Each key and its value are:
+          unit, scale, global_step, global_min, global_max, ndecimals: 
+            See argInfo in _ScanEntry.__init__().
+    """
+    def __init__(self, procdesc: Dict[str, Any], parent: Optional[QWidget] = None):
+        """Extended.
+
+        Args:
+            procdesc: See procdesc in __init__().
```
BECATRUE: "procdesc: See the attributes section." or
"See the attributes section."

iquip/apps/builder.py
BECATRUE: Umm.. I don't remember that I sent you about this😅
About Python coding style, you can refer to this [styleguide](https://google.github.io/styleguide/pyguide.html), but I don't know about PyQt coding style..

iquip/apps/builder.py
```diff
@@ -375,17 +396,120 @@ def value(self) -> Dict[str, Any]:
...
+        Returns the arguments of the no scan.
+        """
+        return {
+            "value": self.valueSpinBox.value(),
+            "repetitions": self.repetitionsSpinBox.value()
+        }
+
+
+class _RangeScan(_BaseScan):
+    """A widget for range scan in _ScanEntry.
```
BECATRUE: Please don't write "A" like `_NoScan`.

iquip/apps/builder.py
BECATRUE: Did you apply this review?

Aijuh:
Yes I will apply your current review.

iquip/apps/builder.py
Aijuh: I also thought abot that.
I thought starting with `init` might cause problem.

Aijuh:
I fixed codes. In terms of simplifying signal, I used `QButtonGroup.checkedId()` method.

BECATRUE:
> I fixed codes. In terms of simplifying signal, I used QButtonGroup.checkedId() method.

Still, I think using an `Enum` class is better because we can check each mapping directly.

Also, you should consider which one between `buttonClicked` and `buttonToggled` signal is better.
The former is emitted with only clicked id (or button), and the latter is emitted with clicked id (or button) and its status.
In other words, you DON'T need to check what button is clicked in `scanTypeToggled` and it is okay to just check its argument.

iquip/apps/builder.py
BECATRUE: @Aijuh

Aijuh:
Updated code to use `enum` (`idToButton`}, that mapps id of button to selected widget's name.
But There was already a dictionary{`scanButtonsDict`} that mapps widget's name to id. Two dictionary have exact information. I want to get rid of one of them. But in order to get key from value, iteration is also needed.
If we remove `scanButtonsDict`, iteration occur only once at the `__init__` method. So it seems more preferable.

And buttonClicked signal emits `QRadioButton` that is clicked, not `id` itself. Even if it can be changed it to emit `int` by `buttonClicked[int].connect`, it caused pylint error.

iquip/apps/builder.py
```diff
@@ -375,17 +398,119 @@ def value(self) -> Dict[str, Any]:
         selected = self.state["selected"]
         return self.state[selected]
 
+    def scanTypeClicked(self, selectedButton: QRadioButton):
```
BECATRUE: Can you check if a `pyqtSlot()` decorator is available?

BECATRUE:
I also think removing the former is efficient if you can remove `scanButtonsDict` or an enum class.

If you use an enum, you can simplify the loop in `__init__()` as follows:
```python
class ScanType(IntEnum):
  NoScan = 0
  RangeScan = 1

def __init__(self, ...):
  ...
  for scanType in _ScanEntry.ScanType:
    ty = scanType.name
    buttonId = scanType.value
    scanCls, buttonName = scanDict[ty]
    ...
```

It makes one easy to change the order of the scan widgets and buttons.

BECATRUE (changes requested):

Aijuh:
I didn't have experience with enum in python and misunderstand the concept.
I thought that `Enum` is only used at C language.
I will upgrade code by using `Enum`.

Aijuh:
I modified the code to use `Enum` class. It simplifies duplicated data. 

In terms of `pyqtSlot`, I should have imported `QAbstractButton` class to add it.

When I defined key of `Enum` class as `NoScan` and `RangeScan`, pylint told that camel case is bad convention. So I disabled pylint for now.

iquip/apps/builder.py
```diff
@@ -3,13 +3,14 @@
 import json
 import logging
 from typing import Any, Dict, Optional, Tuple, Union
+from enum import IntEnum, unique
```
BECATRUE, line 5: Please move this before `from typing ...` for alphabetical order.

iquip/apps/builder.py
```diff
@@ -301,9 +302,17 @@ class _ScanEntry(_BaseEntry):
           "selected": The name of the selected scannable type.
           "NoScan", "RangeScan", "CenterScan", and "ExplicitScan": The dictionary that contains 
             argument info of the corresponding scannable type.
-        stackWidget: The QStackedWidget that contains widgets of each scannable type.
-        rangeWidget: TODO(AIJUH): The Widget that will be removed at next PR.
+        stackedWidget: The QStackedWidget that contains widgets of each scannable type.
+        scanButtonGroup: The QButtonGroup that groups QRadiobuttons for selecting scan widget.
+        scanWidgets: The dictionary that contains each scan widget.
     """
+    @unique
```
BECATRUE, line 310: 👍

iquip/apps/builder.py
```diff
@@ -301,9 +302,17 @@ class _ScanEntry(_BaseEntry):
...
+    # pylint: disable=invalid-name
+    class ScanType(IntEnum):
```
BECATRUE, line 312: Please add an empty line before and after the class definition.

iquip/apps/builder.py
```diff
@@ -355,7 +384,7 @@ def get_state(self) -> Dict[str, Any]:
                     state[ty] = default
```
BECATRUE, line 386: Can you replace the list `["NoScan", "RangeScan", "CenterScan", "ExplicitScan"]` with a tuple, in the line above?

iquip/apps/builder.py
```diff
@@ -320,16 +329,36 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
...
+        scanDict = {"NoScan": (_NoScan, "No scan"), "RangeScan": (_RangeScan, "Range")}
+        self.scanWidgets = {}
+        for scanType in _ScanEntry.ScanType:
+            ty = scanType.name
+            buttonId = scanType.value
+            scanCls, buttonName = scanDict[ty]
+            scanWidget = scanCls(procdesc, self.state[ty])
+            self.stackedWidget.addWidget(scanWidget)
+            self.scanWidgets[ty] = scanWidget
+            button = QRadioButton(buttonName)
```
BECATRUE: Is there any reason not to set the parent to `self`?

iquip/apps/builder.py
```diff
@@ -320,16 +329,36 @@ def __init__(self, name: str, argInfo: Dict[str, Any], parent: Optional[QWidget]
...
+            buttonLayout.addWidget(button)
+            self.scanButtonGroup.addButton(button, buttonId)
+        self.scanButtonGroup.buttonClicked.connect(self.scanTypeClicked)
+        selected = self.state["selected"]
+        self.stackedWidget.setCurrentWidget(self.scanWidgets[selected])
+        selectedEnum = _ScanEntry.ScanType[selected]
```
BECATRUE: `selectedScanType` or `selectedType` looks better.

iquip/apps/builder.py
```diff
@@ -375,17 +404,120 @@ def value(self) -> Dict[str, Any]:
         selected = self.state["selected"]
         return self.state[selected]
 
+    @pyqtSlot(QAbstractButton)
+    def scanTypeClicked(self, selectedButton: QAbstractButton):
+        """Switches current scan widget at stacked layout in _ScanEntry.
```
BECATRUE: How about "in the stacked layout."?

iquip/apps/builder.py
```diff
@@ -375,17 +404,120 @@ def value(self) -> Dict[str, Any]:
...
+        
+        Once the checked button at radiobutton group clicked, this is called.
```
BECATRUE: It looks a little weird.
How about "Once a scan type button is clicked, ..."?

iquip/apps/builder.py
```diff
@@ -375,17 +404,120 @@ def value(self) -> Dict[str, Any]:
...
+
+        Attributes:
+            selectedButton: A QRadioButton that is clicked.
```
BECATRUE: It can be simplified to "The clicked QRadioButton.".

iquip/apps/builder.py
```diff
@@ -375,17 +404,120 @@ def value(self) -> Dict[str, Any]:
...
+        """
+        selectedEnum = _ScanEntry.ScanType(self.scanButtonGroup.id(selectedButton))
```
BECATRUE: Please rename it.

iquip/apps/builder.py
```diff
@@ -301,9 +302,17 @@ class _ScanEntry(_BaseEntry):
           "selected": The name of the selected scannable type.
           "NoScan", "RangeScan", "CenterScan", and "ExplicitScan": The dictionary that contains 
             argument info of the corresponding scannable type.
-        stackWidget: The QStackedWidget that contains widgets of each scannable type.
-        rangeWidget: TODO(AIJUH): The Widget that will be removed at next PR.
+        stackedWidget: The QStackedWidget that contains widgets of each scannable type.
+        scanButtonGroup: The QButtonGroup that groups QRadiobuttons for selecting scan widget.
```
BECATRUE: "a scan widget" or "scan widgets".

BECATRUE (changes requested):
Almost final!

iquip/apps/builder.py
BECATRUE: If you want to use "at radiobutton group", "in the button group" looks better.
And, please write "is clicked".

iquip/apps/builder.py
```diff
@@ -406,15 +408,15 @@ def value(self) -> Dict[str, Any]:
...
-        Once the checked button at radiobutton group clicked, this is called.
+        Once a scan type button at radiobutton group clicked, this is called.
 
         Attributes:
-            selectedButton: A QRadioButton that is clicked.
+            selectedButton: The clicked QRadioButton.
         """
-        selectedEnum = _ScanEntry.ScanType(self.scanButtonGroup.id(selectedButton))
-        self.stackedWidget.setCurrentWidget(self.scanWidgets[selectedEnum.name])
+        selectedType = _ScanEntry.ScanType(self.scanButtonGroup.id(selectedButton))
```
BECATRUE: You renamed the same variable in `_ScanEntry.__init__()` to `selectedScanType`, hence please rename this to the same.

BECATRUE (changes requested):
Final!

Aijuh:
I've modified the code to return the selected arguments in the GUI when submitting an experiment.  The arguments below are submitted to the Artiq proxy.

image

iquip/apps/builder.py
BECATRUE: @Aijuh

iquip/apps/builder.py
```diff
@@ -403,8 +403,8 @@ def value(self) -> Dict[str, Any]:
         
         Returns the dictionary of the selected scannable arguments.
         """
-        selected = self.state["selected"]
-        return self.state[selected]
+        selectedWidget = self.stackedWidget.currentWidget()
```
BECATRUE: How about `selectedScanWidget` as you referred this as `scanWidget` in `__init__()`?

iquip/apps/builder.py
```diff
@@ -564,10 +565,12 @@ def scanArguments(self) -> Dict[str, Any]:
         Returns the arguments of the range scan.
         """
         return {
+            "ty": "RangeScan",
             "start": self.startSpinBox.value(),
             "stop": self.stopSpinBox.value(),
             "npoints": self.npointsSpinBox.value(),
-            "randomize": self.randomizeCheckBox.isChecked()
+            "randomize": self.randomizeCheckBox.isChecked(),
+            "seed": None
```
BECATRUE, line 573: As I said before, the feature to set the seed manually looks necessary, hence please leave this as an issue.

BECATRUE (changes requested):
I totally forgot `value()` should be updated. Thank you👍

Aijuh:
I updated code and made issue about `seed` arguments.

BECATRUE (approved):
LGTM!

Issue #258, opened 2024-03-16
Support remote part in data viewer


BECATRUE:
### Feature you want to implement
I will implement the remote part in data viewer to fetch and show a specific H5 result file.

### How the feature is implemented
For this, the following functions are required.
- Widgets that select date, hour, and RID.
- Function to fetch RID list for a specific date (and hour).
Once a date or an hour is selected, the corresponding RID list is fetched.
- Function to fetch dataset list for a specific RID.
Once a RID is selected, the list of stored dataset names is fetched.  
- Function to fetch a specific dataset.
Once a dataset name is selected, the corresponding dataset is fetched.

To support these features, the corresponding functions should be implemented in ARTIQ-PROXY.

PR #262, opened 2024-03-25
Fetch RID list in remote part

BECATRUE:
This is a part of #258.

What I have implemented is:
- Add widgets to select a target date and hour.
- Once any is changed, fetch the corresponding RID list.
- Reconstruct signals in dataviewer app.

image

The figure means that RIDs executed today are 765 to 769.

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -1016,6 +1016,7 @@ class DataViewerApp(qiwis.BaseApp):
         self.policy: Optional[SimpleScanDataPolicy] = None
         self.axis: Tuple[int, ...] = ()
         self.dataPointIndex: Tuple[int, ...] = ()
+        self.startDatasetListThread()
         realtimePart, remotePart = (self.frame.sourceWidget.stack.widget(buttonId)
                                     for buttonId in SourceWidget.ButtonId)
         realtimePart.syncToggled.connect(self._toggleSync)
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -1012,7 +1012,7 @@ class _RealtimeFetcherThread(QThread):
             logger.exception(msg)
 
 
-class _RidListFromDateHour(QThread):
+class _RidListOfDateHourThread(QThread):
     """QThread for fetching the RID list of the target date and hour.
     
     Signals:
@@ -1059,8 +1059,9 @@ class DataViewerApp(qiwis.BaseApp):
     
     Attributes:
         frame: DataViewerFrame instance.
-        realtimeFetcherThread, realtimeListThread: The most recently executed
-          _RealtimeFetcherThread, _RealtimeListThread instance, respectively.
+        realtimeFetcherThread, realtimeListThread, ridListOfDateHourThread:
+          The most recently executed _RealtimeFetcherThread, _RealtimeListThread, and
+          _RidListOfDateHourThread instance, respectively.
         policy: Data policy instance. None if there is currently no data.
         axis: The current plot axis parameter indices. See SimpleScanDataPolicy.extract().
         dataPointIndex: The most recently selected data point index.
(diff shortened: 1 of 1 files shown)


iquip/apps/dataviewer.py
```diff
@@ -336,33 +340,58 @@ class _RemotePart(QWidget):
...
-        label: Label for showing the execution time of the experiment.
+        dateEdit: QDateEdit for target date.
+        hourCheckBox: QCheckBox for enabling to select an hour.
+        hourSpinBox: QSpinBox for target hour.
+        ridComboBox: QComboBox for showing the RID list of the selected date (and hour).
 
     Signals:
-        ridEditingFinished(rid): The editingFinished signal of spinbox is emitted.
-          The current spinbox value is given as rid.
+        dateHourChanged(date, hour): The target date and hour are changed. If the hour is not set,
```
kangz12345: Let's clarify that the date is in the ISO format.

kangz12345: Also, it does not give any hint about the type of `hour` when it is set.

iquip/apps/dataviewer.py
```diff
@@ -336,33 +340,58 @@ class _RemotePart(QWidget):
...
-        layout.addWidget(self.spinbox)
-        layout.addWidget(self.label)
+        layout.addWidget(self.dateEdit)
+        layout.addWidget(self.hourCheckBox)
+        layout.addWidget(self.hourSpinBox)
+        layout.addWidget(self.ridComboBox)
         # signal connection
-        self.spinbox.editingFinished.connect(self._editingFinished)
+        self.dateEdit.dateChanged.connect(self.updateRidComboBox)
+        self.hourCheckBox.stateChanged.connect(self.enableHourSpinBox)
```
kangz12345: How about using `clicked(checked)` signal and connect `self.hourComboBox.setEnabled` as its slot? `QCheckBox` inherits `QAbstractButton`, but I'm not sure you can use `clicked` signal.

iquip/apps/dataviewer.py
```diff
@@ -336,33 +340,58 @@ class _RemotePart(QWidget):
...
+        self.hourCheckBox.stateChanged.connect(self.updateRidComboBox)
+        self.hourSpinBox.valueChanged.connect(self.updateRidComboBox)
+
+    @pyqtSlot(int)
+    def enableHourSpinBox(self, state: int):
```
kangz12345: In this case, the method name should be `setEnabled...`.

iquip/apps/dataviewer.py
```diff
@@ -886,18 +905,25 @@ def _filter(self, names: List[str]) -> List[str]:
         """
         return [name for name in names if not name.endswith((".parameters", ".units"))]
 
+    def stop(self):
+        """Stops the thread."""
+        try:
+            self.websocket.close()
```
kangz12345, line 903: Is this thread-safe?

kangz12345 (changes requested):
The way of getting the realtime part and remote part looks so weird... Sorry for the messy design 😄 Let's modify that later.

iquip/apps/dataviewer.py
BECATRUE: Umm.. I think it is thread-safe as this method is located in `QThread`.
However I'm not sure about this..

iquip/apps/dataviewer.py
BECATRUE: Wow, it works well! I'll do that.

iquip/apps/dataviewer.py
BECATRUE: I used `clicked` signal instaed of `stateChanged` signal, so now this method becomes unusable.

iquip/apps/dataviewer.py
kangz12345: I mean while the `QThread` is communicating with the artiq proxy using `self.websocket`, someone can call `stop()` in another thread.
Then the `websocket` object must support the thread-safe API.
Several different thread can use the `websocket` instance in this code.

BECATRUE: I see. According to [docs](https://websockets.readthedocs.io/en/stable/reference/sync/client.html#websockets.sync.client.ClientConnection.close), `close()` waits for all incoming messages to be read with `recv()`, so it looks thread-safe.
And I can't say for sure, but looking at the code of `websocket`, it uses `threading.Lock()` for guaranteeing thread-safety.

kangz12345: Okay, then maybe it is thread-safe!

kangz12345 (approved):
LGTM

PR #267, opened 2024-04-01
Restart fetching schedule when the connection is lost

### LINKED ISSUE #250: Scheduler "Restart" button is not enabled
### Describe the bug
AFAIK, the "Restart" button should be enabled when the scheduler fails to fetch the current schedule. (Is it right?)
However, I experienced a network disconnection (so the scheduler, dataviewer, stage controller, VSCode was disconnected), but the "Restart" button of the scheduler was not enabled.
I could see the schedule via `artiq_client show schedule` in a terminal though.
Note that the other apps were able to re-connect.

### Expected behavior
"Restart" button of the scheduler app should be enabled when it loses the connection.

### To Reproduce
I checked that this is reproduced in the following situation:
- `kangz12345/temp-stage` branch
- VPN on, using `iquip` normally
- Turn off the VPN -> all the connection lost (including scheduler, but the restart button is not enabled)
- Turn on the VPN -> Of course the button is still not enabled.

There was no error message in terminal and logger.

BECATRUE:
This resolves #250.

I made sure if the "restart" button got enabled when the connection was lost due to VPN problem.

### DIFF
--- a/iquip/apps/scheduler.py
+++ b/iquip/apps/scheduler.py
@@ -12,7 +12,6 @@ from PyQt5.QtCore import (
 )
 from PyQt5.QtGui import QGuiApplication
 from PyQt5.QtWidgets import QAction, QPushButton, QTableView, QVBoxLayout, QWidget
-from websockets.exceptions import WebSocketException
 from websockets.sync.client import connect
 
 import qiwis
@@ -64,8 +63,7 @@ class _ScheduleFetcherThread(QThread):
                     except TimeoutError:
                         if websocket.ping().wait(5):
                             continue
-                        else:  # connection is lost
-                            break
+                        break  # connection is lost
                     schedule = []
                     for rid, info in json.loads(response).items():
                         expid = info["expid"]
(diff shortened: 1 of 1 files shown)


iquip/apps/scheduler.py
```diff
@@ -73,7 +78,7 @@ def run(self):
                             arguments=expid["arguments"]
                         ))
                     self.fetched.emit(schedule)
-        except WebSocketException:
+        except:  # pylint: disable=bare-except
```
kangz12345: AFAIK, the bare except is in fact not desirable. Instead, `except Exception` would be better.
The bare except will prevent system level halts as well, for example Keyboard interrupt.

kangz12345 (changes requested):

iquip/apps/scheduler.py
BECATRUE: Oh, I didn't know much about the bare except.
I will apply your suggestion.

BECATRUE: Umm.. Your solution causes a [broad-exception-caught](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/broad-exception-caught.html).
Is it okay to disable this warning?

kangz12345: Yes, since this is for preventing the program not to crash and we are logging the exception properly, we can catch such a broad exception.

kangz12345 (approved):
LGTM

PR #276, opened 2024-04-04
Improve builder UI compactly

### LINKED ISSUE #254: More compact builder UI
How about putting the argument, scan parameter, schedule option panels into tabs to make each builder frame smaller?

Each builder frame has 3 tabs: argument, scan parameter, schedule option.

### Additional context
Currently each builder frame is too large (the empty scan parameter panel consumes a lot of display area) and it is quite inconvenient.

BECATRUE:
This closes #254.

What I have implemented is:
- Leave only the class name label, and move the experiment name in its tooltip.
- Group three list widgets; arguments, scannables, and scheduling options.
- Group two buttons.

### DIFF
--- a/iquip/apps/builder.py
+++ b/iquip/apps/builder.py
@@ -293,7 +293,6 @@ class _DateTimeEntry(_BaseEntry):
         return None
 
 
-# TODO(AIJUH): Add other scan type classes.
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
LGTM

PR #280, opened 2024-04-05
Fix the bug that a normal websocket closing leaves an exception log

### LINKED ISSUE #279: Normal websocket closing leave an exception log
### Describe the bug
In some `QThread` using websocket, a `ConnectionClosedOk` exception is not handled, so it leaves an exception-level log.

### Expected behavior
It is intended, so it must not leave any log.

### To Reproduce
1. Go to the dataviewer app.
2. Switch to the remote mode.
3. Check the logger app.

### Correction methods
Like `iquip.apps.dataviewer._RealtimeDatasetThread`, we should handle a `ConnectionClosedOk` exception.

### Screenshots
image

BECATRUE:
This resolved #279.


kangz12345 (approved):
LGTM

PR #285, opened 2024-04-07
Stage controller app

### LINKED ISSUE #245: Motorized stage monitor
A simple (controllable) monitor for motorized stages.
- Realtime position update.
- Absolute move.
- Relative move.

A new app "controller" will be added, which is for non-artiq devices.

### Subtasks
- [x] #248 
- [x] #275
- [x] #285

kangz12345:
This closes #245, implementing the motorized stage controller app.

Key structure of this app:
- It uses a dedicated non-GUI `QThread` to actually control the stage RPC clients.
- `StageManager` lives in the non-GUI thread, and provides signals to control the RPC clients.
- `StageProxy` provides a convenient interface to control the RPC clients which live in a different thread (automatically converts method calls to signal emissions).
- Every proxied RPC call is non-blocking.

Probable issues:
1. There is no way to restart the manager thread if it crashes somehow. However the RPC errors will not cause such crash.
2. The move requests are non-blocking so rapid move requests interrupts the previous moves and even the server does not protect it, i.e., it does not wait for the previous move to be done. For example, when the relative move button is rapidly clicked 5 times with step size of 1um, it may actually move less than 5um.

Screen Shot 2024-04-07 at 16 56 27

### DIFF
--- a/iquip/apps/stage.py
+++ b/iquip/apps/stage.py
@@ -446,12 +446,13 @@ class StageControllerApp(qiwis.BaseApp):
             widget.setConnected(connected)
 
     @pyqtSlot(str, Exception)
-    def handleClientError(self, key: str, _error: Exception):
+    def handleClientError(self, key: str, error: Exception):
         """Handles clientError signal.
         
         Args:
             See StageManager.clientError signal.
         """
+        logger.error("Stage %s reported an error.", exc_info=error)
         self.handleConnectionChanged(key, False)
 
     @pyqtSlot(str, float)
--- a/iquip/apps/stage.py
+++ b/iquip/apps/stage.py
@@ -460,7 +460,7 @@ class StageControllerApp(qiwis.BaseApp):
         Args:
             See StageManager.clientError signal.
         """
-        logger.error("Stage %s reported an error.", exc_info=error)
+        logger.error("Stage %s reported an error.", key, exc_info=error)
         self.handleConnectionChanged(key, False)
 
     @pyqtSlot(str, float)
--- a/iquip/apps/stage.py
+++ b/iquip/apps/stage.py
@@ -4,12 +4,14 @@ from __future__ import annotations
 
 import functools
 import logging
-from typing import Callable, Dict, Optional, Tuple
+from typing import Any, Callable, Dict, Optional, Tuple
 
-from PyQt5.QtCore import pyqtSignal, pyqtSlot, QObject, Qt
+from PyQt5.QtCore import pyqtSignal, pyqtSlot, QObject, Qt, QThread, QTimer
 from PyQt5.QtWidgets import (
-    QAbstractSpinBox, QDoubleSpinBox, QHBoxLayout, QPushButton, QVBoxLayout, QWidget,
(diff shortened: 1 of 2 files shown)


iquip/apps/stage.py
```diff
@@ -336,3 +339,141 @@ def _relativePositiveMove(self):
...
+            See StageControllerApp.
+        """
+        super().__init__(parent=parent)
+        self.widgets: Dict[str, StageWidget] = {}
+        layout = QGridLayout(self)
+        for name, info in stages.items():
+            widget = StageWidget(self)
+            groupbox = QGroupBox(name, self)
+            groupboxLayout = QHBoxLayout(groupbox)
+            groupboxLayout.addWidget(widget)
```
BECATRUE: The `groupboxLayout` is used once, so these two lines can be written in just a line:
`QHBoxLayout(groupbox).addWidget(widget)`

iquip/apps/stage.py
```diff
@@ -336,3 +339,141 @@ def _relativePositiveMove(self):
...
+            layout.addWidget(groupbox, *info["index"])
+            self.widgets[name] = widget
+
+
+class StageControllerApp(qiwis.BaseApp):
+    """App for monitoring and controlling motorized stages."""
```
BECATRUE: Please write the docstrings about attributes!

iquip/apps/stage.py
```diff
@@ -336,3 +339,141 @@ def _relativePositiveMove(self):
...
+        name: str,
+        stages: Dict[str, Dict[str, Any]],
+        period: float = 0.5,
+        parent: Optional[QObject] = None,
+    ):
+        """Extended.
+        
+        Args:
+            stages: Dictionary of stage information. Each key is the name of the
+              stage and the value is agian a dictionary, whose structure is:
```
BECATRUE: "again"

iquip/apps/stage.py
```diff
@@ -336,3 +339,141 @@ def _relativePositiveMove(self):
...
+            widget.setConnected(connected)
+
+    @pyqtSlot(str, Exception)
+    def handleClientError(self, key: str, error: Exception):
+        """Handles clientError signal.
+        
+        Args:
+            See StageManager.clientError signal.
+        """
+        logger.error("Stage %s reported an error.", exc_info=error)
```
BECATRUE: Isn't the first argument, `key`, missing?

iquip/apps/stage.py
```diff
@@ -336,3 +339,141 @@ def _relativePositiveMove(self):
...
+            See StageManager.positionReported signal.
+        """
+        try:
+            widget = self.frame.widgets[key]
+        except KeyError:
+            logger.exception("Position reported key does not exist.")
+        else:
+            widget.setPosition(position_m)
+
+    def __del__(self):
```
BECATRUE, line 480: Cool!

BECATRUE (changes requested):
Do you have any plan to use `StageManager.clear` signal? I'm just curious!

BECATRUE (approved):
LGTM

PR #290, opened 2024-04-09
Use `requirements.txt` instead of direct installation

### LINKED ISSUE #286: Use `requirements.txt` instead of direct installation
To manage easily, we need to use a requirements file instead of direct installation in GitHub actions.

BECATRUE:
This closes #286.

After merged, I will release the first official version, v1.0.0.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11"]
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


kangz12345 (commented):
I think even if `numpy` is included in other packages, we should include it in the `requirements.txt` because the package dependencies might change.
We are using `numpy` so it should be explicitly required to be installed.

BECATRUE:
I see! I will add its dependency.

BECATRUE:
I realized at least the last two-year `numpy` versions support since 3.9.
Sure, it is easy to install different `numpy` version according to the Python version.
However, the support for Python 3.8 will end sometime this year, so why not check about this version in GitHub actions?

kangz12345:
> I realized at least the last two-year `numpy` versions support since 3.9. Sure, it is easy to install different `numpy` version according to the Python version. However, the support for Python 3.8 will end sometime this year, so why not check about this version in GitHub actions?

That makes sense. I agree with you.

kangz12345 (approved):

PR #306, opened 2024-05-09
Fix transposed 2D image of dataviewer

### LINKED ISSUE #299: 2D data viewer renders data points at the wrong position
### Describe the bug
When we select y axis as well, i.e., plot a 2D image, the dataviewer app renders the points at the wrong position.
The axes values are correct, but maybe the indices are switched.
For example, consider the following 2D scan:
- X: range(0, 10, 2)
- Y: range(0, 5, 1)
Then the point for (X=2, Y=2) is (1, 2) in terms of index, and rendered at (X=4, Y=1).

### Expected behavior
Each data point should be rendered at its right place.

### To Reproduce
1. Open iquip and run a 2D scan experiment.
2. Set X and Y axis in dataviewer.
3. See the rendered image (2D plot).

Maybe some mistakes took place while drawing an image.

kangz12345:
The 2D image was transposed as [`pyqtgraph.ImageItem.setImage()`](https://pyqtgraph.readthedocs.io/en/latest/api_reference/graphicsItems/imageitem.html#pyqtgraph.ImageItem.setImage) takes a column-major data instead of row-major, which is the standard for `numpy`.

This resolves #299.

### DIFF
--- a/iquip/apps/dataviewer.py
+++ b/iquip/apps/dataviewer.py
@@ -281,7 +281,7 @@ class ImageViewer(NDArrayViewer):  # pylint: disable=too-few-public-methods
         x, y = np.floor(dataPos.x()), np.floor(dataPos.y())
         w, h = self.image.width(), self.image.height()
         if 0 <= x < w and 0 <= y < h:
-            return int(x), int(y)
+            return int(y), int(x)  # row, column index
         return None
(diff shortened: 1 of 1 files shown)


BECATRUE (approved):
LGTM

PR #308, opened 2024-05-16
Apply scale to scan arguments

### LINKED ISSUE #303: Scan argument scale for common units (e.g. us) is not applied
### Describe the bug
When I use a `Scannable(RangeScan(0., 100*us, 21, True), unit="us"))` and submit the experiment through iquip, the obtained values are not scaled by 1e-6. In other words, they are (0., 5., 10., ...) instead of (0., 5e-6, 1e-5, ...).

### Expected behavior
The appropriate scale should be applied for the common units.

kangz12345:
This resolves #303.


kangz12345:
@BECATRUE Could you review this PR?

iquip/apps/builder.py
```diff
@@ -664,7 +664,7 @@ def __init__(
     def scanArguments(self) -> Dict[str, Any]:
         """Overridden."""
         sequenceText = self.sequenceEdit.text()
-        sequence = [float(x) for x in sequenceText.split()]
+        sequence = [float(x) * self.scale for x in sequenceText.split()]
```
BECATRUE: As I found in [ARTIQ](https://github.com/m-labs/artiq/blob/56bd975a347155181cf1fdd5fc383742d22daf57/artiq/gui/entries.py#L532), in `ExplicitScan`, the scale argument may be not used.
Could you check it?

BECATRUE (changes requested):

iquip/apps/builder.py
kangz12345: Oh, you are right! Explicit scan does not use any unit so it should not use the scale parameter.

BECATRUE (approved):
LGTM!

PR #319, opened 2024-10-29
Add `max-positional-arguments=10` in pylintrc

### LINKED ISSUE #318: Updated pylint complains about the number of positional arguments
As the recent pylint now checks the number of positional arguments, we should update our codes or `.pylintrc`.
Because of this, my new PRs do not pass the pylint check.

I will suggest how we should do about this.
Feel free to give your opinions!

See also: https://pylint.readthedocs.io/en/stable/whatsnew/3/3.3/index.html#new-checks

kangz12345:
This is a work-around for #318. I think we should update the codes later for better design.


BECATRUE (changes requested):
Cool!

However, this is not an ultimate solution but just a temporary detour, so how about removing the "resolve" keyword in the PR explanation?

BECATRUE (approved):

PR #323, opened 2025-02-14
Support tag for builder app

### LINKED ISSUE #322: Support tag for builder app
To open multiple builder apps for the same experiment class, they should be distinguished through their app names.
For this, we plan to support a tag for each builder app, which an user can set just before the builder app is opened.

Currently, there is a modal for selecting an experiment class, which is opened after an experiment file is clicked.
I will expand this modal to support a line edit for tags.

### Additional context
The modal will be implemented through inheriting [QDialog](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QDialog.html).

BECATRUE:
This closes #322.

Now, one can set tags for experiment classes, thus multiple builder apps for the same experiment class are available.

- Dialog: Supports for selecting an experiment class and editing its tag.
- Upper builder: Without a tag.
- Lower builder: With the tag, "123".

### DIFF
--- a/iquip/apps/explorer.py
+++ b/iquip/apps/explorer.py
@@ -7,7 +7,8 @@ from typing import Dict, List, Optional, Tuple, Union
 import requests
 from PyQt5.QtCore import QObject, Qt, QThread, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import (
-    QInputDialog, QPushButton, QTreeWidget, QTreeWidgetItem, QVBoxLayout, QWidget
+    QComboBox, QDialog, QHBoxLayout, QLineEdit, QPushButton, QTreeWidget, QTreeWidgetItem,
+    QVBoxLayout, QWidget,
 )
 
 import qiwis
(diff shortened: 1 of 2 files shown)


iquip/apps/explorer.py
```diff
@@ -96,6 +97,48 @@ def run(self):
...
+        layout = QVBoxLayout(self)
+        layout.addWidget(self.comboBox)
+        layout.addWidget(self.lineEdit)
+        layout.addLayout(buttonLayout)
+
+    def getBuilderInfo(self) -> Tuple[str, str]:
+        """Returns the configured builder info.
+
+        Returns:
+            See BuilderInfoDialog.
```
kangz12345: Let's explicitly document that the returned tuple is `(cls, tag)`.
You can still say "See BuilderInfoDialog."

iquip/apps/explorer.py
```diff
@@ -96,6 +97,48 @@ def run(self):
...
 
+class BuilderInfoDialog(QDialog):
+    """Dialog for setting builder info.
+    
+    There are two types of info.
+        cls: Target experiment class name.
+        tag: Additive tag for multiple builder apps.
+    """
+
+    def __init__(self, experimentInfos: List[str], parent: Optional[QWidget] = None):
```
kangz12345: Why is the name `experimentInfos` instead of `clsNames` for example?

kangz12345: Based on your usage, I think it is not `List` but `Iterable`.

iquip/apps/explorer.py
```diff
@@ -216,54 +259,47 @@ def fetchExperimentInfo(self, item: QTreeWidgetItem):
...
         Args:
-            experimentClsName: The class name of the experiment.
+            cls, tag: See BuilderInfoDialog.
             experimentInfo: The experiment information. See protocols.ExperimentInfo.
         """
+        if tag:
+            tag = f"({tag})"
         self.qiwiscall.createApp(
-            name=f"builder - {self.selectedExperimentPath}:{experimentClsName}",
+            name=f"builder {tag} - {self.selectedExperimentPath}:{cls}",
```
kangz12345: This will make 2 spaces for an empty tag.

kangz12345 (changes requested):

iquip/apps/explorer.py
BECATRUE: > Why is the name experimentInfos instead of clsNames for example?

Umm.. I also think `clsNames` is much proper. I will take it.

iquip/apps/explorer.py
```diff
@@ -96,6 +97,48 @@ def run(self):
...
+    There are two types of info.
+        clsName: Target experiment class name.
+        tag: Additive tag for multiple builder apps.
+    """
+
+    def __init__(self, clsNames: Iterable[str], parent: Optional[QWidget] = None):
+        """Extended.
+        
+        Args:
+            See thread.ExperimentInfoThread.fetched signal.
```
kangz12345, line 112: Now this is actually not a proper redirection, but I think it is pretty intuitive enough. If you agree with it, LGTM. You can update the docstring as well.

kangz12345 (approved):

iquip/apps/explorer.py
BECATRUE: Then, let's keep it as it is.

PR #333, opened 2025-04-17
Implement oven controller

### LINKED ISSUE #329: Support oven controller
For convenience, I will implement an oven controller using the RPC server for PSU developed by snu-quiqcl/quiqcl-device#97.

The overall structure is motivated from the stage controller, but this will handle only one PSU so be simpler than the stage one.

I will implement the following four classes.
1. `OvenControllerApp`
2. `OvenControllerFrame`
3. `OvenManager`
4. `OvenProxy`

BECATRUE:
This closes #329.

I implement an oven controller based on snu-quiqcl/quiqcl-device#99.
The overall structure is similar to the stage controller but much simpler.

The GUI is as follows:
- 1st row: Real-time current & voltage, elapsed time, and timer reset button.
- 2nd row: Target current, expiration time, and on/off button.

image


iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+import logging
+from typing import Any, Callable, List, Optional, Tuple
+
+from PyQt5.QtCore import pyqtSignal, pyqtSlot, QObject, Qt, QThread, QTimer
+from PyQt5.QtWidgets import (
+    QAbstractSpinBox, QDoubleSpinBox, QHBoxLayout, QPushButton, QSpinBox, QVBoxLayout, QWidget
+)
+
+import qiwis
+from sipyco.pc_rpc import Client
```
kangz12345: `sipyco` is in an equivalence class with `PyQt5`, as it is an external 3rd-party package.

iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+    @functools.wraps(function)
+    def wrapped(self: OvenManager, *args, **kwargs):
+        """Adds the client object."""
+        client = self._client  # pylint: disable=protected-access
+        if client is None:
+            logger.error("Failed to get client.")
+            self.clientError.emit(ValueError("There is no client."))
+            return
+        try:
+            function(self, client, *args, **kwargs)
```
kangz12345, line 40: Why not `return function(...)`? Should it return `None` only?

iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+    
+    An instance of this class should be moved to a thread other than the main
+      GUI thread to prevent GUI from freezing.
+    Therefore, the private methods must not be called from the main thread.
+    Instead, use signals to communicate.
+
+    Signals:
+        connectionChanged(connected): The client connection status is changed,
+          with the connection status as True for connected, False for disconnected.
+        outputChanged(outputted): The output status is changed,
```
kangz12345: `outputted` -> `output` would be better, as there is no word 'outputted'.

iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+    def _closeTarget(self):
+        """Closes the RPC client."""
+        if self._client is None:
+            logger.error("Failed to close target: RPC client does not exist.")
+            return
+        self._client.close_rpc()
+        self.connectionChanged.emit(False)
+
+    @pyqtSlot(tuple)
+    def _openTarget(self, info: RPCTargetInfo):
```
kangz12345, line 108: How about unpacking `info` in the body to prevent confusing, for example,
```python
ip, port, target_name, target_channel = info
```
instead of using it as `info[3]`, etc.?

iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+    def setDisplayedTimer(self, on_duration: float):
+        """Sets the timer displayed on the widget.
+        
+        Args:
+            on_duration: Duration while the output is turned on.
+        """
+        self.timerDisplayBox.setValue(on_duration)
+
+    @pyqtSlot(bool)
+    def setOutput(self, output_: bool):
```
kangz12345, line 335: Why is `output` followed by `_`?

iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+        """Sets the current output status.
+
+        This also changes the enabled status and the output button text.
+        
+        Args:
+            open_: True for turned on, False for turned off.
```
kangz12345: Oh, I think this should be updated.

iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+        """
+        self.outputButton.setEnabled(True)
```
kangz12345, line 343: Why is this necessary?

iquip/apps/oven.py
```diff
@@ -0,0 +1,521 @@
...
+    def handleConnectionChanged(self, connected: bool):
+        """Handles connectionChanged signal.
+        
+        Args:
+            See OvenManager.connectionChanged signal.
+        """
+        self.frame.setConnected(connected)
+
+    @pyqtSlot(bool)
+    def handleOutputChanged(self, outputted: bool):
```
kangz12345: Let's change `outputted` -> `output`.

kangz12345 (changes requested):

kangz12345:
There is a bug: the maximum value of the spin box which displays the elapsed time of the timer seems to be set to 100. It stops at 100s.

kangz12345:
Also, how about making a sound when the timer stops?

iquip/apps/oven.py
BECATRUE: `use_client()` decorator is only used for a slot, so its return value would not be used.
Also I just referenced your `use_client()` in `stage.py`😅

iquip/apps/oven.py
BECATRUE: I just wanted to distinguish from `output` signal.

iquip/apps/oven.py
BECATRUE: This is because when the button is clicked, it is disabled in `_outputButtonClicked()` to prevent from that one clicks the button before the output is applied.

BECATRUE:
Only in GitHub action, importing `PyQt5.QtMultimedia` causes several Pylint warnings as below:


I have tried to resolve this problem, but couldn't do that..
So I just disabled theses warnings.

iquip/apps/oven.py
```diff
@@ -463,6 +475,10 @@ def checkTimer(self):
         if self._on_duration_ms >= expirationTime * 1000:
             self.frame.outputButton.click()
             self.resetTimer()
+            print(1)
```
kangz12345: Maybe this should be removed.

iquip/apps/oven.py
```diff
@@ -463,6 +475,10 @@ def checkTimer(self):
...
+            if self.timerExpireMediaPlayer.mediaStatus() == QMediaPlayer.LoadedMedia:
```
kangz12345, line 478: What happens otherwise?

kangz12345 (changes requested):
Is the volume of the sound file increased?

BECATRUE:
> Is the volume of the sound file increased?

Umm.. I remembered the volume was increased in [this commit](https://github.com/snu-quiqcl/iquip/pull/333/commits/d248afdcbb23b4ac841ed96d645894e314e17615).

iquip/apps/oven.py
BECATRUE: I checked the media status to prevent trying to play the media for the case where there is no loaded media.
(For example, in my laptop, playing a media occurs an error, hence I don't specify the media file.)

One can see the possible media status in the [docs](https://doc.qt.io/archives/qtforpython-5/PySide2/QtMultimedia/QMediaPlayer.html#detailed-description).

